PR #1, opened 2025-12-05
Add CI workflow

myeh01:
This PR adds the workflow file for public CI along with documentation. Extensions not yet in upstream QEMU have been temporarily disabled. A test release with tag `v0.0.0` was created to store the QEMU and clang-tidy binaries as assets.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,6 +53,15 @@ jobs:
         CLANG_TIDY_TARBALL: riscv64-unknown-elf-clang-tidy.tar.xz
         GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
+    - name: clang-format
+      run: |
+        HEAD_SHA=${{ github.event.pull_request.head.sha }}
+        BASE_SHA=${{ github.event.pull_request.base.sha }}
+        MERGE_BASE_SHA=$(git merge-base $HEAD_SHA $BASE_SHA)
+        git diff --name-only $MERGE_BASE_SHA $HEAD_SHA | \
+          egrep "\.(c|h)$" | \
+          xargs -I{} bash -c "if [ -f {} ] ; then clang-format --dry-run -style=file --Werror {} ; fi"
+
     - name: Configure and build (tests and benchmarks)
       run: |
         cmake -B ${{ github.workspace }}/build \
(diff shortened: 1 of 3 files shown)


hdelassus (commented):
Looks great!

Do you think the current infrastructure would allow to build all supported extensions but only test the ones that are supported by QEMU? I see that we are omitting a lot of extensions due to the current QEMU limitations.

Are CI steps sequentially run from top to bottom? If so, should we put the `clang-format` step above the configure and build step?

I believe external contributors will not have direct write access to the repository (at least in their early contributions) but will rather open PRs from their own forks. Are such PRs going to trigger CI automatically or will we be able to easily trigger CI manually in this situation?

myeh01:
> Looks great!
> 
> Do you think the current infrastructure would allow to build all supported extensions but only test the ones that are supported by QEMU? I see that we are omitting a lot of extensions due to the current QEMU limitations.

I can look into this. We may be able to with some modifications to the build flow.

> Are CI steps sequentially run from top to bottom? If so, should we put the `clang-format` step above the configure and build step?

Yes, they are. I can move `clang-format` to the top so it's run first.

> I believe external contributors will not have direct write access to the repository (at least in their early contributions) but will rather open PRs from their own forks. Are such PRs going to trigger CI automatically or will we be able to easily trigger CI manually in this situation?

I'll need to check this, but I think if they open a PR from their fork and set the merge base to be the main branch of this repository, it should run.

hdelassus (dismissed):
Approving to unblock, LGTM

myeh01:
> Looks great!
>
> Do you think the current infrastructure would allow to build all supported extensions but only test the ones that are supported by QEMU? I see that we are omitting a lot of extensions due to the current QEMU limitations.

It looks like CMake variables like `RISCV_ZVE32F` are used to determine which source files to include and which tests to build and run. One possible (but maybe not the best) way to address this might be splitting the extensions list in `cmake/riscv.cmake` into "build only" and "build and test" lists and then defining separate CMake variables like `RISCV_ZVE32F_BUILD` and `RISCV_ZVE32F_TEST` to include source files and build/run tests, respectively.

myeh01:
> I believe external contributors will not have direct write access to the repository (at least in their early contributions) but will rather open PRs from their own forks. Are such PRs going to trigger CI automatically or will we be able to easily trigger CI manually in this situation?

Apparently, I was wrong. It looks like running CI on fork branches is disabled, so we'll have to figure out what to do.

hdelassus (approved):

myeh01:
It looks like the `clang-format` step is failing post-merge since `HEAD_SHA` and `BASE_SHA` are only defined for pull requests. I will fix this.

PR #4, opened 2025-12-09
Update xsfvqdotq kernels and test harnesses

myeh01:
This PR:
- updates the `xsfvqdotq` kernel to improve performance
- updates the kernel APIs to conform more closely to the packed GEMM API
- cleans up the `xsfvqdotq` test harnesses
- updates the GEMM documentation


myeh01:
@ericlove I have updated the code to handle the case when A is not aligned or `rsa` is not a multiple of 4. Each kernel now has a general version which uses `memcpy` to load from `a` and an aligned version which uses `lw`.

CMakeLists.txt
```diff
@@ -452,6 +452,8 @@ if (SKL_BUILD_TESTS OR SKL_BUILD_BENCHMARKS)
     skl_add_test(skl_cvt_zvfofp4min cvt/cvt_ofp_zvfofp4min.c "")
   endif()
   if(RISCV_XSFVQDOTQ)
-    skl_add_test(gemm_a1b01_i8_i8p_i32_xsfvqdotq gemm/gemm_a1b01_i8_i8p_i32_xsfvqdotq.c "-DM=128;-DN=128;-DK=256;-DBETA=1")
+    skl_add_test(gemm_a1b01_i8_i8pc_i32_xsfvqdotq gemm/gemm_a1b01_i8_i8pc_i32_xsfvqdotq.c "-DM=128;-DN=128;-DK=256;-DBETA=1")
+    skl_add_test(skl_gemm_a1b01_i8_i8pc_i32_xsfvqdotq_wrapper gemm/gemm_i8rcprc_i8rcprc_i32rcprc.c
+      "-DM0=1;-DN0=1;-DK0=4;-DM1=12;-DN1=128;-DK1=8;;-DRSB0=1;-DCSB0=0;-DALPHA=1;-DBETA=0;")
```
peichialin: I expect `RSB0` and `CSB0` can still be set properly based on `N0` in `gemm/gemm_i8rcprc_i8rcprc_i32rcprc.c` even they are not explictly specified here.
Besides, we can remove the redundant `;` here.

peichialin (dismissed):
The overall changes LGTM!
I'm happy to see this PR addresses the alignment issue of `a` buffer with more solid validation against the general packed GEMM specification.
Thanks for the patience of introducing the necessary refinements and fixes at a time.

CMakeLists.txt
myeh01: Ok, I have removed the extra `;`. The default tile layout is row major, so we need to set `RSB0` and `CSB0` explicitly to specify a column major layout (or like we discussed before, these tiles can be viewed as row major or column major, but we want to emphasize that they should be viewed as column major).

src/gemm/xsfvqdotq/gemm_a1b01_i8_i8pc_i32_xsfvqdotq.h
```diff
@@ -0,0 +1,80 @@
...
+#include <stddef.h>
+#include <stdint.h>
+
+#if defined(__cplusplus)
+extern "C" {
+#endif
+
+/**
+ * @brief Int8 GEMM kernel using RISC-V xsfvqdotq extension with packed matrix
+ * B.
```
ericlove, line 24:
```suggestion
 * @brief Int8 GEMM kernel using Xsfvqdotq extension and packed B matrix.
```

src/gemm/xsfvqdotq/gemm_a1b01_i8_i8pc_i32_xsfvqdotq.h
```diff
@@ -0,0 +1,80 @@
...
+ * @param b_pack - Pointer to packed matrix B.
+ * @param rsb1 - Row stride between blocks of packed matrix B in elements.
+ * @param c - Pointer to matrix C in row-major format.
+ * @param rsc - Stride between rows of matrix C in elements.
+ * @param accum - Determines if output matrix is accumulated or overwritten.
+ *
+ * Computes `C = A * B_pack + C` (if `accum == true`) or `C = A * B_pack` (if
+ * `accum == false`) for int8 matrices A and packed B with int32 output matrix
+ * C. This kernel uses the SiFive xsfvqdotq extension for vector quad widening
+ * 4D dot product operations to achieve high performance on 8-bit integer data.
```
ericlove, line 40: Can we add an explanation of the packing and transposition of the B matrix blocks, whose purpose is to avoid the need for a final reduction sum on every C matrix element?

src/gemm/xsfvqdotq/pack_b_xsfvqdotq.c
```diff
@@ -13,12 +13,13 @@
...
 
 SKL_FUNC void skl_pack_b_i8_xsfvqdotq(size_t k, size_t n, const int8_t *b,
-                                      size_t rsb, int8_t *b_pack) {
+                                      size_t rsb, int8_t *b_pack, size_t rsb1) {
   size_t k_rem = k % 4;
   size_t k_idx = 0;
 
   for (; k_idx + 3 < k; k_idx += 4) {
     const int8_t *b_read = b + k_idx * rsb;
+    int8_t *b_pack_write = b_pack + (k_idx / 4) * rsb1;
```
ericlove, line 22: The name `b_pack_write` is not very helpful for understanding its purpose, since `b_pack` clearly is "written to."  I take it the intent is for something like a block or row pointer?

src/gemm/packed-gemm.md
```diff
@@ -228,18 +228,24 @@ Where the element types of `A` and `B` are `int8_t` and `C` is accumulated as `i
...
-This is done by dividing the `k` dimension into into `k0=4` and `k1=ceil(k/4)`, while keeping the `n` dimension unchanged (`n0=n` and `n1=1`).
-Moreover, since the dot product is performed along the `k` dimension, it must be contiguous in memory, resulting in a column-major layout within each block (`rsb0=1` and `csb0=4`).
+This is done by dividing the `k` dimension into `k0=4` and `k1=ceil(k/4)`, while keeping the `n` dimension unchanged.
+Moreover, since the dot product is performed along the `k` dimension, it must be contiguous in memory, resulting in a column-major layout within each block.
+
+In the actual implementation, we think of blocks as having fixed dimensions `4x1` (`n0=1`) instead of `4xN`.
+Each column of the `4xN` blocks of B can be thought of as one of these `4x1` blocks, so `n1 = N`.
+And since each `4xN` block of B is column-major and stored contiguously in memory, we find that `rsb0=1` and `csb1=k0*n0`.
+Note that `csb0` is unused since each `4x1` block has only one column.
```
ericlove, line 238: Feel absolutely free to reject this, but I asked Augment to help generate an asciiart diagram for this particular layout, and this is what it provided, in case you think it would be useful:

```
Example: B matrix (k=10, n=6) packed for xsfvqdotq:

Original B (10×6):                     Conceptual Blocking (k0=4, n0=1):
┌─────────────────────────┐            ┌───┬───┬───┬───┬───┬───┐
│ b00 b01 b02 b03 b04 b05 │            │ 4 │ 4 │ 4 │ 4 │ 4 │ 4 │ k1=0
│ b10 b11 b12 b13 b14 b15 │            │ × │ × │ × │ × │ × │ × │
│ b20 b21 b22 b23 b24 b25 │            │ 1 │ 1 │ 1 │ 1 │ 1 │ 1 │
│ b30 b31 b32 b33 b34 b35 │            ├───┼───┼───┼───┼───┼───┤
│ b40 b41 b42 b43 b44 b45 │            │ 4 │ 4 │ 4 │ 4 │ 4 │ 4 │ k1=1
│ b50 b51 b52 b53 b54 b55 │            │ × │ × │ × │ × │ × │ × │
│ b60 b61 b62 b63 b64 b65 │            │ 1 │ 1 │ 1 │ 1 │ 1 │ 1 │
│ b70 b71 b72 b73 b74 b75 │            ├───┼───┼───┼───┼───┼───┤
│ b80 b81 b82 b83 b84 b85 │            │ 2 │ 2 │ 2 │ 2 │ 2 │ 2 │ k1=2
│ b90 b91 b92 b93 b94 b95 │            │+p │+p │+p │+p │+p │+p │ (padded)
│  0   0   0   0   0   0  │ (padded)   │ × │ × │ × │ × │ × │ × │
│  0   0   0   0   0   0  │ (padded)   │ 1 │ 1 │ 1 │ 1 │ 1 │ 1 │
└─────────────────────────┘            └───┴───┴───┴───┴───┴───┘
    (padded to 12×6)                    n1=6 blocks (one per column)

Packed Layout in Memory (column-major within each 4×1 block):
Block(0,0): [b00 b10 b20 b30]  rsb1 →  Block(0,1): [b01 b11 b21 b31]  ...
            ↓ rsb0=1                                ↓ rsb0=1
            csb1=4 →                                csb1=4 →
Block(1,0): [b40 b50 b60 b70]          Block(1,1): [b41 b51 b61 b71]  ...
            ↓ rsb0=1                                ↓ rsb0=1
            csb1=4 →                                csb1=4 →
Block(2,0): [b80 b90  0   0 ]          Block(2,1): [b81 b91  0   0 ]  ...
            ↓ rsb0=1                                ↓ rsb0=1

Memory: [b00 b10 b20 b30][b01 b11 b21 b31][b02 b12 b22 b32]...
        [b40 b50 b60 b70][b41 b51 b61 b71][b42 b52 b62 b72]...
        [b80 b90  0   0 ][b81 b91  0   0 ][b82 b92  0   0 ]...
```

peichialin (approved):

france5289 (approved):
Thanks, @myeh01 , for the robust changes to the VQDOT GEMM kernel!
I do have a small question and suggestion regarding the two implementations for handling potential A buffer misalignment issues in the VQDOT microkernel.
In this PR, where you implemented two versions, is certainly more general as it adapts to both aligned and unaligned A buffers. Given this, it seems we could potentially simplify the codebase by keeping only the more general implementation.
My guess is that you might have kept both versions for performance reasons?
(Maybe the `memcpy` will use more instructions.)
Perhaps in the future, we could try running some benchmarks to definitively decide whether to retain both versions or stick solely to the more general one.

myeh01:
> Thanks, @myeh01 , for the robust changes to the VQDOT GEMM kernel! I do have a small question and suggestion regarding the two implementations for handling potential A buffer misalignment issues in the VQDOT microkernel. In this PR, where you implemented two versions, is certainly more general as it adapts to both aligned and unaligned A buffers. Given this, it seems we could potentially simplify the codebase by keeping only the more general implementation. My guess is that you might have kept both versions for performance reasons? (Maybe the `memcpy` will use more instructions.) Perhaps in the future, we could try running some benchmarks to definitively decide whether to retain both versions or stick solely to the more general one.

@france5289 Good question! Here are some benchmark numbers I got for a 128x128x256 problem:
```
Aligned version: 1.0x time 
General version: 2.7x aligned time
```

So you are right. Currently, we need to keep the aligned version because it has the best performance. I did not try to optimize the general version that much. I tried a few things, but @Harlen126's suggestion to use `memcpy` turned out to be the best. We could probably improve the performance of the general version even more, but I'm not sure that it will ever get similar performance to the aligned version because each `lw` from `a` must be replaced by multiple scalar instructions in the general version. But yes, it would be nice to have just one version to simplify the codebase.

ericlove (commented):
Sorry was in the middle of reviewing when this was merged. Looks great! Just posting these comments here to be addressed in future versions.

Harlen126:
> So you are right. Currently, we need to keep the aligned version because it has the best performance. I did not try to optimize the general version that much. I tried a few things, but @Harlen126's suggestion to use memcpy turned out to be the best. We could probably improve the performance of the general version even more, but I'm not sure that it will ever get similar performance to the aligned version because each lw from a must be replaced by multiple scalar instructions in the general version. But yes, it would be nice to have just one version to simplify the codebase.

@myeh01,  I found that using 64-bit loads is a better approach than `memcpy` on 64-bit systems, as the compiler generates fewer scalar instructions with this method. See https://godbolt.org/z/WYndMr3r5

Additionally, I discovered that for the general implementation, we may need to rewrite and reschedule the inline assembly to hide scalar latency. Using the 64-bit load strategy with intrinsics, I achieved only 20% performance degradation compared to the "aligned" implementation, whereas using untuned inline assembly resulted in 40% degradation.

PR #6, opened 2025-12-09
Fix clang-format step in CI

myeh01:
This PR fixes the clang-format step in CI so it will work post-merge as well. Originally, for a PR, clang-format would only run on files that were changed. This PR modifies it so that clang-format will simply run on all `.c` and `.h` files.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -54,6 +54,7 @@ jobs:
         GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
     - name: clang-format
+      if: ${{ github.event_name == 'pull_request' }}
       run: |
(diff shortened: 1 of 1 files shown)


hdelassus:
Can we keep the optimized version (the one that checks only modified files)?

EDIT: Or even skip it entirely post-merge (is there a reliable way to detect it's post merge, e.g. if BASE_SHA or MERGE_SHA aren't defined?)

myeh01:
We can use it, but I think it will run `clang-format` on an empty list of files post-merge. Or maybe that's ok, and we only need to check pre-merge?

Edit: just saw your edit. Yes, I think we can skip the post-merge check then.
Edit 2: I think I found a simpler solution. The clang-format step now runs only if the event is a pull request.

hdelassus (approved):
Perfect fix ✨

PR #7, opened 2025-12-11
CODEOWNERS update

Aaron-Hutchinson:
Populating `CODEOWNERS` with myself.

~I've removed `* @ericlove` (CC @ericlove) because, based on my understanding of how a codeowners file works, that line conflicts with those I've added since only one codeowner line is selected for a given file.~

I ordered the lines in the way they appear on GitHub (alphabetically, but with directories first), and used newlines to group large categories (`src` and `test`).

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,3 +1,5 @@
+* @ericlove
+
 /cmake @Aaron-Hutchinson
 /doc @Aaron-Hutchinson
 /include @Aaron-Hutchinson
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,1 +1,13 @@
-* @ericlove
+/cmake @Aaron-Hutchinson
+/doc @Aaron-Hutchinson
+/include @Aaron-Hutchinson
+
+/src/gemm/rvv @Aaron-Hutchinson
+/src/gemm/scalar @Aaron-Hutchinson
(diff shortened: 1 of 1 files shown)


ericlove (dismissed):
Approving, since it looks like this needs my approval as a codeowner to remove me as a codeowner...

.github/CODEOWNERS
```diff
@@ -1 +1,13 @@
-* @ericlove
```
pvchupin, line 1: I think you should keep it above all lines you've just added.
That will effectively means that everything uncovered will still go to Eric.

pvchupin (changes requested):

.github/CODEOWNERS
pvchupin: https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners#example-of-a-codeowners-file:~:text=%23%20These%20owners%20will%20be%20the%20default%20owners%20for%20everything%20in%0A%23%20the%20repo.%20Unless%20a%20later%20match%20takes%20precedence%2C%0A%23%20%40global%2Downer1%20and%20%40global%2Downer2%20will%20be%20requested%20for%0A%23%20review%20when%20someone%20opens%20a%20pull%20request.%0A*%20%20%20%20%20%20%20%40global%2Downer1%20%40global%2Downer2

pvchupin (approved):

PR #11, opened 2026-01-15
Add a shared benchmark macro to be used by kernel tests

hdelassus:
# What this PR does

It defines three macros:
* `SKL_BENCHMARK_RUN`, to replace benchmark boilerplate and report performance in a unified way.
* `SKL_TEST_PERF_REPORT`, to allow customization of the performance reporting function called by `SKL_BENCHMARK_RUN`. By default set to `report_perf_epc`.
* `SKL_TEST_WARMUP`, to be used by benchmarks to request a warmup run before making any measurements. By default set to `1` (true).

As a test and usage example, the gelu_f32 test replaces the performance measurement and reporting boilerplate by the `SKL_BENCHMARK_RUN` macro.

This is purely a refactoring PR: no functional change is expected. More details follw.

## Shared benchmark macro
In @ericlove 's
> Assuming most benchmarks have a common pattern (some "number of elements", a need to print cycle/instruction counts, and throughput based on size), this PR introduces the `SKL_BENCHMARK_RUN` macro:

```c
/**
 * @brief Run a benchmark and report performance.
 *
 * @param NAME - The name of the benchmark.
 * @param NUM_ELEMS - The number of elements (application-specific).
 * @param WARMUP - Whether to run a warmup iteration.
 * @param FUNC - The function to benchmark (will be applied to __VA_ARGS__).
 * @param ... - The arguments to pass to the function.
 *
 * This macro is a wrapper around riscv_read_mcycle() and
 * riscv_read_minstret() to measure the performance of a function.
 *
 * If WARMUP is true, run the function once before timing to warm up caches.
 * Usually WARMUP is set to SKL_TEST_WARMUP, which is 1 by default.
 *
 * If ENABLE_BENCHMARK is not defined, this macro does nothing.
 */
#if defined(ENABLE_BENCHMARK)
#define SKL_BENCHMARK_RUN(NAME, NUM_ELEMS, WARMUP, FUNC, ...)                  \
```

> It uses the `__VA_ARGS__` feature to forward all necessary arguments to the function-under-test, but prepends them with an element count (defined however makes sense for the benchmark) a name for use in the printout, and a boolean switch to enable/disable warmup (mostly set to a configurable macro `SKL_TEST_WARMUP` analogous to `SKL_TEST_MIN_F32` etc.).

## Common Benchmark Output Format

The `SKL_BENCHMARK_RUN` macro calls a configurable reporting function, set by `SKL_TEST_PERF_REPORT`. It defaults to the existing `report_perf_epc`, whose signature _is planned to be_ expanded to match that of the outer macro.

The configurability will allow SKL users to adjust the output as necessary.

# Motivation

The goals of this effort are twofold:
* to clean-up and unify SKL kernel benchmarks.
* to provide a way for users of the SKL project to customize performance reporting.

# How this was tested

The new macro replaces the boilerplate in the FP32 GELU test as a first example of how it may be used in other SKL kernel tests. Concretely this PR can be tested with the following commands:

```shell
cmake -B build -G Ninja --toolchain cmake/riscv.cmake -DSKL_BUILD_BENCHMARKS=ON
cmake --build build
ctest --test-dir build -R gelu_f32 -V
```

### DIFF
--- a/test/gelu/gelu_f32.c
+++ b/test/gelu/gelu_f32.c
@@ -145,21 +145,6 @@ int main(void) {
   printf("Measuring %d-element GELU:\n", NUM_ELEMS);
   init_random(input, NUM_ELEMS); // input data
 
-#if defined(ENABLE_BENCHMARK)
-  // cycle and instruction counts
-  uint64_t c0, c1, i0, i1; // NOLINT(*-isolate-declaration)
-#define MEASURE_PERF(FUNCTION, NAME)                                           \
-  /* Measure 2nd run after caches warmed */                                    \
-  riscv_fence();                                                               \
-  c0 = riscv_read_mcycle(), i0 = riscv_read_minstret();                        \
-  FUNCTION(output, input, NUM_ELEMS);                                          \
-  riscv_fence();                                                               \
-  c1 = riscv_read_mcycle(), i1 = riscv_read_minstret();                        \
-  report_perf_epc(NAME, c1 - c0, i1 - i0, NUM_ELEMS);
-#else
-#define MEASURE_PERF(FUNCTION, NAME)
-#endif
-
 #if defined(ENABLE_TEST)
   memset(ref_output, 0, NUM_ELEMS * sizeof(*ref_output));
   reference_gelu_f32(ref_output, input, NUM_ELEMS);
@@ -173,7 +158,8 @@ int main(void) {
 #define RUN_(FUNCTION, NAME, TOL, GEN2, GEN1, GE0, SPECIALS)                   \
   memset(output, 0, NUM_ELEMS * sizeof(*output));                              \
   FUNCTION(output, input, NUM_ELEMS);                                          \
-  MEASURE_PERF(FUNCTION, NAME);                                                \
+  SKL_BENCHMARK_RUN(NAME, NUM_ELEMS, SKL_TEST_WARMUP, FUNCTION, output, input, \
+                    NUM_ELEMS);                                                \
   CHECK_RESULT(FUNCTION, NAME, TOL, GEN2, GEN1, GE0, SPECIALS);
 
 #define PASTE3(a, b, c) a##b##c
--- a/test/gelu/gelu_f32.c
+++ b/test/gelu/gelu_f32.c
@@ -157,7 +157,6 @@ int main(void) {
 
 #define RUN_(FUNCTION, NAME, TOL, GEN2, GEN1, GE0, SPECIALS)                   \
   memset(output, 0, NUM_ELEMS * sizeof(*output));                              \
-  FUNCTION(output, input, NUM_ELEMS);                                          \
   SKL_BENCHMARK_RUN(NAME, NUM_ELEMS, SKL_TEST_WARMUP, FUNCTION, output, input, \
                     NUM_ELEMS);                                                \
   CHECK_RESULT(FUNCTION, NAME, TOL, GEN2, GEN1, GE0, SPECIALS);
--- a/test/skl-test.h
+++ b/test/skl-test.h
@@ -13,7 +13,8 @@
 
 /* Extract the test name from SKL_TEST_NAME as a string. */
 #define SKL_TEST_NAME_STR(NAME) #NAME
-#define DECL_SKL_TEST_NAME(NAME) const char *skl_test_name = SKL_TEST_NAME_STR(NAME);
+#define DECL_SKL_TEST_NAME(NAME)                                               \
+  const char *skl_test_name = SKL_TEST_NAME_STR(NAME);
 DECL_SKL_TEST_NAME(SKL_TEST_NAME)
 
 #if __riscv_xlen == 32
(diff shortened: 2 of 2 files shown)


test/skl-test.h
```diff
@@ -60,6 +68,70 @@ static inline void report_perf_epc(const char *name, uint64_t cycles,
...
+ */
+#if defined(ENABLE_BENCHMARK)
+#define SKL_BENCHMARK_RUN(NAME, NUM_ELEMS, WARMUP, FUNC, ...)                  \
+  do {                                                                         \
+    if (WARMUP) {                                                              \
+      FUNC(__VA_ARGS__);                                                       \
+    }                                                                          \
+    riscv_fence();                                                             \
+    uint64_t c0 = riscv_read_mcycle();                                         \
+    uint64_t i0 = riscv_read_minstret();                                       \
```
ebavier, line 122: Would it make any sense to read the cycle counter *after* the instruction counter?  I believe that way the total cycle count might be slightly more accurate?

hdelassus: Agreed, thanks for catching this. I'll make the change.

hdelassus: But that's at the expense of making the total instret count slightly more inaccurate 🙃

hdelassus: Do you think I should still make the change?

ebavier: Just a feeling, we typically are asked more often about cycle counts. 🤷

hdelassus: Right :)

test/gelu/gelu_f32.c
```diff
@@ -173,7 +158,8 @@ int main(void) {
 #define RUN_(FUNCTION, NAME, TOL, GEN2, GEN1, GE0, SPECIALS)                   \
   memset(output, 0, NUM_ELEMS * sizeof(*output));                              \
   FUNCTION(output, input, NUM_ELEMS);                                          \
-  MEASURE_PERF(FUNCTION, NAME);                                                \
+  SKL_BENCHMARK_RUN(NAME, NUM_ELEMS, SKL_TEST_WARMUP, FUNCTION, output, input, \
+                    NUM_ELEMS);                                                \
```
Aaron-Hutchinson, line 161: Shouldn't the `FUNCTION` call on line 160 be removed since the function and warmup are called in `SKL_BENCHMARK_RUN`?

test/skl-test.h
```diff
@@ -60,6 +68,70 @@ static inline void report_perf_epc(const char *name, uint64_t cycles,
...
+    riscv_fence();                                                             \
+    uint64_t i0 = riscv_read_minstret();                                       \
+    uint64_t c0 = riscv_read_mcycle();                                         \
+    FUNC(__VA_ARGS__);                                                         \
+    riscv_fence();                                                             \
+    uint64_t c1 = riscv_read_mcycle();                                         \
+    uint64_t i1 = riscv_read_minstret();                                       \
+    uint64_t cycles = c1 - c0;                                                 \
+    uint64_t insts = i1 - i0;                                                  \
+    SKL_TEST_PERF_REPORT(NAME, cycles, insts, (size_t)(NUM_ELEMS));            \
```
Aaron-Hutchinson, line 130: Wouldn't it be better to separate performance reporting from function execution? Including this line in `SKL_BENCHMARK_RUN` feels a bit limiting due to the fixed function signature that `SKL_TEST_PERF_REPORT` requires.

Is the motivation here just to save the `report_perf_epc(...)` line in each harness?

test/gelu/gelu_f32.c
hdelassus: If benchmark is disabled but test is enabled, the function needs to be called once before `CHECK_RESULT`.
More logic could be done to avoid redundant calls though.

test/skl-test.h
hdelassus: Yes, arguably one could ask the same for `CHECK_RESULT` since testing and result checking are separate there.
How would separating benchmark and reporting lift limitations on the report signature though?

test/gelu/gelu_f32.c
ebavier: How about, if `!ENABLE_BENCHMARKING`, the `SKL_BENCHMARK_RUN` macro calls `FUNC(__VA_ARGS__)`?

Aaron-Hutchinson: Seems like the test logic should be governed by a separate macro like `SKL_TEST_RUN`. But this would be outside of this PR's scope, so feel free to resolve this conversation.

test/skl-test.h
Aaron-Hutchinson: I just meant that if the report printing is separated from `SKL_BENCHMARK_RUN`, it doesn't seem like there would be any need for the `SKL_TEST_PERF_REPORT` macro and you could just call the desired report function directly (using whatever signature it might have).

Maybe I'm missing the point here.

test/skl-test.h
hdelassus: I kept the change.

test/gelu/gelu_f32.c
hdelassus: @Aaron-Hutchinson indeed, thanks.
I've used @ebavier 's suggestion for now.

test/skl-test.h
hdelassus: The point is to allow users to change the performance reporting function behavior without changing the SKL source code by defining their own `SKL_TEST_PERF_REPORT` macro.

ebavier (approved):
LGTM.  Should we expect follow-up PRs to adapt tests for other kernels?

hdelassus:
> LGTM. Should we expect follow-up PRs to adapt tests for other kernels?

Yes!

PR #14, opened 2026-01-22
Support alpha/beta scaling in f32 Xsfmm GEMM kernels

myeh01:
This PR adds support for alpha/beta scaling in the f32 Xsfmm GEMM kernels using the fused kernel API.


src/gemm/xsfmm/gemm_f32c_f32_f32_xsfmm32a32f.c
```diff
@@ -0,0 +1,720 @@
...
+    size_t rsc1, size_t csc1, size_t row, size_t col, void *params) {
+  alpha_beta_f32_f32 *params_cast = (alpha_beta_f32_f32 *)params;
+  float alpha = params_cast->alpha;
+  float beta = params_cast->beta;
+  float *c_tile = c + row * rsc1 + col * csc1;
+
+  size_t te = 0;
+  __asm__ volatile("sf.vsettnt %0, x0, e32, w1" : "=r"(te) : : "vtype", "vl");
+
+  if (te <= (size_t)2 * __riscv_v_min_vlen / (8 * sizeof(float)) && csc0 == 1) {
```
ericlove: I think this should be set dynamically based on machine capabilities, otherwise it doesn't seem like the condition will be satisfied by most machines' TE:
```suggestion
  if (te <= __riscv_vsetvlmax_e32m2() && csc0 == 1) {
```
(Maybe I did my math wrong, but if `__riscv_v_min_vlen = 128`, then the RHS will evaluate to 2*128/32=8, which is much smaller than any TE value I've seen.)

src/gemm/xsfmm/gemm_f32c_f32_f32_xsfmm32a32f.c
```diff
@@ -0,0 +1,720 @@
...
+          [tn] "r"(tn)
+        : "v0", "v1", "v2", "v3", "v4", "v5", "v6", "v7", "v16", "v17", "v18",
+          "v19", "v20", "v21", "v22", "v23", "vtype", "vl", "memory");
+  }
+}
+
+/* Computes C := alpha * tile + beta * C, where tile is the tm x tn tile
+ * specified by tss. tm and tn must be <= TE. This is an optimized
+ * implementation when tile rows fit into m2 register groups and csc0 == 1.
+ * In this case, the availability of extra register groups allows us to
```
ericlove: I would make this slightly more precise (it's not clear what "extra" refers to as a baseline): "the availability of 16 register groups allows unrolling the scaling loop to that depth and software pipelining the tile row transfers to overlap their latency with the scaling computation" or something like that.

src/gemm/xsfmm/gemm_f32c_f32_f32_xsfmm32a32f.c
```diff
@@ -0,0 +1,720 @@
...
+#include <stddef.h>
+
+#include "skl-common.h"
+
+typedef struct {
+  float alpha;
+  float beta;
+} alpha_beta_f32_f32;
+
+typedef void (*fused_ker_f32_f32_t)(size_t tm, size_t tn, size_t tss, float *c,
```
ericlove: Can you add a brief explanatory comment to document this interface?

ericlove (commented):
Very nice contribution! This is a welcome improvement to interface consistency and functionality.  Some minor initial comments for now pending internal review.

ericlove:
> the fused kernel API.

Actually, I think we should probably PR a description of this API as a markdown document in [src/gemm/xsfmm](https://github.com/sifive/skl/tree/main/src/gemm/xsfmm) before this is merged.

src/gemm/xsfmm/gemm_f32c_f32_f32_xsfmm32a32f.c
myeh01: Thanks! Your math looks right. This works much better when `__riscv_v_min_vlen` is set strictly less than the machine's vlen.

src/gemm/xsfmm/gemm_f32c_f32_f32_xsfmm32a32f.c
```diff
@@ -0,0 +1,721 @@
...
+
+#include "skl-common.h"
+
+// params type for the fused alpha/beta scaling kernel
+typedef struct {
+  float alpha;
+  float beta;
+} alpha_beta_f32_f32;
+
+typedef void (*fused_ker_f32_f32_t)(size_t tm, size_t tn, size_t tss, float *c,
```
Aaron-Hutchinson: What does each `f32` in this name refer to?

It seems strange that there are two type fields in the name but (implicitly) three typed parameters: `c`, `alpha`, `beta`.

src/gemm/xsfmm/gemm_f32c_f32_f32_xsfmm32a32f.c
```diff
@@ -0,0 +1,721 @@
...
+  size_t tss_1 = tss_0 + kRowInc;
+  float *c_load_0 = c;
+  float *c_load_1 = c_load_0 + rsc0;
+  float *c_store_0 = c;
+  float *c_store_1 = c_store_0 + rsc0;
+
+  /* Process (tm / 8) * 8 rows. */
+  size_t i = (tm / 8) * 8;
+  __asm__ volatile(
+      "sf.vsettnt x0, %[tn], e32, w1\n"
```
Aaron-Hutchinson, line 311: Sorry for my ignorance on some of the xsfmm details... I don't see where lmul=2 is being set in this function (unless `vsettnt` is doing it, but the docs don't seem to indicate so?)

src/gemm/xsfmm/gemm_f32c_f32_f32_xsfmm32a32f.c
```diff
@@ -0,0 +1,721 @@
...
+ * the names of alpha-beta scaling functions include the input type of the
+ * corresponding GEMM even though they do not depend on it.
+ */
+SKL_XSFMM_IN
+SKL_FUNC_PRIVATE void skl_gemm_alpha_beta_scaling_f32_f32_xsfmmbase(
+    size_t tm, size_t tn, size_t tss, float *c, size_t rsc0, size_t csc0,
+    size_t rsc1, size_t csc1, size_t row, size_t col, void *params) {
+  alpha_beta_f32_f32 *params_cast = (alpha_beta_f32_f32 *)params;
+  float alpha = params_cast->alpha;
+  float beta = params_cast->beta;
```
Aaron-Hutchinson: I probably missed the discussion on the design here. It seems strange that the public API uses alpha/beta parameters explicitly, while the private functions use a `void*` to be cast into an `alpha_beta_f32_f32` struct just to unwrap the values into floats anyway. What's the motivation here?

src/gemm/xsfmm/gemm_f32c_f32_f32_xsfmm32a32f.c
myeh01: One is for the "input" type and the other is the output type for the corresponding GEMM kernel. The input type is probably not needed for the kernel itself, but is added to avoid potential typedef conflicts if SKL_FUNC_PRIVATE is not `static`.

src/gemm/xsfmm/gemm_f32c_f32_f32_xsfmm32a32f.c
myeh01: LMUL is set implicitly through the `vsettnt` instruction. It's explicitly checked in the dispatch code [here](https://github.com/sifive/skl/blob/6cc4974b074844b06a976020eada61e38ce283ce/src/gemm/xsfmm/gemm_f32c_f32_f32_xsfmm32a32f.c#L412). So you're right, this code could set LMUL to something > 2, but the dispatch code ensures it is called only when it's <= 2.

src/gemm/xsfmm/gemm_f32c_f32_f32_xsfmm32a32f.c
myeh01: Besides alpha/beta scaling, we also want GEMM kernels that perform other post-GEMM operations like bias+clamp, applying an activation function, etc. The internal GEMM kernel APIs have been modified to include a pointer to a post-GEMM operator, so users can get all these GEMM varieties simply by passing in a different pointer. The post-GEMM operators all have different function signatures. The only parameters they have in common would relate to the `C` matrix. Any parameters particular to a specific operator have to be passed through a struct.

I'll be PRing some documentation for this fused kernel API. You are welcome to make any suggestions for improvements.

Aaron-Hutchinson: Thanks, that mostly makes sense, except I'm not seeing why the uniform `void*` interface applies only to the private function and not the public ones. But maybe my confusion will be cleared in the documentation PR, so I'm happy to move the discussion there.

myeh01: For now, the public APIs are basically just user-friendly wrappers for the `void*` interface. We do want to provide users with the standard LAPACK GEMM API, for example. The `void*` interface makes it easier for us to reuse the inner loop code internally among all the GEMM variants. Eventually, we will probably make the `void*` interface public so users can write their own custom operators.

src/gemm/xsfmm/gemm_f32c_f32_f32_xsfmm32a32f.c
```diff
@@ -0,0 +1,721 @@
...
+
+  kernel(tm, tn, mt0, c, rsc0, 1, rsc1, csc1, 0, 0, params);
+  kernel(tm, tn, mt4, c, rsc0, 1, rsc1, csc1, 0, 1, params);
+  kernel(tm, tn, mt8, c, rsc0, 1, rsc1, csc1, 1, 0, params);
+  kernel(tm, tn, mt12, c, rsc0, 1, rsc1, csc1, 1, 1, params);
+
+  __asm__ volatile("sf.vtdiscard");
+}
+
+SKL_XSFMM_OUT
```
peichialin: Shall we mark the private fused gemm functions with `SKL_XSFMM_NEW` the same as the public GEMM APIs since they actually end up with writing results to memory (by calling `kernel` inside), not mammoth's tile registers?

src/gemm/xsfmm/gemm_f32c_f32_f32_xsfmm32a32f.c
```diff
@@ -0,0 +1,721 @@
...
+
+  const size_t mt0 = 0;
+  const size_t mt4 = (size_t)(4) << kShiftTile;
+  const size_t mt8 = (size_t)(8) << kShiftTile;
+  const size_t mt12 = (size_t)(12) << kShiftTile;
+
+  kernel(tm, tn, mt0, c, rsc0, 1, rsc1, csc1, 0, 0, params);
+  kernel(tm, tn, mt4, c, rsc0, 1, rsc1, csc1, 0, 1, params);
+  kernel(tm, tn, mt8, c, rsc0, 1, rsc1, csc1, 1, 0, params);
+  kernel(tm, tn, mt12, c, rsc0, 1, rsc1, csc1, 1, 1, params);
```
peichialin: If the kernel is intended to handle a single tile (specfied by `tss`), then it is simpler to pre-calculate the c_tile location based on `row` and `col` index with `rsc1` and `csc1` values.

In other words, we can simplify
```
typedef void (*fused_ker_f32_f32_t)(size_t tm, size_t tn, size_t tss, float *c,
                                    size_t rsc0, size_t csc0, size_t rsc1,
                                    size_t csc1, size_t row, size_t col,
                                    void *params);
```
as 
```
typedef void (*fused_ker_f32_f32_t)(size_t tm, size_t tn, size_t tss, float *c,
                                    size_t rsc0, size_t csc0,
                                    void *params);
```
and modify the caller side accordingly.

src/gemm/xsfmm/gemm_f32c_f32_f32_xsfmm32a32f.c
myeh01: I think `OUT` is correct here because the private fused gemm functions modify the tile state. We cannot use `NEW` because nested `NEW`s are prohibited, and these private functions will be called within functions marked `NEW`.

src/gemm/xsfmm/gemm_f32c_f32_f32_xsfmm32a32f.c
myeh01: We need to rework the API somewhat because of an issue related to this. The reason we need `row` and `col` (or something like them) is if the fused kernel uses an array (such as a bias vector). A pointer to the array (e.g. `float *bias`) is passed to the fused kernel via `params`, but this pointer always points to the beginning of the entire array. The fused kernel needs `row` and `col` so it can calculate the correct offset into this array.

src/gemm/xsfmm/gemm_f32c_f32_f32_xsfmm32a32f.c
peichialin: Thanks for the clarification.
So we should interpret `SKL_XSFMM_OUT` as "the function modifies the Mammoth's tile states", but not indicate the destination of the function's output.

src/gemm/xsfmm/gemm_f32c_f32_f32_xsfmm32a32f.c
peichialin: This indeed a challenge!
I even found that providing `row` and `col` indices of a "block" (2tm2tn or 1tm1tn)  is not sufficient if there are multiple blocks in the original `c` (`m`x`n`) -- The `bias` address is associated with the original `c`, not the block this function is processing.

src/gemm/xsfmm/gemm_f32c_f32_f32_xsfmm32a32f.c
myeh01: Yes, that's my understanding.

src/gemm/xsfmm/gemm_f32c_f32_f32_xsfmm32a32f.c
myeh01: Yes, that's exactly the problem I encountered. We need to modify the API so the fused kernel knows the block's position in the original `c` matrix. I hope to have a proposal soon.

myeh01:
I have made substantial updates to this PR. I think the last thing I need to do is add a README for the fused API; otherwise, this should be ready for review.

src/gemm/xsfmm/gemm_f32c_f32_f32_xsfmm32a32f.c
```diff
@@ -0,0 +1,1378 @@
...
+        "vse32.v v16, (%[c])\n"
+        "add %[c], %[c], %[rsc0]\n"
+        "addi %[i], %[i], -1\n"
+        "bnez %[i], 0b"
+        : [tss] "+&r"(tss), [c] "+&r"(c), [i] "+&r"(i)
+        : [alpha] "f"(alpha), [beta] "f"(beta), [kRowInc] "rI"(kRowInc),
+          [rsc0] "r"(rsc0 * sizeof(float)), [tn] "r"(tn)
+        : "v0", "v1", "v2", "v3", "v4", "v5", "v6", "v7", "v16", "v17", "v18",
+          "v19", "v20", "v21", "v22", "v23", "vtype", "vl", "memory");
+  } else {
```
ericlove, line 257: It seems like a future optimization we could make is to support column-major C matrices (or fusion of transposition to the output) by applying the scaling to columns of the tile registers, rather than rows.  Not suggesting that for this PR, but want to keep it in our back pocket.

src/gemm/xsfmm/gemm_f32c_f32_f32_xsfmm32a32f.c
```diff
@@ -0,0 +1,1378 @@
...
+  }
+}
+
+/* Computes C := alpha * tile + beta * C, where tile is the tm x tn tile
+ * specified by tss. tm and tn must be <= ETE. This is an optimized
+ * implementation when tile rows fit into m2 register groups and csc0 == 1.
+ * In this case, the availability of 16 register groups allows unrolling the
+ * scaling loop by a factor of 8 and software pipelining the tile row transfers
+ * to overlap their latency with the scaling computation.
+ */
```
ericlove, line 291: Can we try to make these bock comments a bit easier to read for key details? No need to be full-on Doxygen (they aren't public APIs, at least not yet), but maybe something like:
```suggestion
/* Computes C := alpha * tile + beta * C where:
 *  - tile is the tm x tn tile specified by tss
 *  - tm and tn must be <= ETE
 *  - csc0 == 1
 *  - tile rows fit into LMUL=2 register groups
 * 
 * The last condition must also be checked by the caller.
 * 
 * In this case, the availability of 16 register groups allows unrolling the
 * scaling loop by a factor of 8 and software pipelining the tile row transfers
 * to overlap their latency with the scaling computation.
```
Or use whatever style you prefer, as long as it breaks up the big chunk.

src/gemm/xsfmm/gemm_f32c_f32_f32_xsfmm32a32f.c
```diff
@@ -0,0 +1,1378 @@
...
+
+#include <riscv_vector.h>
+#include <stddef.h>
+
+#include "skl-common.h"
+
+typedef void (*fused_f32_f32_t)(size_t tm, size_t tn, size_t tss, float *c,
+                                size_t rsc0, size_t csc0, size_t rsc1,
+                                size_t csc1, size_t row1, size_t col1,
+                                void *params);
```
ericlove: This function pointed to by this type needs to have the `SKL_XSFMM_IN` attribute applied, and I think that belongs at the end:
```suggestion
                                void *params) SKL_XSFMM_IN;
```
Asking @topperc and @4vtomat for help here.

src/gemm/xsfmm/gemm_f32c_f32_f32_xsfmm32a32f.c
```diff
@@ -0,0 +1,1378 @@
...
+  }
+
+  const size_t kShiftPattern = 24;
+  const size_t mt0 = 0;
+  const size_t mt0c = mt0 | 1 << kShiftPattern;
+  if (accum) {
+    skl_tile_load_f32rcp_f32_xsfmmbase(m, n, c + row1 * rsc1 + col1 * csc1,
+                                       rsc0, rsc1, csc1, mt0c, 4, 0);
+  } else {
+    skl_tile_zero_f32_f32_xsfmmbase(n, m, mt0, 0, 4);
```
ericlove: Is it really the case that for "column tiles" we swap `m` and `n` for both tile_load and tile_zero, but only swap `rstss` and `cstss` for tile_load (and not for tile_zero)?  And then we  swap `rstss` and `cstss` for application of the fused kernel but _not_ `m` and `n`? It would be helpful to spell this all out in one place.

src/gemm/xsfmm/gemm_f32c_f32_f32_xsfmm32a32f.c
```diff
@@ -0,0 +1,1378 @@
...
+
+  skl_gemm_fused_apply_f32_f32rcprc_xsfmm32a32f(
+      m, n, mt0c, 4, 0, c, rsc0, 1, rsc1, csc1, row1, col1, kernel, params);
+
+  __asm__ volatile("sf.vtdiscard");
+}
+
+/* m <= ETE, n <= 3 * ETE. */
+SKL_XSFMM_NEW
+SKL_FUNC_PRIVATE void skl_gemm_1x3_f32c_f32rcp_f32rcp_xsfmm32a32f(
```
ericlove: I think the multiplicity and repetitiveness of these functions (wrappers around each tiling's inner loop + load/zero and alpha/beta), should be reduced, which would also help make the logic behind each set of arguments more obvious

What I'd like to write is something along the lines of (code below not fully correct/needs more cleanup):
```c++
typedef (*inner_loop_func_t)(size_t m, size_t n, size_t k, const float *a,
                           size_t rsa1, size_t csa1, const float *b,
                           size_t rsb1, size_t csb1);

#define SKL_FULL_GEMM_PARAMS_XSFMM32A32F \
    size_t m, size_t n, size_t k, const float *a, size_t rsa1, size_t csa1, \
    const float *b, size_t rsb1, size_t csb1, float *c, size_t rsc0, \
    size_t rsc1, size_t csc1, bool accum, fused_f32_f32_t kernel, \
    void *params


SKL_XSFMM_NEW SKL_FUNC_PRIVATE void skl_apply_gemm_tiling_xsfmm32a32f(
    size_t m1, size_t n1,
    inner_loop_func_t inner_loop,
    size_t row1, size_t col1,
    SKL_FULL_GEMM_PARAMS_XSFMM32A32F) {
  if (m == 0 || n == 0) {
    return;
  }

  // For {4,3,2}x1 tilings, swap A and B and transpose output blocks
  const bool trans = m1 > n1;
  const size_t kShiftPattern = 24;
  const size_t mt0 = 0;
  const size_t mt0c = trans ? (mt0 | 1 << kShiftPattern) : mt0;
  size_t rstss = 0;
  size_t cstss = (m1 > 0 && n1> 0) ? 8 : 4;
  if (accum) {
    skl_tile_load_f32rcp_f32_xsfmmbase(m, n, c + row1 * rsc1 + col1 * csc1,
                                       rsc0, rsc1, csc1, mt0c,
                                       trans ? cstss : rstss,
                                       trans ? rstss : cstss);
  } else {
    skl_tile_zero_f32_f32_xsfmmbase(trans ? n : m, trans ? m : n, mt0, rstss, cstss);
  }

  // NOLINTBEGIN(readability-suspicious-call-argument)
  if (trans) {
    inner_loop(n, m, k, b, rsb1, csb1, a, rsa1, csa1);
  } else {
    inner_loop(m, n, k, a, rsa1, csa1, b, rsb1, csb1);
  }
  // NOLINTEND(readability-suspicious-call-argument)

  skl_gemm_fused_apply_f32_f32rcprc_xsfmm32a32f(
      m, n, mt0c, 4, 0, c, rsc0, 1, rsc1, csc1, row1, col1, kernel, params);

  __asm__ volatile("sf.vtdiscard");
}
```

src/gemm/xsfmm/gemm_f32c_f32_f32_xsfmm32a32f.c
```diff
@@ -0,0 +1,1378 @@
...
+      params);
+}
+
+SKL_FUNC_PRIVATE void skl_gemm_fused_f32pc_f32cp_f32rcp_xsfmm32a32f(
+    size_t m1, size_t n1, size_t k, const float *a, size_t rsa1, const float *b,
+    size_t csb1, float *c, size_t rsc1, size_t csc1, bool accum,
+    fused_f32_f32_t kernel, void *params) {
+  size_t ete = 0; // Effective tile edge length (always TE for TEW = 32).
+  __asm__ volatile("sf.vsettnt %0, x0, e32, w1" : "=r"(ete) : : "vtype", "vl");
+
```
ericlove: I don't see much utility in these wrapper functions.  Let's just used the "dispatch" function directly, and rename it to `skl_gemm_fused_f32pc_f32cp_f32rcp_xsfmm32a32f`?  If clients want a generic fused-gemm, they can be forced to use the packed API like we do internally anyways.

ericlove (changes requested):
Thanks for making this happen! I think it's a great overall improvement to the library to have consistent APIs across the "GEMM" kernels by dropping the `a1b01` constraint, and the introduction of the general mechanism for fused Xsfmm kernels is big step forward.  I look forward to your design description for that in a subsequent PR (best to split it up, I think).

However, I do have some concerns about the overall legibility and maintainability of the implementation behind that API, which needs to be reorganized.  Obviously there's no way that the big blocks of inline asm will look nice, but I think the hand-written logic around them suffers from too much repetitiveness and a kind of spaghetti argument passing style.

I've made a few suggestions below of how we might reduce the complexity, and improve the clarity of the dispatch into the assembly kernels.  But this is just a starting point.  I think it would be good to spend some time thinking carefully about how to regularize the structure of this file before we move onto other datatypes.

src/gemm/xsfmm/gemm_f32c_f32_f32_xsfmm32a32f.c
```diff
@@ -0,0 +1,1378 @@
...
+    for (; j1 + 1 < n1; j1 += 2) {
+      size_t n_vl = n_avl >= 2 * ete ? 2 * ete : n_avl;
+      skl_gemm_2x2_f32rcp_f32rcp_f32rcp_xsfmm32a32f(
+          2 * ete, n_vl, k, a + i1 * rsa1, rsa1, csa1, b + j1 * csb1, rsb1,
+          csb1, c, rsc0, rsc1, csc1, i1, j1, accum, kernel, params);
+      skl_gemm_2x2_f32rcp_f32rcp_f32rcp_xsfmm32a32f(
+          m_vl - 2 * ete, n_vl, k, a + (i1 + 2) * rsa1, rsa1, csa1,
+          b + j1 * csb1, rsb1, csb1, c, rsc0, rsc1, csc1, i1 + 2, j1, accum,
+          kernel, params);
+      n_avl -= n_vl;
```
ericlove, line 1175: I think there's too many arguments in these function calls, and it's hard to discern the subtle differences between each one.

It would be great if we could factor out much of the similarity into a macro, which might look something like (and assuming my [suggestion from above](https://github.com/sifive/skl/pull/14/changes#r3365435507)):
```c++
SKL_FUNC_PRIVATE void skl_gemm_fused_f32pc_f32cp_f32rcp_xsfmm32a32f(
    SKL_FULL_GEMM_PARAMS_XSFMM32A32F) {
  if (m == 0 || n == 0) {
    return;
  }

  #define SKL_GEMM_TILE(RXC, R, C, ROW1, COL1) \
    skl_apply_gemm_tiling_xsfmm32a32f( \
      (R), (C), \
      skl_gemm_inner_loop_##RXC##_f32rcp_f32rcp_f32_xsfmm32a32f, \
      ((i1)+ROW1), ((j1)+COL1), \
      m_vl-((ROW1) * ete), n_vl-((COL1) * ete), k, \
      a + ((i1)+ROW1) * rsa1, rsa1, csa1, \
      b + ((j1)+COL1) * csb1, rsb1, csb1, \
      c, rsc0, rsc1, csc1, accum, kernel, params)

  size_t ete = 0; // Effective tile edge length (always TE for TEW = 32).
  __asm__ volatile("sf.vsettnt %0, x0, e32, w1" : "=r"(ete) : : "vtype", "vl");

  size_t m1 = (m + ete - 1) / ete;
  size_t n1 = (n + ete - 1) / ete;
  size_t i1 = 0;
  size_t m_avl = m;
  for (; i1 + 3 < m1; i1 += 4) {
    size_t m_vl = SKL_MIN(m_avl, 2 * ete);
    size_t n_avl = n;
    size_t j1 = 0;
    for (; j1 + 1 < n1; j1 += 2) {
      size_t n_vl = SKL_MIN(n_avl, 2 * ete);
      SKL_GEMM_TILE(2x2, 2, 2, 0, 0);
      SKL_GEMM_TILE(2x2, 2, 2, 2, 0);
      n_avl -= n_vl;
    }
    if (j1 < n1) {
      SKL_GEMM_TILE(1x4, 4, 1, 0, 0);
    }
    m_avl -= m_vl;
  }
/// etc.

#undef SKL_GEMM_TILE
````

File: src/gemm/xsfmm/gemm_f32c_f32_f32_xsfmm32a32f.c (hunk 14 above)
Comment (myeh01) [reply src/gemm/xsfmm/gemm_f32c_f32_f32_xsfmm32a32f.c]: I added the wrappers because the underlying packed dispatch function has a slightly different API now. If that's not an issue, then I'll remove the wrappers. Previously, I was thinking in terms of IREE's packed layout where each block of A and B is a te x k panel. After updating the gemm and packing kernel naming scheme in #105, I agree it's better to think of the blocks as te x 1 (resp. 1 x te). This doesn't change the code, but `csa0` and `rsb0` are now `csa1` and `rsb1`, and the order of the stride arguments is different.

Comment (ericlove) [reply src/gemm/xsfmm/gemm_f32c_f32_f32_xsfmm32a32f.c]: If you think this adds something, then I'm fine keeping it.  I was just including this in a longer list of ways we could simplify the implementation, but it's not something I feel strongly about in particular.

File: src/gemm/xsfmm/gemm_f32c_f32_f32_xsfmm32a32f.c (hunk 9 above)
Comment (myeh01) [reply src/gemm/xsfmm/gemm_f32c_f32_f32_xsfmm32a32f.c:257]: Let me check if we even need `csc0` in the fused API. It may be possible to support column-major tiles implicitly by adjusting the `tss` that is passed in. Then for "asymmetric" fused kernels, we would need a "transposed" version of the kernel to handle column major tiles, e.g. if the kernel adds a bias vector to each row, the "transposed" version would add a bias vector to each column. But for "symmetric" kernels like alpha/beta scaling, we just need one version.

File: src/gemm/xsfmm/gemm_f32c_f32_f32_xsfmm32a32f.c (hunk 12 above)
Comment (myeh01) [reply src/gemm/xsfmm/gemm_f32c_f32_f32_xsfmm32a32f.c]: I'll write this up carefully in the documentation. The transpose case is a bit confusing, and I didn't clearly spell out all the details in the comments here.

File: src/gemm/xsfmm/gemm_f32c_f32_f32_xsfmm32a32f.c (hunk 13 above)
Comment (myeh01) [reply src/gemm/xsfmm/gemm_f32c_f32_f32_xsfmm32a32f.c]: Thanks for the suggestion! I think this significantly cleaned up the dispatch code.

File: src/gemm/xsfmm/gemm_f32c_f32_f32_xsfmm32a32f.c (hunk 9 above)
Comment (myeh01) [reply src/gemm/xsfmm/gemm_f32c_f32_f32_xsfmm32a32f.c:257]: I'll probably just go ahead and start writing everything up in the documentation. Then everyone can read it and check it makes sense.

**Comment by myeh01** [comment]
Ready for review.

**Review by ericlove — COMMENTED** [review]
This is looking great now! One more thing: can we fill in the definitions of [these macros](https://github.com/sifive/skl/blob/dc25c974f09d34e6f5656355acd51cd5ac48e84a/include/skl-common.h#L69) before merging it? This code is really not correct without them.

**Comment by myeh01** [comment]
> This is looking great now! One more thing: can we fill in the definitions of [these macros](https://github.com/sifive/skl/blob/dc25c974f09d34e6f5656355acd51cd5ac48e84a/include/skl-common.h#L69) before merging it? This code is really not correct without them.

We need to wait until the ABI attributes are available in the public toolchain, right?

**Comment by ericlove** [comment]
Okay, I guess we just need to keep tracking this PR: https://github.com/llvm/llvm-project/pull/206260

File: src/gemm/xsfmm/gemm_f32c_f32_f32_xsfmm32a32f.c [hunk 16]
```diff
@@ -0,0 +1,1245 @@
...
+    size_t m, size_t n, size_t k, const float *a, size_t rsa1, size_t csa1,
+    const float *b, size_t rsb1, size_t csb1) SKL_XSFMM_INOUT;
+
+/* params type for the alpha/beta scaling kernel */
+typedef struct {
+  float alpha;
+  float beta;
+} skl_gemm_alpha_beta_scaling_params_f32_f32rcptexterc_xsfmmbase_t;
+
+SKL_XSFMM_OUT
```
topperc: The SKL_XSFMM_OUT needs to be between the `)` and the `{` that starts the body.

src/gemm/xsfmm/gemm_f32c_f32_f32_xsfmm32a32f.c
```diff
@@ -0,0 +1,1245 @@
...
+ */
+SKL_XSFMM_OUT
+SKL_FUNC_PRIVATE
+void skl_gemm_tile_zero_f32_f32_xsfmmbase(size_t m, size_t n, size_t tss,
+                                          size_t rstss, size_t cstss) {
+  if (m == 0 || n == 0) {
+    return;
+  }
+
+  void (*tile_zero_functions[])(size_t, size_t) = {
```
topperc: This loses the SKL_XSFMM_OUT attribute.

src/gemm/xsfmm/gemm_f32c_f32_f32_xsfmm32a32f.c
myeh01: Yeah, I need to fix the placement of all the macros.

src/gemm/xsfmm/gemm_f32c_f32_f32_xsfmm32a32f.c
myeh01: Would the correct declaration be
```
void (*tile_zero_functions[])(size_t, size_t) SKL_XSFMM_OUT {
```
?

src/gemm/xsfmm/gemm_f32c_f32_f32_xsfmm32a32f.c
myeh01: Fixed

src/gemm/xsfmm/gemm_f32c_f32_f32_xsfmm32a32f.c
myeh01: Fixed. I will fix the placement in the other source files in future PRs.

src/gemm/xsfmm/gemm_f32c_f32_f32_xsfmm32a32f.c
myeh01: Fixed. I made a typedef for the tile zero functions with the attribute.

ericlove (dismissed):
Approving to unblock this.

src/gemm/xsfmm/gemm_f32c_f32_f32_xsfmm32a32f.c
myeh01: Ok, I added a branch in the alpha/beta kernel to basically call the row-major code if possible if C has column-major tiles.

ericlove (approved):

PR #15, opened 2026-01-29
[GEMM] Add remaining tilings for f32 Xsfmm GEMM

myeh01:
This PR adds 1xn and nx1 f32 Xsfmm GEMM kernels and updates the outer loop code to utilize these in place of the 1x1 kernel where possible for improved performance.

### DIFF
--- a/src/gemm/xsfmm/gemm_a1b01_f32c_f32_f32_xsfmm32a32f.c
+++ b/src/gemm/xsfmm/gemm_a1b01_f32c_f32_f32_xsfmm32a32f.c
@@ -97,6 +97,182 @@ SKL_FUNC_PRIVATE void skl_gemm_1tm1tn_a1b01_f32c_f32_f32_xsfmm32a32f(
       : "vtype", "vl", "memory");
 }
 
+/* C consists of two tm x tn row major tiles C0 and C1. Each tile's row stride
+ * is rsc0, and the distance from C0 to C1 in elements is tsc1. A is a row major
+ * matrix with k rows and row stride rsa. B consists of two row major tiles B0
+ * and B1. Each tile has k rows and row stride rsb0, and the distance from B0 to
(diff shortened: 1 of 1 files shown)


JerryShih (commented):
We have a strong criterial that all inner tile will be TE*TE.
Do we still need to call vsettnt/vsettm in the skl_gemm_*tm*tn()?

ericlove:
> We have a strong criterial that all inner tile will be TE*TE.

Who is "we' in this case? IREE does, because MMT4D packs and pads.  But general users might not, and code in the non-2x2 case will by definition be "fringe handling" where we expect that the matrix dimensions might not be multiples of TE.

That said, I agree that there is a few % (5%?) performance gap in this tightly-scheduled code, so if this is a showstopper for IREE integration, I'm happy to add back the vsettn-removal optimization. Maybe this can just be done as a special case of the innermost loop, so if it's easy enough, @myeh01 should go ahead and add it.  I'll let the two of you decide what to do.

JerryShih:
> > We have a strong criterial that all inner tile will be TE*TE.
> 
> Who is "we' in this case? IREE does, because MMT4D packs and pads. But general users might not, and code in the non-2x2 case will by definition be "fringe handling" where we expect that the matrix dimensions might not be multiples of TE.
> 
> That said, I agree that there is a few % (5%?) performance gap in this tightly-scheduled code, so if this is a showstopper for IREE integration, I'm happy to add back the vsettn-removal optimization. Maybe this can just be done as a special case of the innermost loop, so if it's easy enough, @myeh01 should go ahead and add it. I'll let the two of you decide what to do.

Yes, it's the `IREE` case. Let's wait the IREE benchmark result and check if we need this small piece of gap.

JerryShih (dismissed):

myeh01:
@JerryShih I pushed some fixes, so will need another review on this.

JerryShih (approved):

PR #16, opened 2026-01-30
[GEMM] Convert `gemm_4xm4x1_f32_f32_f32_zve32f_x390` to inline assembly

Aaron-Hutchinson:
This performs a conversion of the `gemm_4xm4x1_f32_f32_f32_zve32f_x390` kernel from intrinsics to inline assembly to obtain a fixed (compiler independent) instruction schedule. The schedule used in the new inline assembly version is the one that was intended in the original RVV intrinsics version. I observed a small performance gain with the inline asm version compared to the intrinsics version.


myeh01 (approved):
LGTM

hdelassus (approved):
LGTM too

PR #17, opened 2026-02-02
[GEMM] Add remaining tilings for f16 Xsfmm GEMM

myeh01:
This PR adds 1xn and nx1 f16 Xsfmm GEMM kernels and updates the outer loop code to utilize these in place of the 1x1 kernel where possible for improved performance.

### DIFF
--- a/src/gemm/xsfmm/gemm_a1b01_f16c_f16_f32_xsfmm32a16f.c
+++ b/src/gemm/xsfmm/gemm_a1b01_f16c_f16_f32_xsfmm32a16f.c
@@ -949,7 +949,7 @@ SKL_FUNC_PRIVATE void skl_gemm_4tm1tn_a1b01_f16pc_f16_f32rcp_xsfmm32a16f(
 
 /* Process 4 (= 2 x 2) contiguous tn x tn tiles of c. tn must be <= TE. */
 SKL_XSFMM_NEW
-SKL_FUNC_PRIVATE void skl_gemm_2tm2tn_a1b01_f16pc_f16cp_f32rcp_xsfmm32a16f(
+SKL_FUNC_PRIVATE void skl_gemm_2tn2tn_a1b01_f16pc_f16cp_f32rcp_xsfmm32a16f(
     size_t tn, size_t k, const _Float16 *a, size_t csa0, size_t rsa1,
     const _Float16 *b, size_t rsb0, size_t csb1, float *c, size_t rsc0,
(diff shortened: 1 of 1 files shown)


JerryShih (dismissed):

src/gemm/xsfmm/gemm_a1b01_f16c_f16_f32_xsfmm32a16f.c
```diff
@@ -947,14 +947,16 @@ SKL_FUNC_PRIVATE void skl_gemm_4tm1tn_a1b01_f16pc_f16_f32rcp_xsfmm32a16f(
...
 }
 
-/* Process 4 (= 2 x 2) contiguous tm x tn tiles of c.
- * tm and tn must be <= TE.
- */
+/* Process 4 (= 2 x 2) contiguous tn x tn tiles of c. tn must be <= TE. */
 SKL_XSFMM_NEW
 SKL_FUNC_PRIVATE void skl_gemm_2tm2tn_a1b01_f16pc_f16cp_f32rcp_xsfmm32a16f(
-    size_t tm, size_t tn, size_t k, const _Float16 *a, size_t csa0, size_t rsa1,
+    size_t tn, size_t k, const _Float16 *a, size_t csa0, size_t rsa1,
```
JerryShih, line 953: why only do specialize for skl_gemm_2tm2tn_a1b01_f16pc_f16cp_f32rcp_xsfmm32a16f()?
how about other 1x4/3/2 func?

JerryShih: is it trying to share the same 1x4/3/2 impls between skl_gemm_a1b01_f16pc() and skl_gemm_a1b01_f16c()?

JerryShih (approved):

src/gemm/xsfmm/gemm_a1b01_f16c_f16_f32_xsfmm32a16f.c
myeh01: Right, we want to try reusing these 1x4/3/2 kernels between the row-major and packed GEMMs. I can add specializations for 1x4/3/2 in a follow up PR if needed. It will just involve adding an `if (tm == tn)` branch into the kernels. For the 1x4/3/2 kernels, we do need to support `tm != tn` since it is needed for the fringe case in row-major GEMM.

PR #18, opened 2026-02-02
Use shared macro for OFP4 and OFP8 conversion benchmarks

hdelassus:
This PR makes OFP4 and OFP8 conversion benchmarks use the `SKL_BENCHMARK_RUN` macro introduced in #11. It can be tested with a toolchain that supports Zvfofp4min and Zvfofp8min by un-commenting the [extensions](https://github.com/sifive/skl/blob/a77b0ca686e32a6662d72f8da247b6f403525809/cmake/riscv.cmake#L28-L29) in the `cmake/riscv.cmake` toolchain file.
The output of `ctest` should remain broadly the same (modulo cycles and instret counter values) whatever the combination of `SKL_BUILD_TESTS` and `SKL_BUILD_BENCHMARKS` is.

### DIFF
--- a/src/README.md
+++ b/src/README.md
@@ -27,6 +27,7 @@ Each datatype is represented in the `<types>` field with a string generally foll
 
 | `<type>` | Description | C Type |
 |----------|-------------|---------|
+| `f4e2m1` | 4-bit OFP4 E2M1 floating point | `uint8_t` |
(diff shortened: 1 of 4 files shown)


Harlen126 (approved):
Thanks for the refinement.
LGTM!

PR #20, opened 2026-02-03
[GEMM] Convert `gemm_1x2m8x2_f32_f32_f32_zve32f_x390` to inline assembly

Aaron-Hutchinson:
This converts the `gemm_1x2m8x2_f32_f32_f32_zve32f_x390` kernel from RVV intrinsics to inline assembly to obtain a fixed instruction schedule. I used the same schedule as the existing one and observed no significant difference in performance between the two versions.

I also deleted the `jj0` "fixup" loop since it's unnecessary with the previous VLA vectorized loop.


hdelassus (approved):
LGTM

ericlove (approved):
Conversion seems fine. I do think there is some further optimization (particularly scheduling) to do in a separate PR.

PR #26, opened 2026-02-10
Move skl_cvt_* to skl-ofp-cvt.h

HanKuanChen:
Including <math.h> and <string.h> may introduce problems for kernels that do not depend on the skl_cvt_* functions, particularly in freestanding environments. This PR move the skl_cvt_* functions into an independent header <skl-ofp-cvt.h>.


ericlove (changes requested):
Please do not add new shared headers to SKL. SKL kernels are only allowed to depend on `skl-common.h`.  This is a [key feature](https://github.com/sifive/skl/?tab=readme-ov-file#features) of the library.  The scalar conversion functions should either be rewritten not to depend on standard library functions, or they should be moved to the test suite and not used in `skl/src`.

hdelassus:
What problem is this PR trying to solve? There are quite a few open source C libraries specifically designed for freestanding implementations, e.g. Newlib nano, many of which supporting `<string.h>` and `<math.h>`.

HanKuanChen:
> What problem is this PR trying to solve? There are quite a few open source C libraries specifically designed for freestanding implementations, e.g. Newlib nano, many of which supporting `<string.h>` and `<math.h>`.

IREE expects microkernels be freestanding.

> Please do not add new shared headers to SKL. SKL kernels are only allowed to depend on `skl-common.h`. This is a [key feature](https://github.com/sifive/skl/?tab=readme-ov-file#features) of the library. The scalar conversion functions should either be rewritten not to depend on standard library functions, or they should be moved to the test suite and not used in `skl/src`.

What do you mean "not used in `skl/src`"? Move `src/gemm/scalar/gemm_f8e4m3rc_f8e4m3rc_f32rc_scalar.c` and `src/gemm/scalar/gemm_f8e4m3rcprc_f8e4m3rcprc_f32rcprc_scalar.c` into test?

ebavier:
The `skl_cvt_f32_*` functions could be rewritten to avoid the `<math.h>` functions (`fabs`, `floorf`, `nearbyintf`, `powf`, `log2f`, `isnan`, and `isinf`).  E.g. `powf` and `log2f` here can be easily replaced with integer ops.
But the `skl_cvt_f32_*` in `skl-common.h` seem to only be used in the `cvt` test suites.  I assume we aren't prohibiting local headers for test suites?

Harlen126:
> But the skl_cvt_f32_* in skl-common.h seem to only be used in the cvt test suites. I assume we aren't prohibiting local headers for test suites?

Actually, only `skl_cvt_f8e4m3_f32()` is currently being called by SKL sources. Therefore, we can move the other conversion functions to the test suites and move them back to `skl-common.h` if new kernels need them.

However, this approach might make new kernel authors unaware of these functions, leading them to implement redundant ones rather than reusing the existing code. For example, @myeh01  and @Harlen126  originally had their own implementations of `skl_cvt_f8e4m3_f32()` in different places before moving and sharing them in a common header. 
I’m wondering how we can provide enough visibility so that new authors don't overlook them.

ericlove:
> What do you mean "not used in `skl/src`"? Move `src/gemm/scalar/gemm_f8e4m3rc_f8e4m3rc_f32rc_scalar.c` and `src/gemm/scalar/gemm_f8e4m3rcprc_f8e4m3rcprc_f32rcprc_scalar.c` into test?

Yes.  I realize that this creates some discrepancies between kernel families (some have scalar reference implementations in `skl/src` while others do not), but that is still better than a libc dependency inside the main library.

Of course, I am happy if we rewrite them to avoid libc calls instead.  I would entrust this task to @ebavier (replacing the rounding code is the trickiest part; `memcpy` can be replaced with `__builtin_memcpy`), but that should be done in another PR.  For now, let's just move these into `skl/test`?

HanKuanChen:
Please check if this is what you want.

ericlove (approved):
Yes, thank you, this looks good!

PR #27, opened 2026-02-10
Add forward declaration of SKL_TEST_PERF_REPORT

hdelassus:
When SKL_TEST_PERF_REPORT is defined, add a forward declaration in skl-test.h so that source files including skl-test.h may call it even if its definition comes from another translation unit.


Aaron-Hutchinson (approved):
LGTM

PR #29, opened 2026-02-11
[Test] [GEMM] Support custom allocators for benchmark buffers

ericlove:
To make the SKL test harness suite more useful for certain benchmarking environments, this PR introduces a new paradigm for primary buffer allocation in kernel benchmarks.  Specifically, the `CUSTOM_ALLOC` and `CUSTOM_FREE` macros allow surrounding environments to control how and, crucially, _where_ benchmark data resides in memory.  For example, some systems may need to place data in fast scratchpad memory, etc.

As an initial demonstration, the new allocation strategy is implemented in the `gemm_f32rc_f32rc_f32rc` benchmark.  After merger, others should be ported to this format in separate PRs.

## Custom Allocator Syntax

If the `CUSTOM_ALLOC` macro is defined, it is expected to name a function of the following signature:
```c
void *CUSTOM_ALLOC(size_t alignment, size_t size);
```
This matches the C11 [aligned_alloc](https://en.cppreference.com/w/c/memory/aligned_alloc) function in argument order and should provide identical semantics.

Client environments that define `CUSTOM_ALLOC` may optionally define `CUSTOM_FREE`, which benchmarks must use during post-execution cleanup.

## Hybrid Static/Dynamic Allocation

Currently, all buffers are defined as static global arrays (`_Alignas(ALIGN) float c[CLEN];`); this PR changes to conditionally declare uninitialized pointers if `CUSTOM_ALLOC` is defined:
```c
#if defined(CUSTOM_ALLOC)
float *a;
float *b;
float *c;
#else
_Alignas(ALIGN) float a[ALEN];
_Alignas(ALIGN) float b[BLEN];
_Alignas(ALIGN) float c[CLEN];
#endif
```

## Rationale

Ideally, this kind of thing could be handled linker scripts and section attributes, but in practice we have found these may not be available in every environment.  At the opposite extreme, if we use `malloc`/`free` or equivalents in the benchmark harnesses, then ideally all allocation could be made dynamic, but unfortunately this too fails on systems without a working `aligned_alloc`.

### DIFF
--- a/test/skl-test.h
+++ b/test/skl-test.h
@@ -512,18 +512,33 @@ SKL_TEST_INIT_FUNC(int32_t, i32, INT, unused)
 
 /**
  * @brief Custom memory allocation function
- * @details If defined, this function will be used to allocate memory for
- * buffers in benchmark mode.
+ *
+ * If defined, this function will be used to allocate memory for
+ * buffers in benchmark mode. If not defined, benchmarks should
+ * declare and initialize buffers as static arrays.
  */
 #if defined(CUSTOM_ALLOC)
+/**
+ * @brief Aligned memory allocation function.
+ *
+ * @param alignment Alignment of memory to allocate in bytes
+ * @param size Size of memory to allocate in bytes
+ * @return Pointer to allocated memory
+ */
 void *CUSTOM_ALLOC(size_t alignment, size_t size);
 #endif
 
 /**
  * @brief Custom memory free function
- * @details If defined, this function will be used to free memory allocated
+ *
+ * If defined, this function will be used to free memory allocated
  * by CUSTOM_ALLOC.
  */
 #if defined(CUSTOM_FREE)
+/**
+ * @brief Custom memory free function.
+ *
+ * @param ptr Pointer to memory to free
+ */
 void CUSTOM_FREE(void *ptr);
 #endif
--- a/test/skl-test.h
+++ b/test/skl-test.h
@@ -516,6 +516,10 @@ SKL_TEST_INIT_FUNC(int32_t, i32, INT, unused)
  * If defined, this function will be used to allocate memory for
  * buffers in benchmark mode. If not defined, benchmarks should
  * declare and initialize buffers as static arrays.
+ *
+ * @note The prototype is declared by this macro so that it is not
+ * it can name an arbitrary symbol in a linked translation unit without
+ * modification to the benchmark source.
  */
 #if defined(CUSTOM_ALLOC)
 /**
--- a/test/skl-test.h
+++ b/test/skl-test.h
@@ -509,3 +509,21 @@ SKL_TEST_INIT_FUNC(int32_t, i32, INT, unused)
 #undef SKL_TEST_INIT_FUNC
 #undef SKL_TEST_INIT_RANDOM_IMPL_FLOAT
 #undef SKL_TEST_INIT_RANDOM_IMPL_INT
+
+/**
+ * @brief Custom memory allocation function
+ * @details If defined, this function will be used to allocate memory for
+ * buffers in benchmark mode.
+ */
+#if defined(CUSTOM_ALLOC)
+void *CUSTOM_ALLOC(size_t alignment, size_t size);
+#endif
+
+/**
+ * @brief Custom memory free function
+ * @details If defined, this function will be used to free memory allocated
+ * by CUSTOM_ALLOC.
+ */
+#if defined(CUSTOM_FREE)
+void CUSTOM_FREE(void *ptr);
+#endif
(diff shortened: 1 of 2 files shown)


test/skl-test.h
```diff
@@ -509,3 +509,40 @@ SKL_TEST_INIT_FUNC(int32_t, i32, INT, unused)
...
+
+/**
+ * @brief Custom memory allocation function
+ *
+ * If defined, this function will be used to allocate memory for
+ * buffers in benchmark mode. If not defined, benchmarks should
+ * declare and initialize buffers as static arrays.
+ *
+ * @note The prototype is declared by this macro so that it is not
+ * it can name an arbitrary symbol in a linked translation unit without
```
Aaron-Hutchinson: I'm guessing you meant:

```suggestion
 * @note The prototype is declared by this macro so that
 * it can name an arbitrary symbol in a linked translation unit without
```

test/skl-test.h
```diff
@@ -509,3 +509,40 @@ SKL_TEST_INIT_FUNC(int32_t, i32, INT, unused)
...
+ * modification to the benchmark source.
+ */
+#if defined(CUSTOM_ALLOC)
```
Aaron-Hutchinson: What dictates whether we use a `SKL_` prefix on macros? It seems we're pretty inconsistent about this in the test directory.

test/skl-test.h
```diff
@@ -509,3 +509,40 @@ SKL_TEST_INIT_FUNC(int32_t, i32, INT, unused)
...
+ * @return Pointer to allocated memory
+ */
+void *CUSTOM_ALLOC(size_t alignment, size_t size);
+#endif
+
+/**
+ * @brief Custom memory free function
+ *
+ * If defined, this function will be used to free memory allocated
+ * by CUSTOM_ALLOC.
```
Aaron-Hutchinson: Shouldn't this have the same `@note` as `CUSTOM_ALLOC`?

test/skl-test.h
```diff
@@ -509,3 +509,40 @@ SKL_TEST_INIT_FUNC(int32_t, i32, INT, unused)
...
+ */
+#if defined(CUSTOM_FREE)
```
Aaron-Hutchinson: It sounds like there is an implicit dependency on `CUSTOM_ALLOC` (at least the gemm harness reflects that), so maybe this should be:

```suggestion
#if defined(CUSTOM_FREE) && defined(CUSTOM_ALLOC)
```

test/skl-test.h
```diff
@@ -509,3 +509,40 @@ SKL_TEST_INIT_FUNC(int32_t, i32, INT, unused)
 #undef SKL_TEST_INIT_FUNC
 #undef SKL_TEST_INIT_RANDOM_IMPL_FLOAT
 #undef SKL_TEST_INIT_RANDOM_IMPL_INT
+
+/**
+ * @brief Custom memory allocation function
+ *
+ * If defined, this function will be used to allocate memory for
+ * buffers in benchmark mode. If not defined, benchmarks should
```
Aaron-Hutchinson: > in benchmark mode

The harness uses it in test mode as well.

test/skl-test.h
ericlove: I agree there isn't really a consistent policy. I think at a subconscious level I had some notion of "a SKL test" and something that changed a test's properties got this prefix (like input data ranges), while others like `ENABLE_BENCHMARK` did not.

We should figure out a better rule.

ericlove: Hmm, maybe it would be better to nip this in the bud right now and call this `SKL_TEST_MALLOC`?

peichialin:
> Client environments that define CUSTOM_ALLOC may optionally define CUSTOM_FREE, which benchmarks must use during post-execution cleanup.

@ericlove What's the reason or concern of allowing not defining `CUSTOM_FREE` when `CUSTOM_ALLOC` is available?

ericlove:
> @ericlove What's the reason or concern of allowing not defining `CUSTOM_FREE` when `CUSTOM_ALLOC` is available?

@peichialin At risk of exposing substandard internal practices: not every custom allocator will have a corresponding free.  This is just slightly more flexible, and avoids a hard requirement for empty `custom_free(void) {}` declarations if a client environment doesn't already have one.

ericlove:
I've [pushed a commit](https://github.com/sifive/skl/pull/29/changes/c0d279611ef352abc5aa207c8a68f6671b52e34a) to rename to `SKL_TEST_{MALLOC,FREE}`.  @Aaron-Hutchinson what do you think of this change?

Aaron-Hutchinson:
I like it, thanks for making it more consistent!

Aaron-Hutchinson (approved):
LGTM, thanks!

PR #34, opened 2026-02-17
[Test] Static buffer init mode & new SKL_TEST_INIT_{*} macros (FP only)

ericlove:
This PR adds a new `TEST_INIT_MODE` -- `STATIC` -- to the buffer init code that copies data from a pre-defined array `BUF##_data` for `BUF`.  This is sometimes needed support execution environments such as RTL simulation where the cost of random initialization during runtime would be too high.

However, since the existing initialization functions cannot access the name of the buffer parameter to associate with a corresponding data array, I have introduced a new initialization paradigm that uses macros instead of functions:

```c
#define SKL_TEST_INIT(BUF, LEN, TYPE, MIN, MAX, IMPL_TYPE, FRAC_TYPE)          \
  {                                                                            \
    TYPE *buf = (BUF);                                                         \
    const size_t len = (LEN);                                                  \
    size_t avl = len;                                                          \
    const size_t buf_len = SKL_TEST_STATIC_DATA_LEN(BUF);                      \
    while (avl > 0) {                                                          \
      size_t vl = avl > buf_len ? buf_len : avl;                               \
      memcpy(buf, SKL_TEST_STATIC_DATA(BUF), vl * sizeof(TYPE));               \
      avl -= vl;                                                               \
      buf += vl;                                                               \
    }                                                                          \
  }
```

This somewhat unwieldy definition is complemented by simpler wrappers:
```c
#define SKL_TEST_INIT_F16(BUF, LEN)  [...]
#define SKL_TEST_INIT_F32(BUF, LEN)  [...]
#define SKL_TEST_INIT_F64(BUF, LEN)  [...]
```

This PR only adds these macros for the F{16,32,64} types, and makes use only of the F32 version in the single benchmark of `gemm_f32rc_f32rc_f32rc.c`.  Subsequent PRs will add the remaining types and test conversions. **Ultimately, the goal is to deprecate and remove the old `skl_test_init_{*}` functions**.

### DIFF
--- a/test/skl-test.h
+++ b/test/skl-test.h
@@ -767,10 +767,10 @@ static inline int skl_check_error_ulp_bf16(const char *name, const __bf16 *res,
  *
  * Behavior is determined by TEST_INIT_MODE macro:
  * - RANDOM: Initialize with random values between min and max
- * - STATIC: Copy values from existing array NAME_data to buf
+ * - STATIC: Copy values from existing array BUF_data to buf
  * - SEQ: Initialize with equally-spaced values from min to max
  *
- * @note For STATIC mode, the array NAME_data must be defined in the header file
+ * @note For STATIC mode, the array BUF_data must be defined in the header file
  * specified by SKL_TEST_DATA_HEADER, but need not have the same length as the
  * buffer being initialized. The data will be repeated as necessary to fill the
  * buffer.
--- a/test/skl-test.h
+++ b/test/skl-test.h
@@ -615,12 +615,12 @@ static inline int skl_check_error_ulp_bf16(const char *name, const __bf16 *res,
  * to test specific ranges of interest. More modes may be added in the future.
  *
  * In STATIC mode, the input data is initialized with static values from
- * pre-defined arrays in data.h. Each
+ * pre-defined arrays in SKL_TEST_DATA_HEADER.
  */
 enum {
   SEQ,    ///< Initialize with equally-spaced values from min to max
   RANDOM, ///< Initialize with random values
-  STATIC  ///< Initialize with static values from arrays in data.h
+  STATIC  ///< Initialize with static values from SKL_TEST_DATA_HEADER
 };
 
 /**
--- a/test/skl-test.h
+++ b/test/skl-test.h
@@ -752,32 +752,37 @@ enum {
 #define SKL_TEST_STATIC_DATA_LEN(NAME) 0
 #endif
 
+#if defined(TEST_INIT_MODE) && TEST_INIT_MODE == STATIC
 #define SKL_TEST_INIT(BUF, LEN, TYPE, MIN, MAX, IMPL_TYPE, FRAC_TYPE)          \
   {                                                                            \
     TYPE *buf = (BUF);                                                         \
     const size_t len = (LEN);                                                  \
-    if (TEST_INIT_MODE == STATIC) {                                            \
-      size_t avl = len;                                                        \
-      const size_t buf_len = SKL_TEST_STATIC_DATA_LEN(BUF);                    \
-      while (avl > 0) {                                                        \
-        size_t vl = avl > buf_len ? buf_len : avl;                             \
-        memcpy(buf, SKL_TEST_STATIC_DATA(BUF), vl * sizeof(TYPE));             \
-        avl -= vl;                                                             \
-        buf += vl;                                                             \
-      }                                                                        \
-    } else {                                                                   \
-      const TYPE min = (MIN);                                                  \
-      const TYPE max = (MAX);                                                  \
-      const TYPE step = (max - min) / len;                                     \
-      for (size_t i = 0; i < len; i++) {                                       \
-        if (TEST_INIT_MODE == RANDOM) {                                        \
-          SKL_TEST_INIT_RANDOM_IMPL_##IMPL_TYPE(TYPE, FRAC_TYPE);              \
-        } else {                                                               \
-          buf[i] = (TYPE)(min + step * i);                                     \
-        }                                                                      \
+    size_t avl = len;                                                          \
+    const size_t buf_len = SKL_TEST_STATIC_DATA_LEN(BUF);                      \
+    while (avl > 0) {                                                          \
+      size_t vl = avl > buf_len ? buf_len : avl;                               \
+      memcpy(buf, SKL_TEST_STATIC_DATA(BUF), vl * sizeof(TYPE));               \
+      avl -= vl;                                                               \
+      buf += vl;                                                               \
+    }                                                                          \
+  }
+#else
+#define SKL_TEST_INIT(BUF, LEN, TYPE, MIN, MAX, IMPL_TYPE, FRAC_TYPE)          \
+  {                                                                            \
+    TYPE *buf = (BUF);                                                         \
+    const size_t len = (LEN);                                                  \
+    const TYPE min = (MIN);                                                    \
+    const TYPE max = (MAX);                                                    \
+    const TYPE step = (max - min) / len;                                       \
+    for (size_t i = 0; i < len; i++) {                                         \
+      if (TEST_INIT_MODE == RANDOM) {                                          \
+        SKL_TEST_INIT_RANDOM_IMPL_##IMPL_TYPE(TYPE, FRAC_TYPE);                \
+      } else {                                                                 \
+        buf[i] = (TYPE)(min + step * i);                                       \
       }                                                                        \
     }                                                                          \
   }
+#endif
 
 #define SKL_TEST_INIT_RANDOM_IMPL_FLOAT(TYPE, FRAC_TYPE)                       \
   FRAC_TYPE frac = (FRAC_TYPE)rand() / (FRAC_TYPE)RAND_MAX;                    \
--- a/test/skl-test.h
+++ b/test/skl-test.h
@@ -612,10 +612,14 @@ static inline int skl_check_error_ulp_bf16(const char *name, const __bf16 *res,
  * Benchmarks will typically use SEQ to reduce overhead from random number
  * generation. Tests will typically use RANDOM though some will likely use SEQ
  * to test specific ranges of interest. More modes may be added in the future.
+ *
+ * In STATIC mode, the input data is initialized with static values from
+ * pre-defined arrays in data.h. Each
  */
 enum {
-  SEQ,   ///< Initialize with equally-spaced values from min to max
-  RANDOM ///< Initialize with random values
+  SEQ,    ///< Initialize with equally-spaced values from min to max
+  RANDOM, ///< Initialize with random values
+  STATIC  ///< Initialize with static values from arrays in data.h
 };
 
 /**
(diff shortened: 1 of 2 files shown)


test/skl-test.h
```diff
@@ -568,10 +569,14 @@ static inline int skl_check_error_ulp_bf16(const char *name, const __bf16 *res,
  * Benchmarks will typically use SEQ to reduce overhead from random number
  * generation. Tests will typically use RANDOM though some will likely use SEQ
  * to test specific ranges of interest. More modes may be added in the future.
+ *
+ * In STATIC mode, the input data is initialized with static values from
+ * pre-defined arrays in data.h. Each
```
hdelassus: Superfluous "Each", or missing the rest of the sentence?

test/skl-test.h
```diff
@@ -696,13 +702,100 @@ enum {
...
+ * These macros are intended to be the standard way to initialize data buffers
+ * in SKL tests and benchmarks. They will displace the direct use of
+ * skl_test_init functions (deprecated).
+ *
+ * All macros in this group share a common interface and behavior:
+ * - @param BUF Output buffer to fill with values
+ * - @param LEN Number of elements to generate
+ * All macros are wrappers around SKL_TEST_INIT, which has more detailed
+ * documentation.
+ * @{
```
hdelassus, line 825: Missing a closing bracket?

hdelassus (dismissed):
LGTM pending some pedantry.

test/skl-test.h
```diff
@@ -696,13 +702,100 @@ enum {
...
+#if defined(TEST_INIT_MODE) && TEST_INIT_MODE == STATIC
+#define SKL_TEST_INIT(BUF, LEN, TYPE, MIN, MAX, IMPL_TYPE, FRAC_TYPE)          \
+  {                                                                            \
+    TYPE *buf = (BUF);                                                         \
+    const size_t len = (LEN);                                                  \
+    size_t avl = len;                                                          \
+    const size_t buf_len = SKL_TEST_STATIC_DATA_LEN(BUF);                      \
+    while (avl > 0) {                                                          \
+      size_t vl = avl > buf_len ? buf_len : avl;                               \
+      memcpy(buf, SKL_TEST_STATIC_DATA(BUF), vl * sizeof(TYPE));               \
```
Aaron-Hutchinson, line 790: I'm probably missing the full vision here. It looks like this assumes that `BUF` and `SKL_TEST_STATIC_DATA(BUF)` have the same type, since `len` and `avl` are measured in units of elements. This implies we'd need different `data.h` files for each type, conditionally included into each harness, right? Or alternatively, bake the datatype into the variable names?

Furthermore a test for a e.g. widening GEMM kernel would need different datatypes among the buffers, which I think would necessitate separating each matrix into its own `data.h` file, and that sounds like a mess of data files and `#include`s.

Am I missing something?

test/skl-test.h
Aaron-Hutchinson: It does appear to be missing. I don't think we currently build the test directory in the documentation, so this whole file's Doxygen documentation is probably untested.

test/skl-test.h
ericlove: The type comes from the init call in the test harness, so it shouldn't be any different from the existing init functions.  For a widening GEMM, you'd call `SKL_TEST_INIT_F16(a, ALEN)` and `SKL_TEST_INIT_F32(b, BLEN)` for example.

ericlove:
Thanks to @Aaron-Hutchinson and @hdelassus for the helpful review.  I have made a couple of changes:

1. Static buffers are expected to be defined in a header designated by `SKL_TEST_DATA_HEADER`. This is included automatically when `TEST_INIT_MODE == STATIC` (and a compile-time error is generated if it is not defined).
2. The `BUF##_len` value is expected to be defined, and not computed automatically. This allows data stored in not necessarily static arrays.
3. Several fixes to doxygen. I did not intend `data.h` to be part of the final design.

test/skl-test.h
Aaron-Hutchinson: My point is that it seems very difficult to share a data header between different harnesses. For example, a widening F16 GEMM kernel could use the same C buffer as the F32 kernel, but the A/B matrices would necessarily need to be in separate headers due to the name collision between the F16 data and F32 data since both harnesses name their variables `a` and `b`.

Aaron-Hutchinson: This is basically caused by us trying to force all initialization through the single macro `SKL_TEST_INIT`, when it would be a lot more flexible for each initialization method to have a dedicated interface.

ericlove: There is no intent to share a data header between different harnesses.  These will likely be (possibly auto-generated) stubs supplied by a client application's build system.  We are assuming each test/benchmark is built in a separate executable.

ericlove:
I also had to change the test mode `enum` to macros, as these values are used in macro conditionals: https://github.com/sifive/skl/pull/34/changes/6a8d3bf7248a88ad0afbc6f97c56754e70d2961c

test/skl-test.h
```diff
@@ -803,10 +900,6 @@ SKL_TEST_INIT_FUNC(int32_t, i32, INT, unused)
 
 /** @} */ // end of test_init group
 
-#undef SKL_TEST_INIT_FUNC
-#undef SKL_TEST_INIT_RANDOM_IMPL_FLOAT
-#undef SKL_TEST_INIT_RANDOM_IMPL_INT
```
Aaron-Hutchinson, line 808: Why were these deleted?

test/skl-test.h
```diff
@@ -740,13 +747,103 @@ enum {
...
+ * @param LEN - The length of the buffer
+ * @param TYPE - The data type of the buffer
+ * @param MIN - The minimum value to use for initialization
+ * @param MAX - The maximum value to use for initialization
+ * @param IMPL_TYPE - Either FLOAT or INT to select implementation
+ * @param FRAC_TYPE - Fractional type for floating point intermediates(float or
+ * double)
+ */
+#if defined(TEST_INIT_MODE) && TEST_INIT_MODE == STATIC
+#define SKL_TEST_INIT(BUF, LEN, TYPE, MIN, MAX, IMPL_TYPE, FRAC_TYPE)          \
```
Aaron-Hutchinson: I'm still not really understanding why the different initialization modes have to use the same interface. Many of these arguments aren't used in the first code path here. Why not have `SKL_TEST_INIT_STATIC`, `SKL_TEST_INIT_RANDOM`, etc, with their own specialized arguments?

test/skl-test.h
```diff
@@ -740,13 +747,103 @@ enum {
...
+ * buffer being initialized. The data will be repeated as necessary to fill the
+ * buffer.
+ *
+ * @param BUF - The buffer to initialize
+ * @param LEN - The length of the buffer
+ * @param TYPE - The data type of the buffer
+ * @param MIN - The minimum value to use for initialization
+ * @param MAX - The maximum value to use for initialization
+ * @param IMPL_TYPE - Either FLOAT or INT to select implementation
+ * @param FRAC_TYPE - Fractional type for floating point intermediates(float or
```
Aaron-Hutchinson: None of these are used in STATIC mode, which should probably be noted here.

test/skl-test.h
ericlove: Now that the test suites use a macro instead of a function, they need to still have `SKL_TEST_INIT_RANDOM_IMPL_{FLOAT,INT}` defined because they are referred to in the macro expansion.

test/skl-test.h
ericlove: This is so that the test harness can call the macro without knowing which mode is configured.

Aaron-Hutchinson: Is it an issue if it does know? If it's just to reduce code duplication the initialization could be factored into a shared header.

This approach just doesn't feel like a great design to me. Some arguments here are only used in one `TEST_INIT_MODE` case, whereas other cases could use additional arguments (like a source/destination buffer for `STATIC` mode), but don't get them to avoid cluttering the interface further. Why does `RANDOM` get its own arguments but not `STATIC`?

ericlove: The `MIN` and `MAX` arguments are necessary because they are set based on the type, which is easier to express in the `_F32` etc. wrappers.  These are a good way to hide anything "bad" about the underlying design.

myeh01 (dismissed):

test/skl-test.h
```diff
@@ -740,13 +747,103 @@ enum {
...
+ * buffer.
+ *
+ * @param BUF - The buffer to initialize
+ * @param LEN - The length of the buffer
+ * @param TYPE - The data type of the buffer
+ * @param MIN - The minimum value to use for initialization (ignored in STATIC)
+ * @param MAX - The maximum value to use for initialization (ignored in STATIC)
+ * @param IMPL_TYPE - Either FLOAT or INT to select implementation
+ * @param FRAC_TYPE - Fractional type for floating point intermediates(float or
+ * double)
```
hdelassus:
```suggestion
 * @param FRAC_TYPE - Fractional type for floating point intermediates (float or
 * double)
```

test/skl-test.h
```diff
@@ -740,13 +747,103 @@ enum {
...
 
+/**
+ * @brief Macro to initialize a buffer in a SKL test/benchmark.
+ *
+ * Behavior is determined by TEST_INIT_MODE macro:
+ * - RANDOM: Initialize with random values between min and max
+ * - STATIC: Copy values from existing array NAME_data to buf
+ * - SEQ: Initialize with equally-spaced values from min to max
+ *
+ * @note For STATIC mode, the array NAME_data must be defined in the header file
```
hdelassus:
```suggestion
 * - STATIC: Copy values from existing array BUF_data to buf
 * - SEQ: Initialize with equally-spaced values from min to max
 *
 * @note For STATIC mode, the array BUF_data must be defined in the header file
```
Isn't NAME_data confusing? The argument to `SKL_TEST_STATIC_DATA` is `BUF` here.

test/skl-test.h
```diff
@@ -740,13 +747,103 @@ enum {
     }                                                                          \
   }
 
+#if defined(TEST_INIT_MODE) && TEST_INIT_MODE == STATIC
+#define SKL_TEST_STATIC_DATA(NAME) NAME##_data
+#define SKL_TEST_STATIC_DATA_LEN(NAME) NAME##_len
+#else
+#define SKL_TEST_STATIC_DATA(NAME) NULL
+#define SKL_TEST_STATIC_DATA_LEN(NAME) 0
```
hdelassus: Seems like this is meant as a safety measure for yet-unwritten broken logic. 
```suggestion
```

test/skl-test.h
```diff
@@ -740,13 +747,103 @@ enum {
...
+#if defined(TEST_INIT_MODE) && TEST_INIT_MODE == STATIC
+#define SKL_TEST_INIT(BUF, LEN, TYPE, MIN, MAX, IMPL_TYPE, FRAC_TYPE)          \
+  {                                                                            \
+    TYPE *buf = (BUF);                                                         \
+    const size_t len = (LEN);                                                  \
+    size_t avl = len;                                                          \
+    const size_t buf_len = SKL_TEST_STATIC_DATA_LEN(BUF);                      \
+    while (avl > 0) {                                                          \
+      size_t vl = avl > buf_len ? buf_len : avl;                               \
+      memcpy(buf, SKL_TEST_STATIC_DATA(BUF), vl * sizeof(TYPE));               \
```
hdelassus, line 790: Is `BUF` sometimes expected to be an rvalue? Otherwise if we require it to be an lvalue, why not just set its value to BUF_data and avoid this whole memcpy?

I believe seeing an example in action would answer some of my possibly dumb questions.

ericlove: The point of this code is to _repeat_ the source array into `buf` as many times as necessary to fill it.

test/skl-test.h
ericlove: You're right: this was leftover from before I separated the use of the macro out from an `if` statement into a conditional macro expansion. So really all the guards around the main definition can go away.

test/skl-test.h
Aaron-Hutchinson: They're necessary *in the random case*, but unused in the others. That's the point I was trying to make.

ericlove: Right but the caller to `SKL_TEST_INIT` doesn't know which mode is set, so there needs to be a uniform interface.

Aaron-Hutchinson: I feel like we're going in circles here. The mode is passed as a compile definition to the harness, so it does know which mode is set. Why are we forcing a uniform interface between calls that have drastically different parameters?

test/skl-test.h
```diff
@@ -740,13 +747,98 @@ enum {
     }                                                                          \
   }
```
hdelassus, line 748: Is this definition meant to be entirely removed once other harnesses are changed to use `SKL_TEST_INIT` in lieu of `skl_test_init_*`?

test/skl-test.h
ericlove: Sorry, what I really mean is that the calling harness _could_ know, but we don't want to force it to check -- why impose additional logic on every user? That's the reason for uniformity of parameters.

test/skl-test.h
ericlove: Yes, that is the idea.

test/skl-test.h
ericlove: I have found a compromise: [this commit](https://github.com/sifive/skl/pull/34/changes/2b515f690c8bf0fb610fee1495cacf258298a5a6) drops the `MIN`/`MAX` arguments in favor of an `MSUFFIX` (`F32`, `F16`, etc.) that, while still only used in the non-`STATIC` case, is less misleading.

hdelassus: `MSUFFIX` can be determined by `TYPE` already, right? Can we drop it?

ericlove: That would require adding a new set of macros to translate between `float` and `F32` so I think it's just more convenient to do it this way.

Aaron-Hutchinson (approved):

PR #36, opened 2026-02-17
Allow customization of GEMM performance reports

hdelassus:
The default performance report for GEMM benchmarks was set to report_perf_mpc in the benchmark source files themselves.  This patch makes this setting conditional on the macro SKL_TEST_PERF_REPORT not already being defined so that a build flow may use another reporting function.


Aaron-Hutchinson (approved):

ericlove (approved):
LGTM, thanks for the fix, and sorry for not catching this initially.

PR #38, opened 2026-02-18
Use shared benchmark macro in logistic harnesses

hdelassus:
Standardize logistic benchmarks by using the SKL_BENCHMARK_RUN macro.


ebavier (approved):
LGTM, Thanks!

PR #39, opened 2026-02-18
Use shared benchmark macro in exp harnesses

hdelassus:
Standardize exp benchmarks by using the SKL_BENCHMARK_RUN macro.


ebavier (approved):
Thanks!

PR #40, opened 2026-02-18
Use shared benchmark macro in depthwise conv2d harnesses

hdelassus:
Standardize depthwise 2D convolution benchmarks by using the SKL_BENCHMARK_RUN macro.


peichialin (approved):
Look good to me. Thanks for help refine the tests!

PR #41, opened 2026-02-18
Use shared benchmark macro in transpose harnesses

hdelassus:
Standardize transpose benchmarks by using the SKL_BENCHMARK_RUN macro.


myeh01 (approved):
LGTM

PR #42, opened 2026-02-23
[Im2Row] Create standalone kernels

peichialin:
1. The im2row kernels are placed under `src/im2row/`. Please start the review from `src/im2row/README.md`.
2. The generic version of scalar API is untyped for functional verification, which may be moved out of src folder in near future.
3. Distinct rvv APIs for e32/e16/e8 are created to aligned with other kernels in skl. (although they share the same underlying implementations)
4. Only e32 test is created now. We can add the tests of other types once we agree with having distinct type-safe APIs.
5. We originally rely on the memcpy and memset in libc with auto-vectorization. If we would like to remove the dependency of `string.h`, we have to prepare our own rvv implementation in skl.

TO: Extract padding from innermost routines.


example/conv2d/conv2d_nhwc_f32.c
```diff
@@ -87,32 +90,38 @@ void conv2d_io_nhwc_filter_hwio_im2row_gemm_f32_f32_f32_zve32f_x390(
...
     const float *in_batch_tile = input + in_offset;
 
-    float *im2row_tile = im2row + output_row_offset;
+    float *row_tile = gemm_input + output_row_offset;
 
-    extract_patch_to_row_hwc(
-        in_w_origin, in_h_origin, input_height, input_width, input_channel,
-        filter_height, filter_width, dilation_width, dilation_height,
-        in_batch_tile, im2row_tile, 0, patch_begin_coord, k_len, sizeof(float));
+    skl_im2row_hwc_zve32x(
```
adam1029: I am not fully understand why the prefix is `zve32x` different from  `conv2d_io_nhwc_filter_hwio_im2row_gemm_f32_f32_f32_zve32f_x390`, the input & output tensor from  skl_im2row_hwc_zve32x are fp32?

ericlove: Because the implementation of im2row does not depend on the format of the the tensor elements, only the element width, so is implemented with "untyped" elements just like transpose.

PR #44, opened 2026-02-24
[Test] Add SKL_TEST_MAIN to optionally define main() in harnesses

ericlove:
Add a new macro parameter for test/benchmark harnesses, `SKL_TEST_MAIN`, which the latter shall use to decide whether or not `int main() {}` is defined or not.  This allows the harness code to be called by a different C `main()` in client apps that need to do their own initialization, or in binaries that combine multiple SKL test harnesses together and would otherwise have a symbol conflict.

The example usage is shown with the `gemm_f32rc_f32rc_f32rc.c` benchmark, which has been converted to use it.  Once agreed on, all others will be converted over time to follow the same pattern.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -304,7 +304,12 @@ function(skl_add_test TEST_NAME TEST_SRC OPTS)
     ${SKL_TEST_COMPILE_OPTIONS}
     ${OPTS}
   )
-  target_compile_definitions(test-${TEST_NAME} PRIVATE SKL_TEST_NAME=${TEST_NAME})
+  target_compile_definitions(
+    test-${TEST_NAME}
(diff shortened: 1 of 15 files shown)


hdelassus (dismissed):
Looks like we're slowly moving towards a test harness _library_, which I'm not against. I'm a bit uncomfortable approving this PR because macros are already quite confusing as they are, and this isn't making our tests an actual test library, but I think we can at least give this a try.

Aaron-Hutchinson:
Agreed with @hdelassus. This seems like more motivation to move toward a driver based approach so that the harnesses don't have to define `main` at all and a client app can supply its own driver if desired.

ericlove:

PR #45, opened 2026-02-24
Fix compiler errors related to #32

myeh01:
This PR fixes overlength string errors introduced by #32. #32 attempted to specialize the inner loops of the f8 and i8 Xsfmm GEMM kernels for the `tm0 == tn0` case by inserting the specialization directly into the inline assembly block, dispatching between the general and specialized versions using branch instructions. But, for the 1x3 and 1x4 kernels, this caused the inline assembly string to be too long. This PR addresses the issue by using a C `if` statement to dispatch between two versions of the inner loop inline asm block for the 1x3 and 1x4 kernels at the cost of increased code size.


JerryShih (approved):
What's the maximum string length?

myeh01:
> What's the maximum string length?

The error message said 4095 characters.

PR #47, opened 2026-02-26
[Test] SKL_TEST_LOG() and use in SKL_TEST_INIT

ericlove:
For debugging purposes it is often useful to have some logging output from the test harnesses. This adds `SKL_TEST_LOG`, which is a wrapper around `printf` that can be enabled or disabled depending on whether `SKL_TEST_LOG_LEVEL` is defined (and > 0).

As an initial use case, it is added to the two `SKL_TEST_INIT` macro definitions to indicate which initialization mode and other parameters were configured.  For example, in a GEMM benchmark:
```
SKL TEST: Init a (0x0000000040000000 +256): range 0.000000 - 1.000000 (mode 1)
SKL TEST: Init b (0x0000000040001000 +256): range 0.000000 - 1.000000 (mode 1)
SKL TEST: Init c (0x0000000040002000 +256): range 0.000000 - 1.000000 (mode 1)
```

### DIFF
--- a/test/skl-test.h
+++ b/test/skl-test.h
@@ -14,7 +14,11 @@
 #include <string.h>
 
 #if defined(SKL_TEST_LOG_LEVEL) && SKL_TEST_LOG_LEVEL >= 1
-#define SKL_TEST_LOG(...) printf(__VA_ARGS__)
+#define SKL_TEST_LOG(...)                                                      \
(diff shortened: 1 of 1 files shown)


Aaron-Hutchinson (approved):

ericlove:

PR #48, opened 2026-02-26
[Test] Rename to SKL_TEST_MEMALIGN

ericlove:
The previous `SKL_TEST_MALLOC` macro's API is really that of `[posix_]memalign(size_t alignment, size_t size)`, so rename to reflect that reality and avoid confusion.  This PR changes the existing use of the old macro in GEMM tests, and converts all remaining GEMM tests to use the custom allocator if defined.


ericlove:

PR #50, opened 2026-02-27
[GEMM] Add missing _clp wrapper & test for xsfvqdotq GEMM

ericlove:
When the CLP-tuned variant of the Xsfvqdotq GEMM kernel was added, a corresponding wrapper + test were not, so this PR simply adds them.


myeh01 (approved):
LGTM

PR #51, opened 2026-03-03
[GEMM] Add packed-A Xsfvqdotq GEMM kernels

myeh01:
This PR adds a packed-A version of the Xsfvqdotq GEMM kernels that allows A to have general `rsa1` and `csa1`. The API of the `1xe32m4` Xsfvqdotq kernel is updated to include `csa1` as a parameter (instead of hardcoding it to `4`), making it consistent with the APIs of the `6xe32m4` and `lt6xe32m4` kernels.


src/gemm/xsfvqdotq/gemm_i8_i8pc_i32_xsfvqdotq.c
```diff
@@ -873,7 +873,7 @@ SKL_FUNC_PRIVATE void skl_mm_lt6xe32m4_aligned_i8i32_vqdotvx(
 }
 
 SKL_FUNC_PRIVATE void skl_vm_1xe32m4_i8i32_vqdotvx(size_t n, size_t k,
-                                                   const int8_t *a,
+                                                   const int8_t *a, size_t csa1,
```
peichialin: I forgot why `csa1` was promoted to an arugment of prviate functions.
If I understand correctly, its value can only be `k0` (= 4), which is the hard constraint of using vqdotq ISA.
Is it proper to declare a constant `csa1` local variable instead?

myeh01: For the case when A is row major, `csa1` should be `k0`, but in general, `csa1` can be any value `>= k0`. Adding `csa1` as an argument allows the kernel to support `A` matrices with more general layouts without significantly impacting performance.

peichialin: Thanks for the clarification. That makes sense to me!

peichialin (dismissed):

ericlove (commented):
Seems like reasonable flexibility in exchange for changing an `addi` to `add`, but: why doesn't this change the public API? Is the purpose to someday allow packed A matrices as well, but not at the time of this PR?

myeh01:
> Seems like reasonable flexibility in exchange for changing an `addi` to `add`, but: why doesn't this change the public API? Is the purpose to someday allow packed A matrices as well, but not at the time of this PR?

I made this change mainly since I noticed there was a discrepancy in the APIs of the private kernels (this one did not include `csa1` while the others did). If we'd like to update the public API or add a packed version of the kernel, I'm happy to do that as well.

Edit: or if we don't have a need for `csa1`, we can also get rid of it in all the APIs.

ericlove (dismissed):
> I made this change mainly since I noticed there was a discrepancy in the APIs of the private kernels (this one did not include csa1 while the others did). If we'd like to update the public API or add a packed version of the kernel, I'm happy to do that as well.

No need to make any changes, I was just trying to confirm my understanding.  Happy to approve this for the sake of internal consistency.

ericlove (dismissed):

myeh01:
This will probably need one more review. I'm going to remove the two kernels `skl_gemm_i8_i8pc_i32_xsfvqdotq` and `skl_gemm_i8_i8pc_i32_xsfvqdotq_x390_clp` and just have
`skl_gemm_i8rcp_i8pc_i32_xsfvqdotq` and `skl_gemm_i8rcp_i8pc_i32_xsfvqdotq_x390_clp`, noting in the documentation that row-major A can be used directly with `rsa1 = rsa` and `csa1 = 4`. I'm currently rewriting the documentation so it's more clear.

JerryShih:
@myeh01 
thx!
it looks good.
but before this pr is merged, i would like to try this with iree to make sure it works with the iree packA/B layout.

JerryShih (changes requested):
I have a quick test with iree. This new interface works.

But I have another issue about memcpy(). Could we remove the usage of `memcpy()` in this pr?
You suggested to use `__builtin_memcpy` instead. I don't know what's the actual code for this.
If it will use the impl as memcpy, please still check the final asm code.
In my experience, only linux-based stdlib has `rvv` optimized memcpy impl. It will use scalar impl in baremetal env.

ericlove:
In addition to what @JerryShih said, libc dependencies are officially disallowed in SKL, so we need to remove them anyways.  This is a good chance to move forward on that fix.  Thanks!

myeh01:
@JerryShih I have updated `skl_mm_6xe32m4_i8i32_vqdotvx` to use `__builtin_memcpy`. `skl_mm_6xe32m4_i8i32_vqdotvx` will most likely not get good performance because it supports misaligned 32-bit accesses from A. For each 32-bit load from A, the assembly shows 4 `lbu`'s with some `slli` and `or` instructions to combine them into a single register.

JerryShih (dismissed):

JerryShih:
@myeh01 
will we have the L1 cache hint for mat_a as rvv_matmul ukernel?
```
#if defined(__riscv_zihintntl)
#define NTL_P1 "ntl.p1 \n\t"
#else
#define NTL_P1
#endif
```

myeh01:
> @myeh01 will we have the L1 cache hint for mat_a as rvv_matmul ukernel?
> 
> ```
> #if defined(__riscv_zihintntl)
> #define NTL_P1 "ntl.p1 \n\t"
> #else
> #define NTL_P1
> #endif
> ```

Yes, I'm working on it right now so we can remove the CLP-tuned kernel. But, it will be done in a follow-up PR.

JerryShih (dismissed):

myeh01:
Need a reapproval on this after rebasing.

ericlove (approved):

src/gemm/xsfvqdotq/gemm_i8rcp_i8pc_i32_xsfvqdotq.h
```diff
@@ -0,0 +1,176 @@
...
+ * pre-allocated with size ((k + 3) / 4) * rsb1 bytes, then
+ * ```
+ * skl_pack_a_i8_xsfvqdotq(m, k, a, rsa, a_pack, rsa1, csa1);
+ * skl_pack_b_i8_xsfvqdotq(k, n, b, rsb, b_pack, rsb1);
+ * skl_gemm_i8rcp_i8pc_i32_xsfvqdotq(m, n, k, alpha, a_pack, rsa1, csa1, b_pack,
+ *                                   rsb1, beta, c, rsc);
+ * ```
+ * is equivalent to scalar call:
+ * ```
+ * skl_gemm_i8_i32_scalar(m, n, k, alpha, a, rsa, b, rsb, beta, c, rsc);
```
hdelassus, line 73: I noticed there seems to be no such scalar reference kernel. We might want to spell out the full `skl_gemm_i8rcprc_i8rcprc_i32rcprc_scalar` call.

src/gemm/xsfvqdotq/gemm_i8rcp_i8pc_i32_xsfvqdotq.h
```diff
@@ -0,0 +1,176 @@
...
+ * pre-allocated with size ((k + 3) / 4) * rsb1 bytes, then
+ * ```
+ * skl_pack_a_i8_xsfvqdotq(m, k, a, rsa, a_pack, rsa1, csa1);
+ * skl_pack_b_i8_xsfvqdotq(k, n, b, rsb, b_pack, rsb1);
+ * skl_gemm_i8rcp_i8pc_i32_xsfvqdotq_x390_clp(m, n, k, alpha, a_pack, rsa1,
+ *                                            csa1, b_pack, rsb1, beta, c, rsc);
+ * ```
+ * is equivalent to scalar call:
+ * ```
+ * skl_gemm_i8_i32_scalar(m, n, k, alpha, a, rsa, b, rsb, beta, c, rsc);
```
hdelassus, line 146: Same thing here.

PR #55, opened 2026-03-06
[Softmax] Updates Xsfvfbfa intrinsics.

ebavier:
Xsfvfbfa "SiFive custom additional BF16 vector compute support" intrinsics were previously experimental and in flux.  Intrinsics have now stabilized with better handling of ambiguous cases and names more consistent with existing vector intrinsics.
This updates affected kernels.  Will require an up-to-date toolchain.

### DIFF
--- a/src/softmax/softmax_bf16_xsfvfbfa.c
+++ b/src/softmax/softmax_bf16_xsfvfbfa.c
@@ -1,4 +1,4 @@
-// Copyright 2025 SiFive, Inc.
+// Copyright 2026 SiFive, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
 #if !defined(__riscv_xsfvfbfa)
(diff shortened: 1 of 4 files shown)


src/softmax/softmax_bf16_xsfvfbfa.c
```diff
@@ -1,4 +1,4 @@
-// Copyright 2025 SiFive, Inc.
+// Copyright 2026 SiFive, Inc.
```
hdelassus, line 1: If you update the copyright here, please also update it in the other files.

hdelassus (dismissed):
Please add a short description, otherwise LGTM. IIRC that's because the BF16 RVV intrinsics were not yet stable, and now the API has stabilized but was changed, right?

hdelassus (approved):

PR #56, opened 2026-03-06
[Test] Refactor test suite into driver, harness, & test [DRAFT]

ericlove:
This PR introduces a new test driver framework that replaces the legacy monolithic test format with a new three-layer structure, and illustrates its usage with the FP32 (unpacked) GEMM kernels.

Now, instead of defining a single test per kernel with `-DPARAM=X` in the top-level `CMakeLists.txt`, each kernel has a test suite that lists tests and benchmarks in a `.c` file.  The old `skl_add_test` CMake command is now renamed `skl_add_test_legacy` to make room for the new system, and allow migration.

Please help review the design.

# New Test Architecture

Tests and benchmarks are now divided into the following components:

1. **Driver** ([`skl-test-driver.c`](https://github.com/sifive/skl/blob/ericl/test/driver-rethink/test/skl-test-driver.c)) - Environment-specific test execution engine (provides memory allocation, performance counters, I/O)
2. **Harness** (e.g., [`gemm/gemm_f32rc_f32rc_f32rc.c`](https://github.com/sifive/skl/blob/ericl/test/driver-rethink/test/gemm/gemm_f32rc_f32rc_f32rc.c) and [header](https://github.com/sifive/skl/blob/ericl/test/driver-rethink/test/gemm/gemm_f32rc_f32rc_f32rc.h)) - Operation-specific test logic shared across all kernel variants in a family (init, verify, report, cleanup callbacks)
3. **Test Suite** (e.g., [`gemm/xsfmm/skl_gemm_*.c`](https://github.com/sifive/skl/blob/ericl/test/driver-rethink/test/gemm/xsfmm/skl_gemm_a1b01_f32c_f32_f32_xsfmm32a32f.c)) - Kernel-specific test definitions (array of test configs + execute callback)

See [README.md](https://github.com/sifive/skl/blob/ericl/test/driver-rethink/test/README.md) for a more detailed description of the new framework.

# Status & Roadmap

This PR only converts FP32 GEMM tests to the new framework as an example.
I would like to hear everyone's feedback and suggestions now based on this initial example.
Once we all agree on the design, I will ask individual owners of each kernel family to migrate their tests to this new framework.

Ultimately, we want to eliminate the legacy `skl-test.h` framework, and possibly that file itself unless we keep some utility functions in there.


test/CMakeLists.txt
```diff
@@ -0,0 +1,61 @@
+set(SKL_TEST_LOG_LEVEL 1)
+
+function(skl_add_test_new TEST_NAME HARNESS_SRC TEST_SRC OPTS)
+    add_executable(
+        test-${TEST_NAME}
+        skl-test-driver.c
```
Aaron-Hutchinson: Is `skl-test-driver.c` being recompiled for every test out of simplicity of this initial mockup, or is there something in the design necessitating this?

test/CMakeLists.txt
```diff
@@ -0,0 +1,61 @@
...
+        PRIVATE ${SKL_COMPILE_OPTIONS} ${SKL_TEST_COMPILE_OPTIONS} ${OPTS}
+    )
+    target_compile_definitions(
+        test-${TEST_NAME}
+        PRIVATE SKL_TEST_NAME=${TEST_NAME}
+    )
+    target_compile_definitions(
+        test-${TEST_NAME}
+        PRIVATE SKL_TEST_LOG_LEVEL=${SKL_TEST_LOG_LEVEL}
+    )
```
Aaron-Hutchinson: I think we can merge these calls.

test/gemm/gemm_f32rc_f32rc_f32rc.h
```diff
@@ -0,0 +1,60 @@
...
+ * @brief Configurable parameters for GEMM tests.
+ */
+typedef struct {
+  size_t m, n, k;
+  float alpha;
+  size_t rsa, csa;
+  size_t rsb, csb;
+  float beta;
+  size_t rsc, csc;
+  bool benchmark;
```
Aaron-Hutchinson: In the current design this `benchmark` field seems out of place here compared to the rest since it's a more of a driver parameter. Was this mistakenly included?

test/gemm/gemm_f32rc_f32rc_f32rc.h
```diff
@@ -0,0 +1,60 @@
...
+extern const char *name;
+
+/**
+ * @brief Derived parameters & buffers for GEMM tests.
+ *
+ * These values are computed from the parameters in `tests` and should not be
+ * modified directly.
+ */
+typedef struct {
+  size_t rsa, csa, rsb, csb, rsc, csc;
```
Aaron-Hutchinson: The overlapping parameters (strides) between this struct and `params_t` above seems strange and error prone.

test/skl-test-driver.h
```diff
@@ -0,0 +1,166 @@
...
+typedef enum {
+  skl_test_init_mode_random,
+  skl_test_init_mode_static,
+  skl_test_init_mode_seq,
+} skl_test_init_mode_t;
+
+typedef struct {
+  const char *name;
+  size_t id;
+  void *data;
```
Aaron-Hutchinson: I might be misunderstanding how everything fits together here. It looks like all tests defined in the test file have to use the same `skl_test_t`. Shouldn't all/most of these parameters be configurable within each test? At the very least I would think the test parameters would include things like `warmup`, `benchmark`, `verify`, `init_mode`, etc. The test data already contains a separate `benchmark` field, which is suggesting to me that these knobs need to be tunable from within the test data.

My suggestion is to basically invert the dependency here by removing this `void* data` field and adding a `skl_test_t` field into the test-specific parameters.

test/skl-test-driver.c
```diff
@@ -0,0 +1,175 @@
...
+
+    printf("\nResults:\n");
+
+    // Rerport test performance, if benchmarking is enabled
+    if (t.benchmark) {
+      uint64_t cycles = cycle1 - cycle0;
+      uint64_t instret = instret1 - instret0;
+      skl_test_driver_result_uint64_t(&t, "cycles", cycles);
+      skl_test_driver_result_uint64_t(&t, "instret", instret);
+      skl_test_report(&t, cycles, instret);
```
Aaron-Hutchinson: These values should probably be passed in as a struct so that these metrics can be expanded in the future if needed without changing the function signature.

test/gemm/xsfmm/skl_gemm_a1b01_f32c_f32_f32_xsfmm32a32f.c
```diff
@@ -0,0 +1,109 @@
...
+     .beta = 1.f,
+     .rsa = 1,
+     .benchmark = false},
+};
+const size_t num_tests = sizeof(tests) / sizeof(tests[0]);
+const char *name = "skl_gemm_a1b01_f32c_f32_f32_xsfmm32a32f";
+
+int skl_test_execute(skl_test_t *t) {
+  const params_t *p = &tests[t->id];
+  test_data_t *d = (test_data_t *)t->data;
```
Aaron-Hutchinson: Couple points here:

1. There's 3 different structs in play here and it's a bit difficult to understand what role each one is playing. Intuitively I would expect the struct being passed to the function to contain most of the relevant data directly.
2. It's a bit strange to me that this code requires knowledge of the test id and has to lookup data based on it. Shouldn't the input to the function already contain all of the data for the test currently being processed?

test/gemm/xsfmm/skl_gemm_a1b01_f32c_f32_f32_xsfmm32a32f.c
```diff
@@ -0,0 +1,109 @@
...
+    {.m = 256,
+     .n = 256,
+     .k = 256,
+     .alpha = 1.f,
+     .beta = 1.f,
+     .rsa = 1,
+     .benchmark = false},
+};
+const size_t num_tests = sizeof(tests) / sizeof(tests[0]);
+const char *name = "skl_gemm_a1b01_f32c_f32_f32_xsfmm32a32f";
```
Aaron-Hutchinson: It would be great if these didn't need to be specified somehow. Can the driver just compute `num_tests` directly?

I don't see a good way to eliminate `name`. It could be moved into the above data but then would have to be specified in each test, which isn't ideal either.

test/skl-test-driver.c
```diff
@@ -0,0 +1,175 @@
...
+  (void)argv;
+  size_t num_tests;
+  CHECK_RESULT(skl_test_harness_init(&num_tests));
+
+  // Main loop over all tests
+  for (size_t i = 0; i < num_tests; ++i) {
+    skl_test_t t = default_test;
+    t.id = i;
+    printf("------------------------------------------------\n");
+    printf("Running test [%zd]\n", t.id);
```
Aaron-Hutchinson: Is it a good idea to include bare `printf`s? It might be better to wrap every print statement in some kind of log function with varying levels for finer control on the verbosity. It looks like this was done in some places but not others.

Aaron-Hutchinson (commented):
Thanks so much for putting this together and trying to address some fundamental issues in the test suite. Overall I think this is a huge step in the right direction.

I've mostly limited my current review to the high level structure being proposed here. My primary concern at the moment is the relationships between `skl_test_t`, `test_data_t`, and `params_t`. See my comments for details.

test/skl-test-driver.h
ericlove: Yes, the idea is that each test harness _can_ override these.  But they are in this struct because they are the only ones that the driver/test library know about.  Everything else is harness-specific, so gets lumped into `data`.

ericlove:
Okay, I think I have addressed many of @Aaron-Hutchinson 's concerns and in the process refined the harness structure considerably.

Now, each _test_ is defined by a new `skl_test_t` struct whose components are divided between test-specific _harness settings_ (`void *` type to allow arbitrary payloads), _driver settings_ (initialized to default values by the driver but overridable by the harness/tests), and a _test suite_ and test id:
```c
typedef struct {
  const skl_test_suite_t *suite; /**< Test suite (set by driver before init) */
  const size_t id; /**< Test ID within suite (set by driver before init) */
  void *harness;   /**< Harness-specific data (set by harness in init) */
  struct {
    bool warmup;          /**< Whether to run warmup iteration */
    bool verify;          /**< Whether to run verification */
    bool benchmark;       /**< Whether to report benchmark results */
    int log_level;        /**< Logging verbosity level */
    skl_test_init_t init; /**< Test data initialization config */
    skl_test_mem_t mem;   /**< Memory allocation config */
  } driver;
 /**< Driver settings (initialized by driver, modifiable by harness)
             */
} skl_test_t;
```

Also, I have proposed a new idiom by which harnesses structure their _parameters_ (configurable by a list of tests) and _context_ state, set during initialization and derived from the parameters:
```c
typedef struct {
  // Test parameters (set by the test in gemm_test_init())
  size_t m, n, k;
  float alpha;
  size_t rsa, csa;
  size_t rsb, csb;
  float beta;
  size_t rsc, csc;
  bool benchmark; // Overrides t->driver.{benchmark,verify,warmup}

  struct { // Derived parameters & buffers (private to the test harness)
    size_t a_len, b_len, c_len;
    float *a, *b, *c;
    double *a_wide, *b_wide;
    float *ref_c;
    double *bound;
  } ctx;
} gemm_harness_t;
```

test/gemm/gemm_f32rc_f32rc_f32rc.c
```diff
@@ -1,169 +1,109 @@
...
-  SKL_TEST_REQUIRE(status, csc == 1);
-  if (status) {
-    exit(status);
+static void compute_derived_params(gemm_harness_t *h) {
+  h->rsa = h->rsa ? h->rsa : h->k;
+  h->csa = h->csa ? h->csa : 1;
+  h->rsb = h->rsb ? h->rsb : h->n;
+  h->csb = h->csb ? h->csb : 1;
+  h->rsc = h->rsc ? h->rsc : h->n;
+  h->csc = h->csc ? h->csc : 1;
```
Aaron-Hutchinson: This is my personal preference so feel free to override it.

These implicit defaults don't seem like good practice to me. The benefit they bring is to reduce verbosity in the `tests` array. But there are some downsides: it's not clear from the `tests` array what these values will be, it's not clear they're even modifiable in the `tests` array without looking at the struct definition, and it prevents using the value `0` as an actual parameter (in cases it makes sense).

test/gemm/gemm_f32rc_f32rc_f32rc.c
```diff
@@ -1,169 +1,109 @@
...
-  SKL_TEST_REQUIRE(status, csb == 1);
-  SKL_TEST_REQUIRE(status, csc == 1);
-  if (status) {
-    exit(status);
-  }
-  skl_gemm_f32_f32_f32_zve32f_x390_clp(m, n, k, alpha, a, rsa, b, rsb, beta, c,
-                                       rsc);
+static void log_test_params(skl_test_t *t) {
+  gemm_harness_t *s = (gemm_harness_t *)t->harness;
+  SKL_TEST_LOG_SZ(t, 1, "m", s->m);
```
Aaron-Hutchinson: Nice!

Can we also give these integer values meaningful names? Currently it's not clear what each value means or when each should be used. I'd propose doing something along the lines of:
```
#define SKL_LOG_ERROR 1
#define SKL_LOG_WARN 2
#define SKL_LOG_INFO 3
#define SKL_LOG_DEBUG 4
```
Then the macro's implementation can ensure all log messages whose `SKL_LOG_*` is not greater than the global log level are printed.

test/gemm/xsfmm/skl_gemm_a1b01_f32c_f32_f32_xsfmm32a32f.c
```diff
@@ -0,0 +1,67 @@
...
+    {.m = 128, .n = 128, .k = 128, .alpha = 1.f, .beta = 0.f, .rsa = 1, .benchmark = true},
+    {.m = 128, .n = 128, .k = 128, .alpha = 1.f, .beta = 1.f, .rsa = 1, .benchmark = true},
+
+    // Verification tests
+    GEMM_A1B01_RSA1_TESTS, // from gemm_f32rc_f32rc_f32rc.h
+};
+// clang-format on
+
+int gemm_test_init(skl_test_t *t) {
+  t->harness = (void *)&tests[t->id];
```
Aaron-Hutchinson: I'm still convinced that this code should not need to perform any lookups based on the test id. It should be the responsibility of the driver to retrieve this data IMO.

test/skl-test-driver.h
```diff
@@ -0,0 +1,520 @@
...
+  const size_t id; /**< Test ID within suite (set by driver before init) */
+  void *harness;   /**< Harness-specific data (set by harness in init) */
+  struct {
+    bool warmup;          /**< Whether to run warmup iteration */
+    bool verify;          /**< Whether to run verification */
+    bool benchmark;       /**< Whether to report benchmark results */
+    int log_level;        /**< Logging verbosity level */
+    skl_test_init_t init; /**< Test data initialization config */
+    skl_test_mem_t mem;   /**< Memory allocation config */
+  } driver; /**< Driver settings (initialized by driver, modifiable by harness)
```
Aaron-Hutchinson: Why not let the test file set the driver values directly within the test array? Was it a design choice to not put it there, or some technical limitation?

test/gemm/xsfmm/skl_gemm_a1b01_f32c_f32_f32_xsfmm32a32f.c
```diff
@@ -0,0 +1,67 @@
...
+  return 0;
+}
+
+int skl_test_suite_init(skl_test_suite_t *suite) {
+  suite->name = "skl_gemm_a1b01_f32c_f32_f32_xsfmm32a32f";
+  suite->num_tests = sizeof(tests) / sizeof(tests[0]);
+  return 0;
+}
```
Aaron-Hutchinson: This file could probably just define the `skl_test_suite_t` object directly, and simultaneously allow the driver to retrieve the i-th test:
```
// skl_gemm_a1b01_f32c_f32_f32_xsfmm32a32f.c:
skl_test_suite_t skl_test_suite = {
  .name = "skl_gemm_a1b01_f32c_f32_f32_xsfmm32a32f",
  .num_tests = sizeof(tests) / sizeof(tests[0]),
  .test_size = sizeof(gemm_harness_t),
  .tests = &tests
};

// skl-test-driver.c:
extern skl_test_suite_t skl_test_suite;
```
The driver can iterate `.tests` in increments of `.test_size`.

test/skl-test-driver.h
```diff
@@ -0,0 +1,520 @@
...
+ * @param t - Test context
+ * @param result - Performance measurement results (cycles and instructions)
+ *
+ * Called after execution (if benchmarking is enabled) to allow the harness
+ * to compute and report test-specific metrics (e.g., MACs/cycle, throughput).
+ * The driver has already reported raw cycle and instruction counts.
+ *
+ * @return 0 on success, non-zero on error
+ */
+int skl_test_report(skl_test_t *t, skl_test_result_t result);
```
Aaron-Hutchinson: Maybe we should put the `skl_test_result_t` into `skl_test_t`. This function is a bit asymmetric compared to the others due to the extra argument.

Additionally the name `skl_test_result_t` feels a bit misleading since I would assume it contains some kind of pass/fail value. Maybe change to something like `skl_test_metrics_t`.

Aaron-Hutchinson (commented):
Thanks for the changes! I'm still asking for some high level clarification and potential design adjustments.

test/gemm/gemm_f32rc_f32rc_f32rc.c
ericlove: As a compromise, how about my latest change where the strides are only overwritten for a given matrix if _both_ are zero?

test/gemm/gemm_f32rc_f32rc_f32rc.c
ericlove: Agreed, though I skipped the WARN (not sure how to use it).

test/gemm/xsfmm/skl_gemm_a1b01_f32c_f32_f32_xsfmm32a32f.c
ericlove: I was skeptical of this, but enforcing an array of harness structs did simplify the contract between the GEMM harness and its tests, so I think this is an overall win. I deleted the `id` field too since it can be computed by pointer arithmetic if needed by the driver.

test/skl-test-driver.h
ericlove: There is a technical limitation: since I was determined to avoid and type dependence of the driver on the harness, it's just not possible to merge these into a unified struct that would allow the driver to access fields in set in the harness struct directly.

test/gemm/xsfmm/skl_gemm_a1b01_f32c_f32_f32_xsfmm32a32f.c
ericlove: I chose to do this in `skl_test_suite_init` rather than using a linker symbol dependency.

test/skl-test-driver.h
ericlove: I think it's fine to have a different signature here to emphasize that the driver is passing new info here that's only used in one function.  But I do think it made more sense to call this `skl_test_counters_t`.  What do you think?

ericlove (commented):
@Aaron-Hutchinson thanks for the re-review. I have responded to your comments below with my latest changes.

test/skl-test-driver.h
Aaron-Hutchinson: Agreed it'd be best to keep the driver independent of the harness. But that doesn't necessarily mean there's no way to make this happen.

Here's one option I can think of. Require that all harness structs (`gemm_harness_t` in the current example) position their `driver_t` value as the first struct member. Then the driver can safely reinterpret the full harness object as a driver configuration without having any knowledge of the rest of the struct. This doesn't seem too abnormal of a practice in C.

To set the default driver configuration, we could do something like:
```
// test-driver.h:
#define DEFAULT_DRIVER_CONFIG \
  .benchmark = false, \
  .verify = true, \
  // ...

// test.c:
gemm_harness_t tests[] = {
    { .driver = {
        DEFAULT_DRIVER_CONFIG,
        .benchmark = true // override
      },
      .m = 128,
      .n = 128,
      .k = 128
    },
};
```
which likely requires disabling a compiler warning, at least locally. This also prevents having to copy each override manually in the harness.

Maybe not the cleanest method, perhaps there's something better. But I don't think it's flexible or sustainable to duplicate each overridable parameter between structs and manually copy them within the harness.

test/skl-test-driver.h
Aaron-Hutchinson: Sounds good

test/skl-test-driver.h
ericlove: I don't think this approach works for a couple of reasons:

1. I don't want any part of the interface to depend on the binary layout of struct fields.  We could fix this with another mandatory harness function `skl_test_get_driver()` that returns a pointer to the driver struct, but that won't solve the next problem
2. The driver defaults should be set within the `driver.c` implementation, not by the `driver.h` interface which is common to all implementations.  You could do this with a driver function, `skl_test_driver_defaults()` that sets them, but then you couldn't use them in a static test description list like you've shown above.

Another possibility would be to place a `skl_driver_t` struct in every test, along with a `bool override` to determine whether or not it is used. But this does not allow selective override of the defaults on a field-by-field basis; it's all or nothing.  To the extent that harnesses will want to override driver settings, it is probably only for a very limited set of features (specific to one datatype in datagen, for example).

The one setting that _every_ test currently wants to override is benchmark/verify/warmup. I think that we could simplify this logic by collapsing these independent parameters into a single `benchmark` boolean, where benchmark implies (warmup and not verify) and (not benchmark) implies (verify and not warmup). If someone wants to muck around with warmup, they can write a customer driver.

Overall, these difficulties suggest the best option is still for each harness to define its own bespoke interface for overriding select driver settings.  The driver setting list can continue to grow without harnesses being aware of additions, because they likely only care about the things they explicitly wanted to make overridable.

Aaron-Hutchinson: > I don't want any part of the interface to depend on the binary layout of struct fields.

Could you expand on what the specific worry is? If there's an issue with what I proposed I'd like to understand it better.

> The driver defaults should be set within the `driver.c` implementation

I'm struggling to understand this. If the driver defaults are set in `driver.c` and most fields aren't overridable in the test struct, how can those fields ever be changed? Write a separate driver implementation with a different configuration?

Also, my use of the word "default" was probably a poor choice: if needed we could provide multiple "presets" that particular tests could choose between and override.

> The one setting that every test currently wants to override is benchmark/verify/warmup.

I don't really agree that this is the only value to be frequently changed. We already have use cases where we want to test/benchmark memory port vs CLP allocated buffers, as well as randomly generated vs static data. It's also useful to be able to add temporary tests with specific overrides during development and debugging.

ericlove: > Could you expand on what the specific worry is? If there's an issue with what I proposed I'd like to understand it better.

The position of fields within a struct is subject to various alignment and padding constraints, some of which are implementation-dependent. Unless we are writing code to process e.g. network packets, we should not be depending on memory layouts of this sort. This is a bad practice and recipe for disaster.

> I'm struggling to understand this. If the driver defaults are set in driver.c and most fields aren't overridable in the test struct, how can those fields ever be changed? Write a separate driver implementation with a different configuration?

The driver sets these in `main()`. More importantly, multiple drivers will implement this, but they will all share `skl-test-driver.h` which doesn't change between implementations. The defaults can be overridden because `skl_test_init()` is called after the driver sets them (or it should be at least).

> We already have use cases where we want to test/benchmark memory port vs CLP allocated buffers,

This can never be set by a test, because it doesn't know whether a driver has a CLP allocator or some other kind of allocator. That's a property of the system/driver.  They only way to change this is for _all_ tests/benchmarks simultaneously.  You set this once per binary based on the capabilities of your driver.

> as well as randomly generated vs static data.

This is also mostly useful as a universal, rather than test-specific setting.  Static data is just a way to get all the benchmarks to execute quickly due to constraints on the execution environment (such as the slowness of RTL simulation).

Some correctness tests will want to use static data as a way of introducing specific corner cases, but they can already do so by implementing their own custom override mechanism when they implement `skl_test_init()`.

It would be better if we could make a standard method across all harness designs, but I don't see a way of doing this for reasons mentioned above. The best we can do is to encourage standard idioms, enforced by reviewers on GitHub rather than by the interface itself.

Aaron-Hutchinson: > The position of fields within a struct is subject to various alignment and padding constraints, some of which are implementation-dependent.

This is generally true, but not for the first struct member. Assuming cppreference is a trustworthy source (emphasis mine):
> Within a struct object, addresses of its elements (and the addresses of the bit-field allocation units) increase in order in which the members were defined. **A pointer to a struct can be cast to a pointer to its first member** (or, if the member is a bit-field, to its allocation unit). Likewise, a pointer to the first member of a struct can be cast to a pointer to the enclosing struct. There may be unnamed padding between any two members of a struct or after the last member, but not before the first member. The size of a struct is at least as large as the sum of the sizes of its members. 

I can't think of a situation where we'd need to do this for any member other than the first, since any data the driver needs should be contained in the driver sub-struct.

ref: https://en.cppreference.com/w/c/language/struct.html

> This can never be set by a test, because it doesn't know whether a driver has a CLP allocator or some other kind of allocator. That's a property of the system/driver.

It can be set by the test if the test knows about the allocator. My impression was that the primary motivator for this design was that external applications would provide their own test file, at which point they could just `#include "clp_allocator.h"` so that the allocator function could be referenced within the test array.

I don't feel that swapping out the driver for a custom version in an external application is something that should happen very often. If we can make the test specification flexible enough I don't see many reasons to rewrite the driver.

ericlove: I concede that you are right about the C standard. However, I still don't like it.  I've played around with a number of different strategies to support easy defaults + overriding, and I don't really like any of them.

My latest idea is to have each harness struct (e.g. `gemm_harness_t`) contain an optional pointer to an "init" function that can access all fields of the `skl_test_t` struct, and is called at the beginning of `skl_test_init()` to override harness params, or driver params, based on test-specific settings.

I have to admit you're right that a test is distinct from a harness in that it can have knowledge of the driver.  So tests _could_ (but in my intended use case internally likely won't) define an init func like `bench_clp` and `bench_memport` etc. ands sets `driver.benchmark` and `driver.mem.aligned_alloc` appropriately.

Additionally, some tests for things like SFMM might want to use this as a way of setting some test parameters dynamically, such as scaling `m` and `n` by TE in GEMM.  It also let me remove the "default to row major" stride calculations from the general GEMM harness.

This might not be mutually-exclusive with your proposal, but I'm still reluctant to require a whole driver struct's initialization in every test declaration, and to require that these defaults be accessible from a header file (and thus hard-coded, rather than set in driver init, though this is a somewhat dubious desideratum).

Aaron-Hutchinson: A pointer to an init function is slightly better but IMO still not flexible enough since you'd have to write a separate init for each configuration you want to support.

> a whole driver struct's initialization in every test declaration, and to require that these defaults be accessible from a header file (and thus hard-coded, rather than set in driver init, though this is a somewhat dubious desideratum).

I was imagining the initialization in most cases would amount to just `.driver = { SKL_DRIVER_PRESET_TEST }`, where the macro hides most of the initialization. I'm also not sure what you mean by hard-coded, since if all the defaults aren't overridable then they're hard-coded somewhere anyway.

I've admittedly become biased toward my own proposal and am having a hard time considering anything that doesn't allow full driver customization in the test specification. I think we should bring in a third party to help iron out this part of the design.

test/skl-test-driver.h
```diff
@@ -0,0 +1,537 @@
...
+ * @brief Test context structure.
+ *
+ * Contains all state for a single test execution, including driver
+ * configuration and harness-specific data. The driver initializes this
+ * structure and passes it to all harness callback functions.
+ */
+typedef struct {
+  const skl_test_suite_t *suite; /**< Test suite (set by driver before init) */
+  void *harness; /**< Harness-specific data (set by harness in init) */
+  skl_test_steps_t steps;
```
Aaron-Hutchinson: Shouldn't this be a pointer to the harness' `steps` field? Currently we have to do a copy, which means there will be 2 different instances of the steps for each test and that seems potentially error prone.

test/skl-test-driver.h
```diff
@@ -0,0 +1,537 @@
...
+ */
+int skl_test_suite_init(skl_test_suite_t *suite);
+
+/**
+ * @brief Initialize a single test.
+ *
+ * @param t - Test context with suite and harness already set by the driver
+ *
+ * Called before each test execution. The harness should allocate and initialize
+ * test-specific data, and optionally modify test execution settings in t->steps
```
Aaron-Hutchinson: What's the incentive in allowing the harness to modify the execution steps? It's seems pretty unintuitive if what was specified in the test array isn't what actually ends up happening. I would actually go to the other extreme in proposing that the `steps` field should be `const`.

test/gemm/gemm_f32rc_f32rc_f32rc.c
```diff
@@ -1,169 +1,99 @@
...
 
-/* Input, output, reference, and error bound arrays */
-enum {
-  ALIGN = 4096,
-  CLEN = ((M - 1) * RSC + (N - 1) * CSC + 1),
-  ALEN = ((M - 1) * RSA + (K - 1) * CSA + 1),
-  BLEN = ((K - 1) * RSB + (N - 1) * CSB + 1),
-};
+  // Copy steps from harness to test context
+  t->steps = h->steps;
```
Aaron-Hutchinson: Nice! I'm happy with this alternative to what I was proposing. Thanks for revising this so many times to accommodate my insistence on customization of the steps.

test/gemm/gemm_f32rc_f32rc_f32rc.h
```diff
@@ -0,0 +1,57 @@
...
+#define GEMM_F32                                                               \
+  .a = {.min = -1.0f, .max = 1.0f, .mode = SKL_TEST_RANDOM},                   \
+  .b = {.min = -1.0f, .max = 1.0f, .mode = SKL_TEST_RANDOM},                   \
+  .c = {.min = -1.0f, .max = 1.0f, .mode = SKL_TEST_RANDOM}, .alpha = 1.0f
+
+#define GEMM_F32_TEST                                                          \
+  GEMM_F32, .steps = {.warmup = false, .verify = true, .benchmark = false}
+
+#define GEMM_F32_BENCH                                                         \
+  GEMM_F32, .steps = {.warmup = true, .verify = false, .benchmark = true}
```
Aaron-Hutchinson: These steps seem common enough to be shared by nearly every harness. Maybe put them into `skl-test-driver.h`?

test/CMakeLists.txt
```diff
@@ -0,0 +1,50 @@
+set(SKL_TEST_LOG_LEVEL SKL_TEST_LOG_INFO)
```
Aaron-Hutchinson: Can we make this overridable from the command line?

test/skl-test-driver.h
```diff
@@ -0,0 +1,537 @@
...
+#define SKL_TEST_BUFFER(TYPE)                                                  \
+  struct {                                                                     \
+    TYPE *data;                                                                \
+    size_t len;                                                                \
+    skl_test_init_mode_t mode;                                                 \
+    TYPE min, max;                                                             \
+    TYPE *static_data;                                                         \
+    size_t static_data_len;                                                    \
+    size_t region;                                                             \
+  }
```
Aaron-Hutchinson, line 366: Nice, I like this flexibility a lot!

The `static_data` field should maybe be `const TYPE *` to avoid individual tests in a suite accidentally modifying the data that subsequent tests intend to use.

What is `region` referring to?

Shouldn't each buffer be able to specify its own allocation and free functions?

test/CMakeLists.txt
```diff
@@ -0,0 +1,50 @@
...
+target_compile_definitions(
+    skl-test-driver
+    PRIVATE SKL_TEST_LOG_LEVEL=${SKL_TEST_LOG_LEVEL}
+)
+target_compile_options(
+    skl-test-driver
+    PRIVATE ${SKL_COMPILE_OPTIONS} ${SKL_TEST_COMPILE_OPTIONS}
+)
+
+function(skl_add_test_new TEST_NAME HARNESS_SRC TEST_SRC OPTS)
```
Aaron-Hutchinson: I don't really like the `new` suffix since once all tests are transitioned it will look very odd. What about changing the existing function to `skl_add_test_legacy` and this one can be `skl_add_test`?

test/CMakeLists.txt
```diff
@@ -0,0 +1,50 @@
...
+        COMMAND
+            ${SKL_TESTER} ${SKL_TESTER_OPTIONS} $<TARGET_FILE:test-${TEST_NAME}>
+    )
+endfunction()
+
+if(RISCV_ZVE32F)
+    skl_add_test_new(
+        skl_gemm_f32_f32_f32_zve32f_x390
+        gemm/gemm_f32rc_f32rc_f32rc.c
+        gemm/rvv/skl_gemm_f32_f32_f32_zve32f_x390.c
```
Aaron-Hutchinson: Could we make this relative to the SKL root so that it's easier to tell what is a source file and what is a harness?

test/gemm/gemm_f32rc_f32rc_f32rc.c
```diff
@@ -181,78 +111,73 @@ int check_error(void) {
...
-        return 1;
+  for (size_t i = 0; i < h->m; ++i) {
+    for (size_t j = 0; j < h->n; ++j) {
+      size_t idx = i * h->rsc + j * h->csc;
+      if (fabs((double)h->c.data[idx] - (double)h->ctx.ref_c[idx]) >
+          h->ctx.bound[idx]) {
+        skl_test_driver_error(
+            t, "result [%zu, %zu] (%f) != reference (%f) [bound = %f]\n", i, j,
+            h->c.data[idx], h->ctx.ref_c[idx], h->ctx.bound[idx]);
+        return 1; // Error
```
Aaron-Hutchinson: Would be nice to have a way to print all errors instead of just the first for debugging.

test/gemm/gemm_f32rc_f32rc_f32rc.c
```diff
@@ -1,169 +1,99 @@
...
-                                              const float *b, size_t rsb,
-                                              size_t csb, float beta, float *c,
-                                              size_t rsc, size_t csc) {
-  int status = 0;
-  SKL_TEST_REQUIRE(status, csa == 1);
-  SKL_TEST_REQUIRE(status, csb == 1);
-  SKL_TEST_REQUIRE(status, csc == 1);
-  if (status) {
-    exit(status);
+static void compute_derived_params(gemm_f32_t *h) {
```
Aaron-Hutchinson: There are a few functions in this file that seem like they'll be common to all gemm harnesses (or maybe non-packed ones). We could consider factoring the code out to be shared between them, but this can be done in a follow up PR.

Aaron-Hutchinson (commented):
Thanks for the many revisions on this refactor! I'm overall pretty happy with the current design, modulo some small details. I think this is ready for review from the broader team.

test/skl-test-driver.h
Aaron-Hutchinson: Your latest revision has satisfied my request here, so I'm resolving this conversation.

test/skl-test-driver.h
ericlove: I think its easier to hide all the details of alignment and which malloc function to use inside the driver, and just provide an opaque notion of memory region here.  This cleaned up the harness code considerably.  Most tests will just default to `region = 0`; only custom tests that know something about their driver will set it to nonzero.

ericlove: Agree with the const issue though.

test/skl-test-driver.h
ericlove: Not really my intention, probably just a holdover from previous iterations and find/replace.

test/gemm/gemm_f32rc_f32rc_f32rc.c
ericlove: This is an implementation detail that harness authors can fix later.

test/gemm/gemm_f32rc_f32rc_f32rc.c
ericlove: I think actually this shouldn't be a separate function at all, but I asked the LLM to solve a lint issue ("cognitive complexity too high") and this was a good default.  Let me see if I can merge this stuff back in.

test/CMakeLists.txt
ericlove: There are no source files...these are all harness or tests.

test/gemm/gemm_f32rc_f32rc_f32rc.h
ericlove: I can't think of a good generic name that is short enough so I'll let you deal with this later.

test/CMakeLists.txt
Aaron-Hutchinson: Oh. I obviously didn't read this carefully enough.

test/gemm/xsfmm/skl_gemm_a1b01_f32c_f32_f32_xsfmm32a32f.c
```diff
@@ -0,0 +1,115 @@
...
+ *  - Matrix B is row-major (csb == 1)
+ *  - Matrix C is row-major (csc == 1)
+ *  - Alpha must be 1.0
+ *  - Beta must be 0.0 or 1.0
+ *
+ * The kernel computes C = A * B (beta=0) or C += A * B (beta=1).
+ */
+
+#define TEST GEMM_F32RC_F32RC_F32RC_DEFAULTS, .warmup = false, .verify = true
+#define BENCH GEMM_F32RC_F32RC_F32RC_DEFAULTS, .warmup = true, .verify = false
```
Aaron-Hutchinson: With the function pointer struct added, can't `warmup` and `verify` be inferred based on whether the respective function pointer is non-null?

More to the point, I like the function pointer approach but it feels unnecessarily limited to not be able to specify each pointer in the test array. The pointers can go into a separate struct, and `TEST` and `BENCH` can set each field appropriately with NULL meaning skip.

ericlove: I gave up on that idea because you need to pass these pointers _outside_ of the config struct in order for the driver to even know the address of an init function capable of interpreting the `void *` harness type. That meant it had to go in the suite struct, so no way to NULL-ify pointers on a per-test basis.

Aaron-Hutchinson: Ah, I see. Then how about a suite-wide steps-getter function defined in the harness? Then the suite can be:
```
static skl_test_suite_t suite = {
    .name = "skl_gemm_a1b01_f32c_f32_f32_xsfmm32a32f",
    .num_tests = sizeof(tests) / sizeof(tests[0]),
    .test_size = sizeof(gemm_f32rc_f32rc_f32rc_t),
    .tests = tests,
    .steps = &gemm_f32rc_f32rc_f32rc_steps
}
```
with
```
const skl_test_steps_t* gemm_f32rc_f32rc_f32rc_steps(void* test) {
  return test->steps;
}
```
Yes it's another function in the suite interface, but it would makes things more modular and remove the redundant `warmup` and `verify` booleans.

test/skl-test-driver.c
Aaron-Hutchinson, line 1: Nit: all the other filenames use an underscore separator, so for consistency I think these should too.

test/CMakeLists.txt
```diff
@@ -0,0 +1,40 @@
...
+
+function(skl_add_test TEST_NAME HARNESS_SRC TEST_SRC OPTS)
+    add_executable(test-${TEST_NAME} ${HARNESS_SRC} ${TEST_SRC} skl-test-driver.c)
+    target_link_libraries(
+        test-${TEST_NAME}
+        PRIVATE ${SKL_LIBRARY} ${SKL_TEST_LIBS}
+    )
+    target_compile_options(
+        test-${TEST_NAME}
+        PRIVATE ${SKL_COMPILE_OPTIONS} ${SKL_TEST_COMPILE_OPTIONS} ${OPTS}
```
Aaron-Hutchinson: Shouldn't `SKL_COMPILE_OPTIONS` be omitted? I assumed these were options specific to SKL source files, which may be different than what's desired in the tests.

test/gemm/gemm_f32rc_f32rc_f32rc.c
```diff
@@ -166,78 +85,84 @@ int check_error(void) {
...
+  return 0; // Success
+}
+
+int gemm_f32rc_f32rc_f32rc_report(skl_test_t *t) {
+  gemm_f32rc_f32rc_f32rc_t *h = (gemm_f32rc_f32rc_f32rc_t *)t->harness;
+
+  size_t maccs = h->m * h->n * h->k;
+  float mpc = (float)maccs / (float)t->counters.cycles;
+
+#define INFO(fmt, ...) SKL_TEST_LOG(t, SKL_TEST_LOG_INFO, fmt, __VA_ARGS__)
```
Aaron-Hutchinson, line 129: As a good practice should this be `#undef`ed below?

test/skl-test-driver.h
```diff
@@ -0,0 +1,343 @@
...
+ * Contains all state for a single test execution, including driver
+ * configuration and harness-specific data. The driver initializes this
+ * structure and passes it to all harness callback functions.
+ */
+struct skl_test_t {
+  const skl_test_suite_t *suite; /**< Test suite (set by driver before init) */
+  void *harness; /**< Harness-specific data (set by harness in init) */
+  int log_level; /**< Logging verbosity level (set by driver before init) */
+  skl_test_counters_t counters; /**< Performance counters (updated by driver) */
+  skl_test_status_t status;     /**< Test pass/fail status */
```
Aaron-Hutchinson: What do you think about making `status` a struct that can potentially hold more useful debugging information than pass/fail? Maybe things like what stage failed, what might have caused the failure, etc. This would also make the data more flexible to change without affecting the driver interface, similar to `counters`.

test/gemm/gemm_f32rc_f32rc_f32rc.c
```diff
@@ -166,78 +85,84 @@ int check_error(void) {
...
+
+  INFO("M: %zd, N: %zd, K: %zd\n", h->m, h->n, h->k);
+  INFO("CSA: %zd, RSB: %zd, RSC: %zd\n", h->csa, h->rsb, h->rsc);
```
Aaron-Hutchinson, line 132: Is csa supposed to be rsa? Or, shouldn't this harness be printing all 6 strides?

test/gemm/gemm_f32rc_f32rc_f32rc.c
```diff
@@ -166,78 +85,84 @@ int check_error(void) {
...
-               ref_c[idx]);
-        return 1;
+  for (size_t i = 0; i < h->m; ++i) {
+    for (size_t j = 0; j < h->n; ++j) {
+      size_t idx = i * h->rsc + j * h->csc;
+      if (fabs((double)h->c.data[idx] - (double)h->ctx.ref_c[idx]) >
+          h->ctx.bound[idx]) {
+        skl_test_driver_error(
+            t, "result [%zu, %zu] (%f) != reference (%f) [bound = %f]\n", i, j,
+            h->c.data[idx], h->ctx.ref_c[idx], h->ctx.bound[idx]);
```
Aaron-Hutchinson: This code (and the original code) is missing a check for clobbered inter-stride values. Given that this harness uses row and column strides, the logic for these checks might be a bit complex.

We could push this off to a follow up PR... however I suspect all gemm harnesses will use this one as a template, which means this problem will multiply. Should we just address it now?

test/gemm/rvv/skl_gemm_f32_f32_f32_zve32f_x390.c
```diff
@@ -0,0 +1,95 @@
...
+static int execute(skl_test_t *t) {
+  const gemm_f32rc_f32rc_f32rc_t *h = (gemm_f32rc_f32rc_f32rc_t *)t->harness;
+
+  SKL_TEST_REQUIRE(t, h->csa == 1);
+  SKL_TEST_REQUIRE(t, h->csb == 1);
+  SKL_TEST_REQUIRE(t, h->csc == 1);
+  skl_gemm_f32_f32_f32_zve32f_x390(h->m, h->n, h->k, h->alpha, h->a.data,
+                                   h->rsa, h->b.data, h->rsb, h->beta,
+                                   h->c.data, h->rsc);
+  return (t->status == SKL_TEST_PASS) ? 0 : 1;
```
Aaron-Hutchinson: Shouldn't this just always return 0? I'm very confused why ~it's returning the opposite status as `t->status`~ a failure elsewhere should imply a failure here.

test/gemm/rvv/skl_gemm_f32_f32_f32_zve32f_x390.c
```diff
@@ -0,0 +1,95 @@
...
+}
+
+int main(void) {
+  // Set default strides: all matrices are row-major
```
Aaron-Hutchinson, line 95: Row-major is a hard requirement for this kernel. Maybe you meant "row stride is row length"?

(I would refer to this as "packed", but we use that term to mean something else...)

test/CMakeLists.txt
Aaron-Hutchinson, line 1: Nit: the top level CMakeLists (mostly) uses an indentation of 2 spaces, but this one uses 4. I like 2, but whatever we use should be consistent.

Aaron-Hutchinson (commented):
While converting one of the other harnesses to this paradigm I noticed a few other small things.

test/gemm/gemm_f32rc_f32rc_f32rc.c
ericlove: No, because I am not willing to spend any more time on this. I don't think this PR is a good place to worry about _features_ of any component, rather than _interfaces_.  This harness will be deleted anyways, and replaced with the one that generalizes to packed layouts, but I plan to have @myeh01 do that.

test/gemm/gemm_f32rc_f32rc_f32rc.c
ericlove: Yeah, this was probably an autocomplete fail after making the Xsfmm one.

test/gemm/gemm_f32rc_f32rc_f32rc.h
```diff
@@ -0,0 +1,58 @@
...
+ *  - Scalar coefficients ALPHA and BETA
+ *  - Row and column strides (RSA, CSA, RSB, CSB, RSC, CSC)
+ *  - The GEMM kernel function to test
+ *
+ * Matrix layouts:
+ *  - A is M x K with row stride RSA and column stride CSA
+ *  - B is K x N with row stride RSB and column stride CSB
+ *  - C is M x N with row stride RSC and column stride CSC
+ */
+
```
Aaron-Hutchinson, line 21: Missing a `#pragma once`

test/gemm/gemm_f32rc_f32rc_f32rc.h
```diff
@@ -0,0 +1,58 @@
+// Copyright 2026 SiFive, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+/**
+ * @brief Test and benchmark for GEMM: C = alpha * A * B + beta * C.
+ *
+ * This test uses a table-driven approach where test configurations are defined
```
Aaron-Hutchinson, line 9: This comment feels a bit outdated given the last several revisions and I think it can be simplified greatly. The approach is no longer table-driven (the harness doesn't do any lookups), and it doesn't seem necessary to describe all the parameters a test should specify since I think it's pretty clear from the struct definition below.

Maybe something like:
```
A test suite should be provided in a separate file, which
defines an array of `gemm_f32rc_f32rc_f32rc_t` test configurations,
along with a `skl_test_suite_t` object. See `test/README.md` for details.

For a particular test configuration, the derived matrix dimensions are:
  - A is m x k with row stride rsa and column stride csa
  - B is k x n with row stride rsb and column stride csb
  - C is m x n with row stride rsc and column stride csc
```

test/gemm/rvv/skl_gemm_f32_f32_f32_zve32f_x390.c
ericlove: Is your objection to the "Set default strides:" part?  I'm not sure how this comment differs from your statement.

ericlove (commented):
@Aaron-Hutchinson thanks for the additional comments. At this point, it would be more effective for you to PR any desired changes, as I don't have enough bandwidth.

test/gemm/xsfmm/skl_gemm_a1b01_f32c_f32_f32_xsfmm32a32f.c
Aaron-Hutchinson: Implemented through #63.

test/skl-test-driver.h
```diff
@@ -0,0 +1,352 @@
...
+ * Contains all state for a single test execution, including driver
+ * configuration and harness-specific data. The driver initializes this
+ * structure and passes it to all harness callback functions.
+ */
+struct skl_test_t {
+  const skl_test_suite_t *suite; /**< Test suite (set by driver before init) */
+  const skl_test_steps_t
+      *steps;    /**< Pointers to test functions (set by test) */
+  void *harness; /**< Harness-specific data (set by harness in init) */
+  int log_level; /**< Logging verbosity level (set by driver before init) */
```
peichialin: It seems we cannot set different log levels for the tests belong to the same suite from the cmake entry. Will we consider to move it to skl_test_suite_t?

test/CMakeLists.txt
```diff
@@ -0,0 +1,52 @@
...
+    skl-test-driver
+    PRIVATE ${SKL_COMPILE_OPTIONS} ${SKL_TEST_COMPILE_OPTIONS}
+)
+
+function(skl_add_test TEST_NAME HARNESS_SRC TEST_SRC OPTS)
+    add_executable(test-${TEST_NAME} ${HARNESS_SRC} ${TEST_SRC})
+    target_link_libraries(
+        test-${TEST_NAME}
+        PRIVATE ${SKL_LIBRARY} ${SKL_TEST_LIBS} skl-test-driver
+    )
```
peichialin: Why don't we link `skl` here through `target_link_libraries` here?

peichialin:
The overall interfaces look good to me. Few questions were posted above.

test/CMakeLists.txt
Aaron-Hutchinson: I had the same confusion initially -- it's linked using the `${SKL_LIBRARY}` variable, which should maybe be removed throughout the repository.

test/skl-test-driver.h
Aaron-Hutchinson: The log level can be set globally from CMake using `-DSKL_TEST_LOG_LEVEL`, and specific test suites could be ran with `ctest -R [regex]`. Does this workflow meet your needs?

test/gemm/gemm_f32rc_f32rc_f32rc.c
```diff
@@ -166,78 +74,84 @@ int check_error(void) {
...
+#define INFO(fmt, ...) SKL_TEST_LOG(t, SKL_TEST_LOG_INFO, fmt, __VA_ARGS__)
+
+  INFO("M: %zd, N: %zd, K: %zd\n", h->m, h->n, h->k);
+  INFO("CSA: %zd, RSB: %zd, RSC: %zd\n", h->csa, h->rsb, h->rsc);
+  INFO("Alpha: %f, Beta: %f\n", h->alpha, h->beta);
+  INFO("%s", "\n");
+  INFO("Warmup: %s\n", t->steps->warmup ? "yes" : "no");
+  INFO("Cycles: %zd\n", t->counters.cycles);
+  INFO("Instructions: %zd\n", t->counters.instret);
+  INFO("MACs/Cycle: %f\n", mpc);
```
ebavier, line 138: We might have trouble formatting floating-point values in bare-metal environments (the reason we currently have `print_float` in `skl-test.h`).

test/gemm/rvv/skl_gemm_f32_f32_f32_zve32f_x390.c
ebavier: `t->status` here is set by the above `SKL_TEST_REQUIRE`, but I assume we'd want to check `t->status` and `return 1` prior to the GEMM call if necessary.

test/skl-test-driver.h
```diff
@@ -0,0 +1,352 @@
...
+/**
+ * @brief Test context structure.
+ *
+ * Contains all state for a single test execution, including driver
+ * configuration and harness-specific data. The driver initializes this
+ * structure and passes it to all harness callback functions.
+ */
+struct skl_test_t {
+  const skl_test_suite_t *suite; /**< Test suite (set by driver before init) */
+  const skl_test_steps_t
```
ebavier: I don't understand why this member needs to be present, versus just mandating `skl_test_steps_t` be the first member of the harness struct and just executing steps directly from each individual harness.
Doing this would remove this struct member, the `*suite` member (which I think only exists to get access to the `steps` function), as well as the boilerplate `steps` function that clients currently need to define.

test/skl-test-driver.c
```diff
@@ -0,0 +1,180 @@
...
+  free(ptr);
+}
+
+static size_t fprintf_prefix(FILE *stream, skl_test_t *t) {
+  if (t == NULL) {
+    return 0;
+  }
+  size_t id =
+      ((char *)t->harness - (char *)t->suite->tests) / t->suite->test_size;
+  return fprintf(stream, "[%zd]: ", id);
```
ebavier: 1. If we frequently need a numerical index for tests, let's just calculate it up front by the driver.  I don't like the idea of tests reaching up into the suite too often.
2. Printing a numerical id for context is great, but this appears to be the only context given.  It would be nice if some additional context were given, like a summary of the current test parameters.  Maybe this could be enforced by requiring a callback function to print test details, or by convention by having a `init` log the current parameters?

Numerical IDs are less helpful when debugging if any macros or runtime generation are used to populate a test suite, since it may be difficult to trace an ID back to a concrete set of runtime parameters.

test/skl-test-driver.c
```diff
@@ -0,0 +1,180 @@
...
+}
+
+void skl_test_driver_error(skl_test_t *t, const char *fmt, ...) {
+  va_list args;
+  va_start(args, fmt);
+  (void)fprintf_prefix(stderr, t);
+  (void)vfprintf(stderr, fmt, args);
```
ebavier: Do we want to fold this function into below `skl_test_driver_log` and set up `SKL_TEST_LOG` to send `SKL_TEST_LOG_ERROR` messages to `stderr`?

test/skl-test-driver.h
```diff
@@ -0,0 +1,352 @@
...
+ *
+ * If COND is false, logs an error message with the test name, condition,
+ * file, and line number, then sets T->status to SKL_TEST_FAIL. Does not
+ * return early, allowing multiple requirements to be checked.
+ */
+#define SKL_TEST_REQUIRE(T, COND)                                              \
+  do {                                                                         \
+    if (!(COND)) {                                                             \
+      skl_test_driver_error((T), "%s: %s\n", (T)->suite->name, #COND);         \
+      skl_test_driver_error((T), "    %s:%d\n", __FILE__, __LINE__);           \
```
ebavier: This `__FILE__` and `__LINE__` are a nice touch, but how they are used in `skl-test-driver.c`'s `skl_test_driver_run_suite` for the `TRY` macro, they only ever point into that function, which is not super helpful.  I would expect instead to be pointed at the line where test parameters are declared/defined.

test/skl-test-driver.h
```diff
@@ -0,0 +1,352 @@
...
+  int (*init)(
+      skl_test_t *t); /**< Optional initialization function, NULL to skip */
+  int (*warmup)(skl_test_t *t);  /**< Optional warmup function, NULL to skip */
+  int (*execute)(skl_test_t *t); /**< Required execution function */
+  int (*verify)(
+      skl_test_t *t); /**< Optional verification function, NULL to skip */
+  int (*report)(
+      skl_test_t *t); /**< Optional reporting function, NULL to skip */
+  int (*cleanup)(skl_test_t *t); /**< Optional cleanup function, NULL to skip */
+} skl_test_steps_t;
```
ebavier, line 153: The expectations of these callback functions are poorly documented.  E.g. they all return `int`, but it's only clear from reading the driver code that those return values are treated in a non-zero-is-failure sense.   I assume there are expectations on the particular actions of each function that could probably be documented.  E.g. do we expect the `verify` to be silent and test errors to be reported only in `report`? Or is `report` strictly for benchmark reporting?

test/skl-test-driver.c
```diff
@@ -0,0 +1,180 @@
...
+    if (t.steps->report != NULL) {
+      int report_status = t.steps->report(&t);
+      if (report_status != 0) {
+        skl_test_driver_error(&t, "Report function failed\n");
+        t.status = SKL_TEST_FAIL;
+      }
+    }
+
+    // Check final status and update failure count
+    if (t.status != SKL_TEST_PASS) {
```
ebavier: The `t.status` here refers to the status reported by `verify` by the driver's `SKL_TEST_REQUIRE`.  The `cleanup` and `report` functions cannot be allowed to use `SKL_TEST_REQUIRE`, or the driver needs to cache the status earlier.

test/skl-test-driver.c
```diff
@@ -0,0 +1,180 @@
...
+    // Verify test results, if verification is enabled
+    TRY(steps->verify);
+    if (t.steps->verify) {
+      SKL_TEST_LOG(&t, SKL_TEST_LOG_INFO, "Verification passed\n");
+    }
+
+  cleanup:
+    // Always run cleanup if provided
+    if (t.steps->cleanup != NULL) {
+      int cleanup_status = t.steps->cleanup(&t);
```
ebavier: I think it would be reasonable for a `report` to want to maybe examine the values found in test results.  Would it make sense to run `cleanup` as the last step?

test/skl-test-driver.h
```diff
@@ -0,0 +1,352 @@
...
+ *
+ * Allocates a buffer and initializes it according to the parameters in BUF.
+ * The initialization mode determines how the buffer is filled:
+ * - SKL_TEST_RANDOM: Fill with random values in [min, max]
+ * - SKL_TEST_STATIC: Fill from static_data array (cycling if needed)
+ * - SKL_TEST_SEQ: Fill with sequential values from min to max
+ *
+ * Logs allocation at log level SKL_TEST_LOG_DEBUG.
+ */
+#define SKL_TEST_BUF_CREATE(T, TYPE, BUF)                                      \
```
ebavier, line 394: `T` here seems to be used exclusively for logging, but has more potential.  Like tracking allocated pointers and free'ing them automatically later.

test/skl-test-driver.c
```diff
@@ -0,0 +1,180 @@
...
+
+    // Run the test, including optional cache warmup
+    TRY(steps->warmup);
+    skl_test_driver_update_counters(&t.counters);
+    TRY(steps->execute);
+    skl_test_driver_update_counters(&t.counters);
+
+    // Verify test results, if verification is enabled
+    TRY(steps->verify);
+    if (t.steps->verify) {
```
ebavier:
```suggestion
    if (t.steps->verify != NULL) {
```

test/skl-test-driver.c
```diff
@@ -0,0 +1,180 @@
...
+      SKL_TEST_LOG(&t, SKL_TEST_LOG_INFO, "Verification passed\n");
+    }
+
+  cleanup:
+    // Always run cleanup if provided
+    if (t.steps->cleanup != NULL) {
+      int cleanup_status = t.steps->cleanup(&t);
+      if (cleanup_status != 0) {
```
ebavier:
```suggestion
      if (cleanup_status != SKL_TEST_PASS) {
```

ebavier (commented):
A few comments/questions here, and also inline:

1. It seems there's no way to disable function tests during a benchmark run?
2. It's not clear to me in what circumstances, and how, someone might want to change the allocation mode for input data.  E.g. I might want to use `SKL_TEST_SEQ` in benchmark tests in order to limit simulation time taken up by random float generation; would that need to be declared at the test cases level?

test/skl-test-driver.c
```diff
@@ -0,0 +1,180 @@
...
+  (void)vfprintf(stdout, fmt, args);
+  va_end(args);
+}
+
+static void log_separator(skl_test_t *t, char sep) {
+  enum { SEP_LEN = 10 };
+  char buf[SEP_LEN];
+  memset(buf, sep, sizeof(buf));
+  buf[sizeof(buf) - 1] = '\n';
+  SKL_TEST_LOG(t, SKL_TEST_LOG_INFO, "%s", buf);
```
ebavier:
```suggestion
  buf[sizeof(buf) - 1] = '\0';
  SKL_TEST_LOG(t, SKL_TEST_LOG_INFO, "%s\n", buf);
```
Otherwise the formatted print may print past the end of `buf`

test/skl-test-driver.h
ericlove: This is my fault. I had an irrational aversion to the implicit pass-by-layout convention, but I will concede that you and @Aaron-Hutchinson have the better idea here.

Aaron-Hutchinson: > just mandating `skl_test_steps_t` be the first member of the harness struct 

Yeah, we discussed this approach in one of the resolved threads above. Admittedly I don't love mandating this, but it likely does result in a much cleaner infrastructure. I'll try this out today to see how much it simplifies things.

Aaron-Hutchinson:
> It seems there's no way to disable function tests during a benchmark run?

The driver has no concept of test vs benchmark, and I don't think it should since the use of possibly-NULL function pointers blurs the line between them. I would say this should be left up to the test file to handle, where test authors could organize their test file as something like:
```
gemm_f32rc_f32rc_f32rc_t tests[] = {
#if SKL_ENABLE_BENCHMARKS
  {BENCHMARK, .m = 32, .n = 32, .k = 32},
#endif

#if SKL_ENABLE_TESTS
  {TEST, .m = 32, .n = 32, .k = 32},
#endif
};
```
But I'm open to other suggestions. What do you think?

> It's not clear to me in what circumstances, and how, someone might want to change the allocation mode for input data. E.g. I might want to use `SKL_TEST_SEQ` in benchmark tests in order to limit simulation time taken up by random float generation; would that need to be declared at the test cases level?

If you just mean the initialization mode, this should be configurable within individual tests by setting the `SKL_TEST_BUFFER` parameters appropriately. e.g. you should be able to do:
```
.a = {.min = 0, .max = 16, .mode = SKL_TEST_SEQ}
```
If you mean the memory allocation function used to allocate the underlying buffer, this is set globally by the driver with `skl_test_driver_alloc` and tests currently can't override it.

test/skl-test-driver.h
Aaron-Hutchinson: That sounds very useful but I'm having trouble seeing the full picture. Can you give some more details on how this might work?

ebavier: I'm imagining the `skl_test_t` could have an array of pointers, modestly sized.  `SKL_TEST_BUF_CREATE` would add the allocated pointer to the array.  All pointers are then free'd by `skl_test_driver_run_suite` before/after `cleanup`.  If we provide an analogous allocation for non-initialized memory, I'd imagine most test harnesses would have a no-op `clean`.

test/skl-test-driver.c
Aaron-Hutchinson: I like that idea and made the change. We could also consider removing `SKL_TEST_LOG` altogether by moving the conditional wrapper into `skl_test_driver_log`.

test/skl-test-driver.h
Aaron-Hutchinson: I made this change, which simplified some of the requirements of the tests.

I did remove the `*suite` member, but added `id` and `suite_name` members in `skl_test_t` since these fields are referred to in error reporting and I agree that harnesses shouldn't be reaching up into the actual suite struct.

Aaron-Hutchinson: There's likely still some documentation that needs updating in regard to this change. I haven't touched `README.md` at all yet, which is also missing some information on my previous changes from last week.

ericlove:
@Aaron-Hutchinson we should add some features to the driver that let `main()` programs easily filter tests, and one common predicate would be "verify?" to distinguish tests from benchmarks.

test/skl-test-driver.c
Aaron-Hutchinson: > If we frequently need a numerical index for tests, let's just calculate it up front by the driver. I don't like the idea of tests reaching up into the suite too often.

Agreed, I added an `id` member to `skl_test_t` and removed the suite pointer altogether.

> Printing a numerical id for context is great, but this appears to be the only context given. It would be nice if some additional context were given, like a summary of the current test parameters. Maybe this could be enforced by requiring a callback function to print test details, or by convention by having a `init` log the current parameters?

I think by convention through the `report` function is the current method.

Aaron-Hutchinson:
> @Aaron-Hutchinson we should add some features to the driver that let main() programs easily filter tests, and one common predicate would be "verify?" to distinguish tests from benchmarks.

I don't think this should be the driver's responsibility; it's given a test suite and it should just run it. Also `main` is implemented by the test files now.

If the toggle between test or benchmark mode is to be supplied via compile definition anyway (how else?), what's wrong with the suggestion I gave above?

test/skl-test-driver.c
Aaron-Hutchinson: Upon looking into a good solution to this, it feels like there's some redundancy and conflicting responsibility between the `t.status` field and the return value of the callback functions. I think we should use one of these but not both. Since the callbacks (like `report`) may want to know the pass/fail status, I propose removing the return value of the callbacks and relying entirely on `t.status`.

test/README.md
```diff
@@ -0,0 +1,195 @@
...
+- `verify`: Verification function called after the test execution.
+  It is responsible for checking the results of the test against a reference implementation.
+- `report`: Reporting function called after the test execution.
+  It is responsible for printing any test-specific results or statistics.
+- `cleanup`: Cleanup function called at the end of each test case.
+  It is responsible for freeing any test-specific resources.
+
+Most of these functions should be reusable across different kernels in the same family, so only the `execute` function needs to be customized for each kernel variant to call the appropriate kernel function and pass it the correct parameters.
+
+All callback functions take a `skl_test_t` pointer as their only argument, which is a generic test context structure defined in [skl-test.h](skl-test.h).
```
france5289: Should this be `skl-test-driver.h` instead? It looks like `skl_test_t` is defined there rather than in `skl-test.h`.

```suggestion
All callback functions take a `skl_test_t` pointer as their only argument, which is a generic test context structure defined in [skl-test-driver.h](skl-test-driver.h).
```

test/skl-test-driver.c
```diff
@@ -0,0 +1,174 @@
...
+
+static void log_separator(skl_test_t *t, char sep) {
+  enum { SEP_LEN = 10 };
+  char buf[SEP_LEN];
+  memset(buf, sep, sizeof(buf));
+  buf[sizeof(buf) - 1] = '\0';
+  SKL_TEST_LOG(t, SKL_TEST_LOG_INFO, "%s\n", buf);
+}
+
+int skl_test_driver_run_suite(skl_test_suite_t *suite) {
```
france5289, line 104: The implementation is solid, but the readability is a bit low due to the complexity (long loop body). Breaking this into helpers would help a lot. Here are some refactoring suggestions that might be worth considering:

## Summary of changes

| | Before | After |
|---|---|---|
| `TRY` definition site | Inside `for` loop body | File scope, before `run_one_test` |
| `TRY` macro scope leak | Yes (no `#undef`) | No (`#undef` after function) |
| `TRY` captures local var | Yes (captures `t` by name) | No (takes `T` as explicit parameter) |
| Per-test lifecycle | Inline in the `for` loop (~55 lines) | Extracted into `run_one_test` |
| `run_suite` loop body | ~55 lines | 1 line (`failures += run_one_test(...)`) |
| cleanup/report pattern | Hand-written `if` blocks, reason implicit | Same pattern, reason made explicit by comment |

```c
// At file scope, before run_suite:
#define TRY(T, STEP)                                \
  do {                                              \
    if ((STEP) != NULL) {                           \
      SKL_TEST_REQUIRE((T), (STEP)((T)) == 0);     \
      if ((T)->status != SKL_TEST_PASS)             \
        goto cleanup;                               \
    }                                               \
  } while (0)

static int run_one_test(skl_test_suite_t *suite, size_t i, int log_level) {
  skl_test_t t = {
      .harness    = (char *)suite->tests + i * suite->test_size,
      .suite_name = suite->name,
      .id         = i,
      .log_level  = log_level,
      .counters   = {.cycles = 0, .instret = 0},
      .status     = SKL_TEST_PASS,
  };

  // The first member of the harness struct must be skl_test_steps_t.
  skl_test_steps_t *steps = (skl_test_steps_t *)t.harness;

  log_separator(&t, '=');

  TRY(&t, steps->init);

  TRY(&t, steps->warmup);
  skl_test_driver_update_counters(&t.counters);
  TRY(&t, steps->execute);
  skl_test_driver_update_counters(&t.counters);

  TRY(&t, steps->verify);
  if (steps->verify != NULL)
    SKL_TEST_LOG(&t, SKL_TEST_LOG_INFO, "Verification passed\n");

cleanup:
  // cleanup and report always run, even when a previous step failed.
  if (steps->cleanup != NULL) {
    if (steps->cleanup(&t) != SKL_TEST_PASS) {
      skl_test_driver_log(&t, stderr, "Cleanup function failed\n");
      t.status = SKL_TEST_FAIL;
    }
  }
  if (steps->report != NULL) {
    if (steps->report(&t) != SKL_TEST_PASS) {
      skl_test_driver_log(&t, stderr, "Report function failed\n");
      t.status = SKL_TEST_FAIL;
    }
  }

  if (t.status != SKL_TEST_PASS)
    SKL_TEST_LOG(&t, SKL_TEST_LOG_ERROR, "Test failed\n");

  log_separator(&t, '=');
  SKL_TEST_LOG(&t, SKL_TEST_LOG_INFO, "\n");

  return (t.status != SKL_TEST_PASS) ? 1 : 0;
}
#undef TRY

int skl_test_driver_run_suite(skl_test_suite_t *suite) {
  int failures = 0;
  int log_level = SKL_TEST_LOG_LEVEL;

  if (suite->num_tests > 0 && suite->name != NULL &&
      log_level >= SKL_TEST_LOG_INFO) {
    skl_test_driver_log(NULL, stdout, "Running test suite %s with %zd tests\n",
                        suite->name, suite->num_tests);
  }

  for (size_t i = 0; i < suite->num_tests; ++i)
    failures += run_one_test(suite, i, log_level);

  if (log_level >= SKL_TEST_LOG_INFO) {
    skl_test_driver_log(NULL, stdout,
                        "Test suite %s completed %zd tests with %d failures\n",
                        suite->name, suite->num_tests, failures);
  }

  return failures;
}
```

test/skl-test-driver.c
Aaron-Hutchinson: I've just pushed a major change regarding this that replaces the callback return values and the `t.status` value with a struct of status values for each callback. This way, the status of each step can be reported and tracked separately. This also had the side effect of simplifying the driver loop considerably.

test/skl-test-driver.c
Aaron-Hutchinson: @france5289 Thanks for the very detailed suggestion! In trying to resolve the "status" issue discussed in one of the above threads, I made a significant revision of the driver loop body that I think helped the readability significantly. Some of your suggestions were orthogonal to the changes I made, but in particular if you think it's still worth factoring out the loop body into a dedicated function I'm not opposed to this.

test/skl-test-driver.h
Aaron-Hutchinson: I've expanded this documentation significantly. I didn't include any statements regarding the verbosity of `verify` to leave this up to individual harnesses, but let me know if you think we should make a definitive policy here.

test/README.md
```diff
@@ -0,0 +1,209 @@
...
+The list of test cases may include both correctness tests and performance benchmarks, distinguished by the callback functions they specify.
+Any callback except `execute` may be provided as `NULL` to skip the execution of such step.
+
+The test suite must also define a `skl_test_suite_t` structure that describes the test suite to the driver, and includes the array of test configurations:
+```c
+skl_test_suite_t suite = {.name = "skl_gemm_a1b01_f32c_f32_f32_xsfmm32a32f",
+                          .num_tests = sizeof(tests) / sizeof(tests[0]),
+                          .test_size = sizeof(gemm_f32rc_f32rc_f32rc_t),
+                          .tests = tests};
+};
```
france5289: Looks like there are some redundant `}` and `;` here.

test/skl-test-driver.c
france5289: Thanks @Aaron-Hutchinson. This refactor greatly enhances the readability of the loop. I agree that moving it to a dedicated function is no longer necessary.

test/skl-test-driver.h
```diff
@@ -0,0 +1,479 @@
...
+  do {                                                                         \
+    (BUF)->data = (TYPE *)skl_test_driver_alloc((BUF)->region,                 \
```
france5289, line 396: It's probably safer to check the return value of `skl_test_driver_alloc` for nullity before using it. Since it relies on `aligned_alloc`, adding this check will help prevent potential null pointer dereferences.

test/skl-test-driver.h
```diff
@@ -0,0 +1,479 @@
...
+ *
+ * Describes a buffer of test data with initialization parameters.
+ * Test harnesses use SKL_TEST_BUFFER() to define buffer members in their
+ * test configuration structures. These can then be allocated and initialized by
+ * the driver using SKL_TEST_BUF_CREATE.
+ */
+#define SKL_TEST_BUFFER(TYPE)                                                  \
+  struct {                                                                     \
+    TYPE *data;                                                                \
+    size_t len;                                                                \
```
france5289, line 360: Since `len` is widely used as a divisor, we should add a check to ensure it's never zero. This will help prevent potential division-by-zero errors if a developer accidentally misconfigures it.

test/gemm/gemm_f32rc_f32rc_f32rc.c
france5289: While the current macro names are unlikely to cause conflicts, it's generally a good practice to `#undef` them after use. This prevents macro pollution and avoids potential naming collisions that could be very difficult to debug in the future.
So I suggest we should `#undef` any macros in our code.

test/skl-test-driver.h
Aaron-Hutchinson: That's a good point. I attempted to add the following after the allocation:
```
if (!(BUF)->data) {
  SKL_TEST_LOG((T), SKL_TEST_LOG_ERROR, "failed to allocate %s\n", #BUF);
  break;
}
```
But the code then exceeds our cognitive complexity threshold (understandably so!). This macro is already difficult to read and understand, and should probably be decomposed or refactored somehow...

test/skl-test-driver.h
Aaron-Hutchinson: I think this depends on the use case. GEMM (or rather, BLAS) treats the K=0 case specially and it's convenient to allow a zero `len` in those tests.

Aaron-Hutchinson: As a suggestion, this requirement could be asserted in `init` using something like `SKL_TEST_REQUIRE(t, init_status, buf->len != 0);`.

Aaron-Hutchinson (approved):

ericlove (commented):
Can't officially approve this as author, but I give @Aaron-Hutchinson permission to merge.

PR #57, opened 2026-03-11
[GEMM] Retune general FP32 GEMM for L1 cache misses

Aaron-Hutchinson:
This performs a major revision of the `skl_gemm_4xm4x1_f32_f32_f32_zve32f_x390` kernel to improve performance on larger input sizes that result in a large number of L1 cache misses.

Summary of changes:
- Unroll the K loop by a factor of 4, resulting in a `4 x m4 x 4` tile size. The kernel name was updated accordingly.
- Load the A matrix values one (unrolled) K iteration in advance.
- Utilize the `c.ntl.p1` instruction of the [Zihintntl extension](https://github.com/riscv/riscv-isa-manual/blob/main/src/zihintntl.adoc) on each `flw` of the K loop.
  - This prevents the A matrix values from persisting in the L1 cache, which ensures the `flw` is a non-blocking load. This allows the latency of each block of `flw`s to overlap. With the other changes above, there is sufficient load-to-use time to hide this latency.
- I found that this revised kernel also has great performance when using CLP allocated matrices. As such, this PR deletes the existing RVV `_clp` kernel.


ericlove (commented):
Thanks for this important optimization.  I'm thinking about how to deal with the `zihintntl` ISA requirement: in theory this should be part of the kernel name if it's required, but to ease integration, I think we should just get around this problem with a macro that optionally uses them:
```c
#if defined(__riscv_zihintntl)
#define NTL_P1 "ntl.p1 \n\t"
#else
#define NTL_P1
#endif
```

Then we can just make the issue go away without violating any of the conventions. What do you think?

Aaron-Hutchinson:
@ericlove I had that exact same idea so I'm glad you brought it up. I'll make the change.

Aaron-Hutchinson:
Unfortunately clang-format thought our idea was awful. How strict are we about `clang-format off` on small segments of code?

ericlove:
Go ahead and turn it off for any inline asm segments especially 👍

CMakeLists.txt
```diff
@@ -194,6 +196,10 @@ if(RISCV_ZVE32F)
   target_sources(${SKL_LIBRARY} PRIVATE ${SKL_ZVE32F_SRCS})
 endif()
 
+if(RISCV_ZVE32F AND RISCV_ZIHINTNTL)
+  target_sources(${SKL_LIBRARY} PRIVATE ${SKL_ZVE32F_ZIHINTNTL_SRCS})
+endif()
+
```
hdelassus: Shouldn't we compile (and test) this regardless of `RISCV_ZIHINTNTL`?

Aaron-Hutchinson: It was a requirement before switching to the `NTL_P1` macro as discussed above, but I missed removing this condition after adding the macro. Thanks!

ericlove (approved):
LGTM!

PR #58, opened 2026-03-11
[GEMM] Add BF16 -> F32 Zvfbfwma kernel

Aaron-Hutchinson:
This adds a general GEMM kernel for BF16 input types with F32 accumulator and output type, using the Zvfbfwma extension, in intrinsics. In particular, the `vfwmaccbf16` intrinsic is used for accumulation, and a `vfmv` with 0 is used for initialization instead of a widening multiply.

This kernel uses no micro-architectural suffix in its name since it's intended to be a general purpose kernel for the Zvfbfwma extension. For this initial PR the kernel uses a standard `4xm4x1` tile size, which has shown to give sufficiently high performance.

Some CMake logic is added to support Zvfbfwma, and a test is also added.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -64,6 +64,10 @@ set(SKL_ZVFH_SRCS
   src/softmax/softmax_f16_zvfh.c
 )
 
+set(SKL_ZVFBFWMA_SRCS
+  src/gemm/rvv/gemm_bf16_bf16_f32_zvfbfwma.c
+)
+
 set(SKL_ZVE32F_SRCS
   src/depthwise_conv2d/depthwise_conv2d_f32_f32_f32_zve32f.c
   src/exp/exp_bf16_zve32f.c
@@ -187,6 +191,10 @@ endif()
 
 if(RISCV_ZVFH)
   target_sources(${SKL_LIBRARY} PRIVATE ${SKL_ZVFH_SRCS})
+endif()
+
+if(RISCV_ZVFBFWMA)
+  target_sources(${SKL_LIBRARY} PRIVATE ${SKL_ZVFBFWMA_SRCS})
 endif()
 
 if(RISCV_ZVE32F)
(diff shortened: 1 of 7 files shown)


hdelassus (commented):
I believe the CI toolchain supports Zvfbfwma. Can you try enabling it in the CMake toolchain file?

Aaron-Hutchinson:
> I believe the CI toolchain supports Zvfbfwma. Can you try enabling it in the CMake toolchain file?

Forgot about that, thanks. I added it and it seems like CI is compiling the kernel and running the test.

hdelassus (dismissed):
LGTM!

hdelassus (dismissed):

CMakeLists.txt
```diff
@@ -379,6 +387,10 @@ if (SKL_BUILD_TESTS OR SKL_BUILD_BENCHMARKS)
       "-DM=32;-DN=32;-DK=32;-DALPHA=1.f;-DBETA=0.f")
     skl_add_test(logistic_f16 logistic/logistic_f16.c "-DRUN_ALL")
   endif()
+  if(RISCV_ZVFBFWMA)
+    skl_add_test(skl_gemm_bf16_bf16_f32_zvfbfwma_wrapper gemm/gemm_bf16rc_bf16rc_f32rc.c
+      "-DRSA=128;-DCSA=1;-DM=4;-DN=16;-DK=128;-DALPHA=1.f;-DBETA=0.f")
```
ericlove: Why is this test setting explicit strides for the A matrix but no others?

ericlove (dismissed):

CMakeLists.txt
Aaron-Hutchinson: Oops. Removed.

Aaron-Hutchinson:
Sorry to dismiss the reviews (again). I did a readability pass over the kernel code since this is partially intended as an RVV reference kernel and I think it's been improved significantly.

hdelassus (approved):

PR #59, opened 2026-03-19
[Pack][Transpose] Add f32 Xsfmm pack-A kernel

myeh01:
This PR adds a packing kernel `skl_pack_tex1_e32_e32pc_xsfmmbase` that can be used to repack a row-major A matrix into the correct format for the f32 Xsfmm GEMM kernels. The more general underlying routine `skl_pack_e32_e32rcpc_xsfmmbase` repacks a row major matrix into the packed format described in [packed_gemm.md](https://github.com/sifive/skl/blob/main/src/gemm/packed-gemm.md) with column major tiles (`rs0 = 1`), as long as the tile size `m0 x n0` does not exceed `TE x TE`. This routine can be used to implement matrix transposition. Whereas the previous implementation of matrix transposition transposed one tile at a time, this routine interleaves the load and storing of tiles, improving performance.


ericlove (commented):
Thank you for this important performance improvement in an essential kernel for SFMM engine use.  I'm glad to see that this optimization is available through the existing transpose API, as well as in a more flexible interface that allows packing into matrix panels as well for applications that need it.

Before reviewing the implementation itself, I'd like to discuss how this fits into SKL generally and how it's presented:

- "repack a row-major A matrix into the correct format for the f32 Xsfmm GEMM kernels." Let's avoid implying that there is a specific correct format for SFMM LHS matrices other than column-major. This is a property of a particular application, rather than of the ISA.
- I understand that this kernel produces an output with a packed layout, but the primary function seems to be transposition.  Given that fact and that it's in the `transpose/` directory (and transpose file), it might be clarifying to call it `transpose()`, but with the `tex1_` and `_e32cp_` infixes.
- That said, applications using this kernel will most likely pack the B matrix as well, using a similar panel size albeit without transposition.  (Admitting that _the_ application is IREE, we can say this with certainty.)  Would we plan to add a symmetric `pack_1xte_` kernel for the sake of completeness?
- If the answer to the last question is "yes" then we'll have a potentially-odd arrangement because the `_1xte` kernel only needs vector loads/stores, and doesn't use the matrix engine. Would we still include this in the `_xsfmmbase` collection? (Maybe yes because it would still need that ISA to read TE, rather than receive it as an argument.)
- Similarly, an application that multiplies rectangular panels of A by panels of B will produce a TExTE blocked C matrix output.  Will we add a corresponding unpacking function?
- Would we ever generalize the SFMM version to support arbitrary multiples of TE?
- Whatever we choose, we should align it with the other ISAs that have associated packed layouts: Xsfvqdotq and Xsfvqmaccqoq.  The latter is still missing from SKL, while the former provides `skl_pack_b_i8_xsfvqdotq` from within the `src/gemm/` directory next to its packed GEMM implementation.  It has no reference to any generic pack reference function.

Sorry for thinking aloud here, and don't feel the need to respond to each of these questions. I just want to write them down here so we can try to think though the structure of our growing collection of pack/transpose kernels. (Having done so, I'm more convinced the right answer is to do what you did here.)

ericlove:
I have opened a PR (https://github.com/sifive/skl/pull/59) with a proposed general approach to the incorporation of pack/unpack kernels in SKL.  Please help review whether it makes sense, and then we can consider what implications, if any, it has for this work.

I think overall I have reasoned my way into supporting what @myeh01 basically did here: put the block dimensions in the kernel name, and use e.g. `_tex1` to imply transposition _inside_ the blocks.  I think the general API is good for both transposition and "paneling", as proposed in the `pack` version of this PR's code.

Remaining open question raised by this PR: should we just duplicate the code between `transpose` and `pack` if they need to be in two separate directories? Or would we consider eliminating "transposition" as a kernel family entirely if it can be treated as a special case of packing? I don't entirely like that, as it will make some users wonder at first where and whether transposition is supported.  But we can also consider a third name such as "reorder" or a corny but effective portmanteau like "transpack".

src/transpose/xsfmm/transpose_e32_xsfmmbase.h
```diff
@@ -14,6 +14,37 @@
...
+ * );
+ * ```
+ *
+ * This function is intended to provide packing functionality for all 32-bit
+ * datatypes by way of type-punning through the input/output array pointers and
+ * of the padding value.
+ */
+void skl_pack_tex1_e32_e32pc_xsfmmbase(size_t m, size_t n, const uint32_t *a,
+                                       size_t rsa, uint32_t *a_pack,
+                                       size_t rsa1, uint32_t padding_value);
```
ericlove: Are we aware of any use case where a padding value other than `0` is required?

myeh01: Not currently, but the proposed IREE pack/unpack interface has the padding value as a parameter.

myeh01: @JerryShih mentioned the zero-point might be used as padding in quantized matmul

ericlove:
@peichialin @Harlen126 I propose to get this merged, and then @Harlen126 can move these kernels to `pack/` as part of his execution plan presented in the slide deck.  Do we agree?

src/transpose/xsfmm/transpose_e32_xsfmmbase.h
peichialin: If the padding value is for quantized matmul with asymmetric quantization, then I expect that `a` should be in smaller SEW like e8, shouldn't it?

peichialin:
I agree with merging it by keeping the implementation under src/transpose now after the review is done.
We can move them to the new folder when combining transposition and pack.

src/transpose/xsfmm/transpose_e32_xsfmmbase.h
myeh01: @peichialin Right, then we would probably need to add another kernel for packing an `e8` matrix. But, we'd still like to know whether any kernels would use a nonzero padding value.

phoebesv:
My only concern is that the API interface and parameters differ from those of the generic pack API.
That said, as we discussed, one possible approach would be to add a wrapper to provide a unified interface on our side, which could help simplify integration across different implementations. This is based on our current understanding. If there are any design considerations, constraints, or trade-offs that we may not have considered, we’d be happy to discuss them further.

phoebesv:
Does the SFMM transpose pack implementation currently support other data types—such as FP16, BF16, or INT8? If not, are there any related PRs we could refer to, or are there any plans to add support for these types in the future?

myeh01:
> My only concern is that the API interface and parameters differ from those of the generic pack API. That said, as we discussed, one possible approach would be to add a wrapper to provide a unified interface on our side, which could help simplify integration across different implementations. This is based on our current understanding. If there are any design considerations, constraints, or trade-offs that we may not have considered, we’d be happy to discuss them further.

The Xsfmm pack kernels should be more specialized than the generic kernel, so they will not support the full API. The main reason we want to use the Xsfmm engine is for faster transposition. If the packing problem does not require transposition (`cs0 == 1`), then the RVV kernel should be used.

Currently, only an `ETE` x 1 packing kernel is exposed publicly for Xsfmm GEMM. Do you have other use cases where you would like to use Xsfmm for packing? We may be able to generalize the kernel for different `m0` and `n0`.

myeh01:
> Does the SFMM transpose pack implementation currently support other data types—such as FP16, BF16, or INT8? If not, are there any related PRs we could refer to, or are there any plans to add support for these types in the future?

After this one is merged, I will open PRs for the remaining widths.

src/transpose/xsfmm/transpose_e32_xsfmmbase.c
```diff
@@ -7,11 +7,344 @@
...
+        src + i1 * m0 * rs + (j1 + 1) * n0, rs, mt4, pad);
+    skl_pack_tile_store_pad_bottom_e32_e32_xsfmmbase(
+        tn1, mt4c, n0_right, m0_bottom, dst + i1 * rs1 + (j1 + 1) * cs1, cs0,
+        pad);
+  }
+  // NOLINTEND(readability-suspicious-call-argument)
+}
+
+SKL_XSFMM_NEW
+SKL_FUNC void skl_pack_e32_e32rcptex1c_xsfmmbase(size_t m, size_t n,
```
adam1029, line 336: Could we update the parameters to match the general pack API format as shown below:  
 
```c
void skl_pack_e32_e32rcptex1c_xsfmmbase(
    size_t m,
    size_t n,
    const uint32_t *src,
    size_t rs,
    size_t cs,
    size_t m0,
    size_t n0,
    uint32_t *dst,
    size_t rs0,
    size_t cs0,
    size_t rs1,
    size_t cs1,
    uint32_t pad) {
    ...
    ...
    skl_pack_padding_optional_e32_e32rcptextec_xsfmmbase(
      m, n, src, rs, dst, cs1, rs1, ete * cs1, false, true, pad);
    ...
    ...
   }
```

adam1029 (commented):
-

test/pack/xsfmm/skl_pack_e32_e32rcptex1c_xsfmmbase.c
```diff
@@ -0,0 +1,132 @@
...
+  {TEST, .m = 191, .n = 127},
+  {TEST, .m = 191, .n = 191},
+  {TEST, .m = 191, .n = 255},
+
+  {TEST, .m = 192, .n = 256, .rs1 = (size_t)(256 * 65), .cs1 = 65},
+  {TEST, .m = 192, .n = 256, .rs1 = 64, .cs1 = 192},
+  {TEST, .m = 192, .n = 256, .rs1 = 65, .cs1 = (size_t)(3 * 65)},
+  {TEST, .m = 191, .n = 255, .rs1 = (size_t)(256 * 65), .cs1 = 65},
+  {TEST, .m = 191, .n = 255, .rs1 = 64, .cs1 = 192},
+  {TEST, .m = 191, .n = 255, .rs1 = 65, .cs1 = (size_t)(3 * 65)},
```
Harlen126: It seems these tests only work for `ete == 64` cases.
Changing `xsfmm64t` to `xsfmm32t` or `xsfmm128t` will lead to failures.

src/transpose/xsfmm/transpose_e32_xsfmmbase.c
Harlen126: Calling 
```
skl_pack_e32_e32rcptex1c_xsfmmbase(m, n, src, rs, dst, rs1, cs1, pad);
```
is equivalent to calling
```
skl_pack_e32rc_e32rcprc_ref(m, n, rs, 1 /*cs*/, 64 /*m0*/, 1 /*n0*/, dst, 1 /*rs0*/, 0 /*cs0, don't care*/, rs1, cs1, pad);
```
Some of the parameters are fixed, so they are not included as input parameters.

I think you are requesting an API such as `skl_pack_e32rc_e32rcprc_xsfmmbase()`. However, the specialized "copy" implementation will be identical to `skl_pack_e32rc_e32rcprc_zve32x()`, and will not require `xsfmmbase`. As a result, we will not provide a general pack API for the "xsfmmbase" pack.

phoebesv:
> Currently, only an `ETE` x 1 packing kernel is exposed publicly for Xsfmm GEMM. Do you have other use cases where you would like to use Xsfmm for packing? We may be able to generalize the kernel for different `m0` and `n0`.

Currently, llama.cpp's use case requires transposing matrices from (m×k) to (k×m) layout after packing, which might result in an ETEx1 format, but I'm not sure whether vLLM or other projects have the same requirement.

src/transpose/xsfmm/transpose_e32_xsfmmbase.c
ericlove: Right, as @Harlen126 said, it has a different API than the generic version because some of the parameters are fixed to concrete values.

It is part of SKL's general design philosophy to remove parameters that have been specialized, forcing users to call the correct kernel via their own dispatch.  This helps avoid accidental misuse, and avoids `assert`s within the library.  More importantly, it helps preserve the separability of individual kernels, which can be copied out of the codebase easily without any dependencies.

I realize this adds complexity to the use of SKL kernels, but that is expected. SKL is not intended to be "easy" to use -- rather, it is only intended to provide access to the highest-performance implementations that exploit specific ISA or HW features.  The choice of which kernel to use and how is the responsibility of higher-level libraries, and can be considered part of their "value-add".

ericlove:
> Currently, llama.cpp's use case requires transposing matrices from (m×k) to (k×m) layout after packing, which might result in an ETEx1 format, but I'm not sure whether vLLM or other projects have the same requirement.

What does "packing" mean in the llama.cpp context then?

phoebesv:
> What does "packing" mean in the llama.cpp context then?

The packing here means we would like to optimize cache performance in later work by pre-packing matrices into blocked format for better cache locality. Currently, llama.cpp's XSFMM implementation does not use packing, which we want to add as a future optimization.

test/pack/xsfmm/skl_pack_e32_e32rcptex1c_xsfmmbase.c
myeh01: Thanks for catching this! I have modified these tests so the strides are set in `main` based on `ete`. Please let me know what you think.

myeh01:
Ready for review.

test/pack/xsfmm/skl_pack_e32_e32rcptex1c_xsfmmbase.c
Harlen126: I'm not sure if it's a good idea to bind the test cases to the stride initialization in `main`.
Since there are only `Xsfmm16t`, `Xsfmm32t`, `Xsfmm64t`, and `Xsfmm128t` defined currently, we could use the preprocessor to initialize the strides instead:

```
#if defined(__riscv_xsfmm16t)
#define ETE (16)
#elif defined(__riscv_xsfmm32t)
#define ETE (32)
#elif defined(__riscv_xsfmm64t)
#define ETE (64)
#elif defined(__riscv_xsfmm128t)
#define ETE (128)
#else
#error ETE undefined
#endif
```

ericlove: Perhaps we can use `__riscv_min_xsfmm_te` instead? Based on some experiments, it seems that will actually take the largest defined TE `march` parameter:
```
$ riscv64-unknown-elf-clang -xc -dM -E /dev/null -march=rv64gcv_xsfmmbase_xsfmm32t_xsfmm64t_xsfmm16t | grep xsf
#define __riscv_min_xsfmm_te 64
#define __riscv_xsfmm16t 6000
#define __riscv_xsfmm32t 6000
#define __riscv_xsfmm64t 6000
#define __riscv_xsfmmbase 6000
```

myeh01: We've been setting "`ete`" by
```
__asm__ volatile("sf.vsettnt %0, x0, e8, w1" : "=r"(ete) : : "vtype", "vl");
```
Couldn't `ete` be strictly less than `TE` on machines with small `VLEN`?

ericlove: Your point is that we don't handle the SEW=64 case correctly? I would agree with that but we'll have to deal with that later.

test/pack/xsfmm/skl_pack_e32_e32rcptex1c_xsfmmbase.c
```diff
@@ -0,0 +1,153 @@
...
+pack_e32rc_e32rcprc_t tests[] = {
+#ifdef SKL_ENABLE_BENCHMARKS
+  // Benchmark tests
+  {BENCH, .m = 256, .n = 256},
+#endif
+
+#ifdef SKL_ENABLE_TESTS
+  // Verification tests
+  /* Test different {r,c}s1, set in main */
+  {TEST, .m = 192, .n = 256},
```
ericlove: I think we need to apply the lessons from this conversation https://github.com/sifive/skl/pull/59#discussion_r3503696004 or did we decide not to?

test/pack/xsfmm/skl_pack_e32_e32rcptex1c_xsfmmbase.c
```diff
@@ -0,0 +1,153 @@
...
+    tests[i].n0 = 1;
+
+    tests[i].rs0 = 1;
+    tests[i].cs0 = tests[i].m0;
+    tests[i].cs1 = tests[i].cs1 ? tests[i].cs1 : tests[i].m0 * tests[i].n0;
+    tests[i].rs1 = tests[i].rs1 ? tests[i].rs1 : tests[i].n * tests[i].cs1;
+  }
+
+#ifdef SKL_ENABLE_TESTS
+  tests[0].cs1 = ete + 1;
```
ericlove: If we really are going to override tests from the static array, we should set those to zero or something obvious to make it clear the intent is to replace them

test/pack/xsfmm/skl_pack_e32_e32rcptex1c_xsfmmbase.c
myeh01: I was looking at this equation from the spec:
```
sf.vsettn rd, rs1 // rd = tn/vl = min(rs1=ATN, LMUL*EVE, ETE)
```
My concern was there may be a case where `LMUL*EVE < ETE`, but maybe this is not possible? I'm trying to work through the math. We do have the constraint `TE <= VLEN / 4`.

myeh01: Ok, if I did this correctly, it looks like `LMUL*EVE >= ETE`, so we could probably use Eric's suggestion. I will write out the calculations later so people can check.

myeh01: Since `TE <= VLEN / 4`,
```
TEW < 64
ceil(ETE/EVE) = ceil(ETE * SEW / VLEN) = ceil(TE * SEW / VLEN) <= ceil(SEW / 4) = SEW / 4

TEW == 64
ceil(ETE/EVE) = ceil(ETE * SEW / VLEN) = ceil(TE * SEW / (2 * VLEN)) <= ceil(SEW / 8) = SEW / 8
```
Supported `SEW`, `TWIDEN`, `KMAX` values from the spec, `min(8/KMAX, 8/TWIDEN)`, and `SEW/4 or SEW/8`
```
TEW=8
SEW TWIDEN KMAX    min(8/KMAX, 8/TWIDEN)    SEW/4
 8   1      4       2                        2

TEW=16
SEW TWIDEN KMAX    min(8/KMAX, 8/TWIDEN)    SEW/4
 8   2      4       2                        2
16   1      2       4                        4

TEW=32
SEW TWIDEN KMAX    min(8/KMAX, 8/TWIDEN)    SEW/4
 8   4      4       2                        2
16   2      2       4                        4
32   1      1       8                        8

TEW=64
SEW TWIDEN KMAX    min(8/KMAX, 8/TWIDEN)    SEW/8
16   4      2       2                        2
32   2      1       4                        4
64   1      1       8                        8
```
So, `ceil(ETE/EVE) <= min(8/KMAX, 8/TWIDEN)` and therefore
```
LMUL := min(8/KMAX, 8/TWIDEN, ceil(ETE/EVE)) = ceil(ETE/EVE)
```
which means
```
LMUL*EVE = ceil(ETE/EVE) * EVE >= ETE
```

myeh01: Ok, tests have been updated.

Harlen126: The calculations look good to me.
Thanks.

Harlen126 (dismissed):
LGTM! My only minor suggestion is: can we use a macro like `#define ETE __riscv_min_xsfmm_te` for readability?

myeh01:
> LGTM! My only minor suggestion is: can we use a macro like `#define ETE __riscv_min_xsfmm_te` for readability?

How about `SKL_XSFMM_TE`? Let me know what you think: https://github.com/sifive/skl/pull/59/changes/7ef8ff507819539797ef7643553dd39267b52570.

Harlen126:
> How about `SKL_XSFMM_TE`? Let me know what you think: [7ef8ff5](https://github.com/sifive/skl/commit/7ef8ff507819539797ef7643553dd39267b52570).

Looks great! Thanks.

Harlen126 (dismissed):

myeh01:
Need a reapproval after rebasing.

Harlen126 (approved):

PR #60, opened 2026-03-19
Add Xsfvfexp32e+Zvfbfmin BF16 softmax to libskl and test harness

hdelassus:
This patch adds the already existing Xsfvfexp32e+Zvfbfmin BF16 softmax kernel to the build system, public header, and BF16 softmax test harness.


ericlove (approved):
Thanks for filling in this gap.

PR #62, opened 2026-03-21
[Infra] Clean up CMakeLists.txt source handling

Aaron-Hutchinson:
This PR attempts some cleanup of the top level CMakeLists.txt file's handling of the library source files, which has become quite verbose and clunky.

Major changes that try to address this:
- Replace the individual `RISCV_{extension}` variables with a single list `RISCV_EXTENSIONS` of extensions, populated in `get-riscv-extensions.cmake`. An extension can now be checked for with `if("extension" IN_LIST RISCV_EXTENSIONS)`.
- Rewrote the source handling in `CMakeLists.txt` to use a ~list function based approach so that the sources can be looped over instead of checked one-by-one~ function based approach. 
  - ~CMake doesn't support lists-of-lists or dictionaries, so this was admittedly somewhat hacky: I used a list where each "entry" consisted of a delimiter (newline), a source file, and a list of required RISC-V extensions. I'm open to better suggestions, but I think this is still a big improvement.~

In addition to reviewing the overall approach, each code owner should be check that their sources are present and have the right extensions attached.

---

In doing this revision I noticed `softmax_bf16_xsfvfexp32e_zvfbfmin.c` wasn't being compiled into the library; I assumed this was a mistake and added it to the source list. CC @ebavier.


ericlove (commented):
I like this cleanup a lot! Thanks for taking on this work.

One idea: why not define a function that takes a src file and an ISA as parameters and conditionally adds the source to libskl?

Aaron-Hutchinson:
I don't know if it's cleaner, but the logic is certainly easier to understand with a function. I replaced the loop with a function in the last commit. One minor drawback is that sources with no extension dependencies have to now pass an empty list `""`.

CMakeLists.txt
```diff
@@ -27,270 +27,106 @@ target_include_directories(${SKL_LIBRARY} PUBLIC
...
-  target_sources(${SKL_LIBRARY} PRIVATE ${SKL_XSFVFBFEXP16E_XSFVFBFA_SRCS})
-endif()
-
-if(RISCV_XSFVFBFEXP16E AND RISCV_ZVFBFMIN)
-  target_sources(${SKL_LIBRARY} PRIVATE ${SKL_XSFVFBFEXP16E_ZVFBFMIN_SRCS})
-endif()
+function(add_skl_src SRC REQUIRED_EXTENSIONS)
+  # Add the source file SRC to the SKL_LIBRARY if all RISC-V
+  # extensions in REQUIRED_EXTENSIONS are present in RISCV_EXTENSIONS.
+  foreach(EXT ${REQUIRED_EXTENSIONS})
```
ericlove, line 37: Another idea (not sure if this works!): what if instead of conditionally returning a value, this function side-effectfully called `target_sources(${SKL_LIBRARY} ...)` to just add sources?

ebavier:
> have to now pass an empty list `""`

CMake supports optional arguments, right?

CMakeLists.txt
Aaron-Hutchinson: I think that's what the code is currently doing. You might have to scroll down in the diff to see the entire function.

Aaron-Hutchinson:
> CMake supports optional arguments, right?

I thought about using them but wasn't sure if it would be making things too complicated. I'm open to trying it if we think it's a better approach.

Aaron-Hutchinson:
Also, what do others think about converting to lower casing for the RISC-V extension names (as they appear as C macros `__riscv_{ext}`)? I think they were only upper case due to formerly being CMake variables.

CMakeLists.txt
```diff
@@ -364,12 +200,12 @@ if (SKL_BUILD_TESTS OR SKL_BUILD_BENCHMARKS)
   enable_testing()
 
   # Examples
-  if (RISCV_ZVE32F)
+  if("ZVE32F" IN_LIST RISCV_EXTENSIONS)
```
myeh01: Would it also be good to do something similar for the tests in case they depend on multiple extensions?

Aaron-Hutchinson: Yes I agree that would be a huge improvement as well. However all of the `add_skl_test` lines here will eventually be removed as part of the refactor happening in #56, so we can address this in the revised `add_skl_test` function after that PR merges.

For `add_skl_example` we could do something similar, but I'd rather leave that to future work to keep this PR more focused.

ericlove:
> Also, what do others think about converting to lower casing for the RISC-V extension names (as they appear as C macros `__riscv_{ext}`)? I think they were only upper case due to formerly being CMake variables.

SGTM 👍

hdelassus (dismissed):

ericlove (dismissed):

hdelassus (dismissed):

ericlove (dismissed):

Aaron-Hutchinson:
I just rebased this to resolve conflicts, which involved a lot of retyping with the possibility of more typos. Please glance over it for any errors before reapproving.

ericlove (dismissed):

Aaron-Hutchinson:
Sorry to dismiss reviews yet again. I forgot to update the new `test/CMakeLists.txt` file introduced in #56 after rebasing, which I've now done. I used a similar approach in `add_skl_test` as `add_skl_src`, where the test is conditionally added when all required extensions are present.

hdelassus (approved):
Last changes LGTM

PR #63, opened 2026-03-23
[Test] Let tests set their own function pointers

Aaron-Hutchinson:
This is a preview of what #56 would look like to allow each test to specify its own collection of steps ran by the test driver.

The various function pointers in `skl_test_suite_t` were moved into a new struct `skl_test_steps_t` that each test has a pointer to. Each test in the array specifies a `skl_test_steps_t` object that the driver acquires by means of a new `steps` function defined in the test file.

This is slightly more verbose, but it allows very fine control on what every single test is doing. It also enables setting functions to `NULL` on a per-test basis, which simplifies some of the harness logic.


ericlove (approved):
I think this is a good proposal 👍

PR #64, opened 2026-03-24
[Packing] Name Scheme & Documentation for Pack/Unpack Kernel Family

ericlove:
Inspired by questions raised in @myeh01 's work (https://github.com/sifive/skl/pull/59), this moves the "pack & unpack" kernel description out of the [packed GEMM doc](https://github.com/sifive/skl/blob/main/src/gemm/packed-gemm.md) into a separate README in `skl/src/pack`, which it also creates.

## Changes
It makes the following proposals:

1. A standard naming scheme: `skl_pack_<m0xn0>_` for `m0` x `n0` blocks.
2. Drop `_a_` / `_b_` / `_c_` matrix references in names.
3. Blocks are presumed to have row-major order internally, but this only shows up so far in Xsfvqmaccqoq (4x4 and 4x8).
4. Explicit differentiation between _mandatory_ (needed by ISA) and _optional_ (helps framework integration) packing.
5. List of optional packing kernels that will be provided (due to anticipated demand) and those that won't (because they can be handled without specific optimization).
6. A proposal -- yet unimplemented -- of a "generic RVV packing" kernel that should handle all "easy" cases (block sizes are large enough that you can vectorize without dimension-specific hacks).

## Review and Feedback
Please share your opinions of all the proposals outlined above (not feeling so sure about many of them!), and when we agree on the overall design, then help review the document itself.

## Status & Plans
Keeping this as a draft for now, as the changes it proposes are not reflected in the actual kernel source code.
I will wait for consensus on the plan before we start making changes to the code.

### DIFF
--- a/src/pack/README.md
+++ b/src/pack/README.md
@@ -71,7 +71,7 @@ It is assumed by default that the ordering of elements within a block is fixed t
 If instead the layout within a block is column-major, the inner-block-rder term is `c`, and if it may be either, `rc` is used.
 Similarly, the ordering of blocks relative to each other is assumed to be block-row-major, but if it is either block-column-major or configurable, a block-order term of `bc` or `brc` is added to the end of the layout term.
 
-When one of the dimensions is `1`, there is no difference in memory ordering between row- and column-major; however, the order is still said to be "column-major" when the second dimension is `1`, as this corresponds to the canonical layout of a column vector, and better indicates the transposition implied by the layout.
+When one of the dimensions is `1`, there is technically no difference in memory ordering between row- and column-major; however, the order is still said to be "column-major" when the second dimension is `1`, as this corresponds to the canonical layout of a column vector, and better communicates the transposition implied by the layout.
 
 Use of `1` for either block dimension effectively creates a block with one side of arbitrary size, referred to as a _panel_.
 
@@ -103,10 +103,10 @@ Memory layout: [0,4,8,12, 1,5,9,13, 2,6,10,14, 3,7,11,15, 16,20,24,28, 17,21,25,
                └─Block(0,0)┘ └─Block(0,1)┘ └─Block(0,2)┘ └─Block(0,3)┘ └─Block(1,0)─┘ └─Block(1,1)─┘
 
 Strides: rs0=1 (row stride within block), cs0=4 (column stride within block)
-         rs1=4 (row stride between blocks), cs1=16 (column stride between blocks)
+         rs1=16 (row stride between blocks), cs1=4 (column stride between blocks)
 ```
 
-### Example: 1x4 Block Layout (Block-Column-Major with Outer Transposition)
+### Example: 1x4bc Block Layout (Block-Column-Major with Outer Transposition)
 ```
 Original row-major matrix (4x8):        Packed layout [m0=1, n0=4, m1=4, n1=2]:
 
@@ -157,7 +157,7 @@ Memory layout: [0,4,1,5, 2,6,3,7, 8,12,9,13, 10,14,11,15]
                └Block(0,0)┘ └Block(0,1)┘ └─Block(1,0)─┘ └─Block(1,1)─┘
 
 Strides: rs0=1 (row stride within block), cs0=2 (column stride within block)
-         rs1=4 (row stride between blocks), cs1=4 (column stride between blocks)
+         rs1=8 (row stride between blocks), cs1=4 (column stride between blocks)
 ```
 Note: This layout does not occur in any of the current SKL GEMM kernels.
 It is for illustrative purposes only.
--- a/src/pack/README.md
+++ b/src/pack/README.md
@@ -99,8 +99,9 @@ Original row-major matrix (8x4):        Packed layout [m0=4, n0=1, m1=2, n1=4]:
                                         │ 28  │     │ 29  │     │ 30  │     │ 31  │
                                         └─────┘     └─────┘     └─────┘     └─────┘
 
-Memory layout: [0,4,8,12, 1,5,9,13, 2,6,10,14, 3,7,11,15, 16,20,24,28, 17,21,25,29, ...]
-               └─Block(0,0)┘ └─Block(0,1)┘ └─Block(0,2)┘ └─Block(0,3)┘ └─Block(1,0)─┘ └─Block(1,1)─┘
+Memory layout:
+[ 0, 4, 8,12,  1, 5, 9,13,  2, 6,10,14,  3, 7,11,15, 16,20,24,28, 17,21,25,29, ...]
+ └Block(0,0)┘ └Block(0,1)┘ └Block(0,2)┘ └Block(0,3)┘ └Block(1,0)┘ └Block(1,1)┘
 
 Strides: rs0=1 (row stride within block), cs0=4 (column stride within block)
          rs1=16 (row stride between blocks), cs1=4 (column stride between blocks)
@@ -130,8 +131,9 @@ Original row-major matrix (4x8):        Packed layout [m0=1, n0=4, m1=4, n1=2]:
                                         │24 25 26 27│ │28 29 30 31│
                                         └───────────┘ └───────────┘
 
-Memory layout: [0,1,2,3, 8,9,10,11, 16,17,18,19, 24,25,26,27, 4,5,6,7, 12,13,14,15, ...]
-               └Block(0,0)┘ └Block(1,0)─┘ └─Block(2,0)─┘ └─Block(3,0)─┘ └Block(0,1)┘ └─Block(1,1)─┘
+Memory layout:
+[ 0, 1, 2, 3,  8, 9,10,11, 16,17,18,19, 24,25,26,27,  4, 5, 6, 7, 12,13,14,15, ...]
+ └Block(0,0)┘ └Block(1,0)┘ └Block(2,0)┘ └Block(3,0)┘ └Block(0,1)┘ └Block(1,1)┘
 
 Strides: rs0=4 (row stride within block), cs0=1 (column stride within block)
          rs1=4 (row stride between blocks), cs1=16 (column stride between blocks)
@@ -153,8 +155,9 @@ Original row-major matrix (4x4):        Packed layout [m0=2, n0=2, m1=2, n1=2]:
                                         │ 9 13  │   │11 15  │
                                         └───────┘   └───────┘
 
-Memory layout: [0,4,1,5, 2,6,3,7, 8,12,9,13, 10,14,11,15]
-               └Block(0,0)┘ └Block(0,1)┘ └─Block(1,0)─┘ └─Block(1,1)─┘
+Memory layout:
+[ 0, 4, 1, 5,  2, 6, 3, 7,  8,12, 9,13, 10,14,11,15 ]
+ └Block(0,0)┘ └Block(0,1)┘ └Block(1,0)┘ └Block(1,1)┘
 
 Strides: rs0=1 (row stride within block), cs0=2 (column stride within block)
          rs1=8 (row stride between blocks), cs1=4 (column stride between blocks)
--- a/src/pack/README.md
+++ b/src/pack/README.md
@@ -14,7 +14,7 @@ See the [list of packing kernels](#list-of-packing-kernels) for details.
 
 These functions specialize the general form:
 ```c
-void skl_pack_<m0xn0>_<datatype>_<isa>(
+void skl_pack_rcbrc_<datatype>_<isa>(
   size_t m,           // Num. rows in input matrix
   size_t n,           // Num. columns in input matrix
   const <type>* src,  // Input matrix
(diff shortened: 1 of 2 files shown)


src/pack/README.md
```diff
@@ -0,0 +1,204 @@
...
+
+Not all extensions require packing for all matrices, but kernels for such cases may nevertheless be provided, as some applications may require similar layouts across all matrices, or seek locality benefits from arranging data in the shape of the underlying multiplication operation, or one of its dimensions.
+The documentation for each kernel indicates whether it is required or optional.
+See the [list of packing kernels](#list-of-packing-kernels) for details.
+
+## General API for Packing & Unpacking Kernels
+
+These functions specialize the general form:
+```c
+void skl_pack_<m0xn0>_<datatype>_<isa>(
```
myeh01: Should we remove the tile size from the name since `m0` and `n0` are arguments?
```suggestion
void skl_pack_<datatype>_<isa>(
```

src/pack/README.md
```diff
@@ -0,0 +1,204 @@
...
+  size_t rs,          // Row stride of input matrix
+  size_t cs,          // Column stride of input matrix
+  size_t m0,          // Num. rows in a block of the input matrix
+  size_t n0,          // Num. columns in a block of the input matrix
+  <type>* dst,        // Output packed matrix [m1 x n1]
+  size_t rs0,         // Row stride within a block of the output matrix
+  size_t cs0,         // Column stride within a block of the output matrix
+  size_t rs1,         // Row stride between blocks of the output matrix
+  size_t cs1          // Column stride between blocks of the output matrix
+) {
```
myeh01, line 30: Some applications (e.g. IREE) include an additional argument that specifies the padding value. Not sure if we want to include that in our API as well.

src/pack/README.md
```diff
@@ -0,0 +1,204 @@
...
+The general form is `skl_pack_<m0>x<n0>_<datatype>_<isa>`, where `<m0>` and `<n0>` are the block dimensions, `<datatype>` is the datatype of the matrix, and `<isa>` is the ISA extension required by the kernel.
+
+> **Note**: As in other kernel families, the ISA suffix indicates the minimum requirement to execute the packing kernel itself (normally `zve32x`), but its primary intended use may be for a particular matrix extension. This is indicated in the documentation for each kernel (packing and GEMM).
+>
+> The `skl_pack_tex1_e8_xsfmm` function is an exception, as it uses the matrix engine to accelerate transposition within blocks.
+
+If it is ever ambiguous whether the layout of elements within a block is row- or column-major, the layout term of the name will end in `c` in the case of column-major.
+However, this is omitted when one of the dimensions is 1, as the two layouts are the same.
+The ordering of blocks relative to each other is never reflected in the name, as this is always set by configurable block strides (there should be no performance advantage to fixing these at compile time).
```
myeh01: For the B matrix in the Xsfvqdotq kernels, we do want to fix the block column stride so that the `4 x 1` blocks are in block-row-major order. Should we consider reflecting the ordering of blocks relative to each other in the name?

src/pack/README.md
```diff
@@ -0,0 +1,204 @@
...
+
+The name of the packing kernel indicates the block size and ISA extension required by the kernel, and whether it is required or optional.
+The general form is `skl_pack_<m0>x<n0>_<datatype>_<isa>`, where `<m0>` and `<n0>` are the block dimensions, `<datatype>` is the datatype of the matrix, and `<isa>` is the ISA extension required by the kernel.
+
+> **Note**: As in other kernel families, the ISA suffix indicates the minimum requirement to execute the packing kernel itself (normally `zve32x`), but its primary intended use may be for a particular matrix extension. This is indicated in the documentation for each kernel (packing and GEMM).
+>
+> The `skl_pack_tex1_e8_xsfmm` function is an exception, as it uses the matrix engine to accelerate transposition within blocks.
+
+If it is ever ambiguous whether the layout of elements within a block is row- or column-major, the layout term of the name will end in `c` in the case of column-major.
+However, this is omitted when one of the dimensions is 1, as the two layouts are the same.
```
myeh01: Very nit: for blocks with one dimension equal to 1, I think we should not make an exception and still allow distinguishing between row and column major even though in memory they may be stored the same. One example I had in mind was the 4x1 blocks of the B matrix in the Xsfvqdotq kernels, which I think should be thought of as column major.

src/pack/README.md
```diff
@@ -0,0 +1,204 @@
...
+The ordering of blocks relative to each other is never reflected in the name, as this is always set by configurable block strides (there should be no performance advantage to fixing these at compile time).
+
+The choice of block dimensions and block ordering determines the overall data layout and can achieve different effects:
+
+- A `4x1` block implies transposition when the input matrix is row-major, since elements from different rows are grouped together within each block. The blocks themselves can be stored in either block-row- or block-column-major order. Block-column-major ordering is equivalent to full matrix transposition, with the `4` dimension only affecting padding.
```
myeh01: Very nit: this is related to the above comment about blocks with one dimension equal to 1. The `4x1` block could be viewed as row major (i.e. not transposed), a kind of degenerate case of a `4xn` block.

src/pack/README.md
```diff
@@ -0,0 +1,204 @@
...
+                                        ┌───────┐   ┌───────┐
+                                        │ 8 12  │   │10 14  │
+                                        │ 9 13  │   │11 15  │
+                                        └───────┘   └───────┘
+
+Memory layout: [0,4,1,5, 2,6,3,7, 8,12,9,13, 10,14,11,15]
+               └Block(0,0)┘ └Block(0,1)┘ └─Block(1,0)─┘ └─Block(1,1)─┘
+
+Strides: rs0=1 (row stride within block), cs0=2 (column stride within block)
+         rs1=4 (row stride between blocks), cs1=4 (column stride between blocks)
```
myeh01:
```suggestion
         rs1=8 (row stride between blocks), cs1=4 (column stride between blocks)
```

src/pack/README.md
```diff
@@ -0,0 +1,204 @@
...
+
+Strides: rs0=4 (row stride within block), cs0=1 (column stride within block)
+         rs1=4 (row stride between blocks), cs1=16 (column stride between blocks)
+```
+
+### Example: 2x2c Block Layout (Block-Row-Major with Column-Major Within Blocks)
+```
+Original row-major matrix (4x4):        Packed layout [m0=2, n0=2, m1=2, n1=2]:
+
+ 0  1  2  3                             Block(0,0)  Block(0,1)
```
myeh01, line 146: Should the blocks in the diagram not be transposed? Since blocks have both row and column stride parameters, there's no "default" layout.

myeh01:
Thanks for writing this up, @ericlove! I think I mostly agree with this proposal, but left a few comments about specifying the strides.

src/pack/README.md
ericlove: I guess this example is confusing because it presents the generic version, but the idea is that most implementations would fix `m0` and `n0` to specific values (e.g. `4x1` for the VQDOT pack-B kernel) and therefore move them from the argument list to the name.

src/pack/README.md
ericlove: As we discussed in our meeting with @peichialin just now, this is an important and open question.  I think the answer is probably yes, though many kernels will then fix the padding value to the block size.

src/pack/README.md
ericlove: I have fixed this instead by replacing it with `rcbrc` indicating this generic implementation's support for row- and column-major layouts within and between blocks.

src/pack/README.md
ericlove: Yes, this comment revealed a fundamental flaw in the current proposal (and my understanding of the mechanics of the packing kernel implementations).  I have made some serious changes to the default assumptions, where row-major is assumed as the default, and fixed.  Only explicit `c` or `rc` will now be used to indicate column-major or support for either. Let me know what you think!

src/pack/README.md
ericlove: I think you've convinced me that this is a good idea.  Should be reflected in recent changes.

src/pack/README.md
ericlove: Fixed this myself, along with some other dimension errors in the illustrations. Thanks for the catch, and please help review the changes!

ericlove:
After this is merged, I think we may want to revise the naming convention for packed GEMM kernels themselves.  In the pack_ing_ kernels, we indicate an output tensor's layout with `m0xn0<rc>b<rc>`, which is separate from the (single) type term (`_f32` etc.).

My idea is that we would would combine these two schemes: in the GEMM kernels, we would attach a specific layout suffix to each packed matrix type term, just after the `p`.  So for example `skl_gemm_a1b01_i8_i8pc_i32_xsfvqdotq` would become `skl_gemm_a1b01_i8p4x1c_i32_xsfvqdotq`.  And then we might want to further combine the two by renaming the packing kernel to `skl_pack_i84x1c_i8_zve32x`. These would then match with each other nicely.

For VQMACC, we'd have `skl_gemm_i8p8x4_i8p4x8_i32p4x4_xsfvqmaccqoq`, and for IREE-GEMM-SFMM, we'd have `skl_gemm_f32p1xtec_f32ptex1bc_f32ptexte_xsfmm32a32f` etc.

Thoughts? @myeh01 @hdelassus

src/pack/README.md
ericlove: (Specifically, I did not understand that the difference between transposition into panels and transposition _tout court_ manifests as the difference between unit-stride loads followed by strided-segmented stores for the former and a strided-segmented load by unit-stride stores in the latter.)

src/pack/README.md
```diff
@@ -0,0 +1,204 @@
...
+> **Note**: As in other kernel families, the ISA suffix indicates the minimum requirement to execute the packing kernel itself (normally `zve32x`), but its primary intended use may be for a particular matrix extension. This is indicated in the documentation for each kernel (packing and GEMM).
+>
+> The `skl_pack_tex1c_e8_xsfmm` function is an exception, as it uses the matrix engine to accelerate transposition within blocks.
+
+#### Layout Term
+
+The layout term may further be decomposed as `<m0>x<n0>[<inner-block-order>][<block-order>]`, where `m0` and `n0` are the block dimensions (rows and columns per block, respectively).
+
+It is assumed by default that the ordering of elements within a block is fixed to be row-major.
+If instead the layout within a block is column-major, the inner-block-rder term is `c`, and if it may be either, `rc` is used.
```
hdelassus:
```suggestion
If instead the layout within a block is column-major, the inner-block-order term is `c`, and if it may be either, `rc` is used.
```

src/pack/README.md
```diff
@@ -0,0 +1,204 @@
...
+(For Xsfmm, it is never required to pack any matrix; only [transposition](../transpose/README.md) is required if that A matrix is in row-major format, or B is in column-major format.)
+- `skl_pack_tex1c_e8_xsfmm`: Transpose (TE x K) panels using the matrix engine. Used on a row-major A matrix or column-major B matrix.
+
+### Packing Kernels That Do Not Exist
+These names refer to kernels that users might expect to exist, but do not, because they can be implemented with an appropriate block size in the generic RVV packing kernel at reasonable performance.
+In general, so long as the block size is large enough in at least one dimension (as is the case in paneling), a generic packing kernel with a large enough vector length can achieve reasonable performance.
+
+- ~~`skl_pack_1xte_e8_zve32x`~~: Pack (K x TE) panels. Used on a column-major A matrix or row-major B matrix.
+- ~~`skl_pack_texte_e32_zve32x`~~: Pack (TE x TE) blocks without transposition. Used on a C matrix.
+- ~~`skl_pack_1xvlm8_e8_zve32x`~~: Pack (1 x VLMAX*8) column panels.
```
hdelassus, line 197: The strikethrough + monospace combo makes it particularly difficult to read when Markdown is rendered, at least for me. Perhaps we'll want to remove the strikethrough?

src/pack/README.md
```diff
@@ -0,0 +1,204 @@
...
+These kernels are required for use of the `sf.vqmacc.4x8x4` instruction.
+Because the of the small block size, significant specialization is required to achieve reasonable performance.
+- `skl_pack_4x4_e32_zve32x`: Used for the C matrix.
+- `skl_unpack_4x4_e32_zve32x`: Inverse of `skl_pack_4x4_e32_zve32x`.
+- `skl_pack_4x8_e8_zve32x`: Used for the A matrix.
+- `skl_pack_8x4_e8_zve32x`: Used for the B matrix.
+
+### Optional Packing Kernels
+None of these are strictly required for correct operation of the corresponding GEMM kernels, but they may be useful for performance or compatibility reasons.
+(For Xsfmm, it is never required to pack any matrix; only [transposition](../transpose/README.md) is required if that A matrix is in row-major format, or B is in column-major format.)
```
hdelassus:
```suggestion
(For Xsfmm, it is never required to pack any matrix; only [transposition](../transpose/README.md) is required if the A matrix is in row-major format, or B is in column-major format.)
```
Not sure if that "that" means anything specific.

hdelassus (dismissed):
Thanks for taking the time to reformulate and add to this documentation. Looks good to me overall.

src/pack/README.md
```diff
@@ -0,0 +1,204 @@
...
+                                        Block(1,0)  Block(1,1)  Block(1,2)  Block(1,3)
+                                        ┌─────┐     ┌─────┐     ┌─────┐     ┌─────┐
+                                        │ 16  │     │ 17  │     │ 18  │     │ 19  │
+                                        │ 20  │     │ 21  │     │ 22  │     │ 23  │
+                                        │ 24  │     │ 25  │     │ 26  │     │ 27  │
+                                        │ 28  │     │ 29  │     │ 30  │     │ 31  │
+                                        └─────┘     └─────┘     └─────┘     └─────┘
+
+Memory layout: [0,4,8,12, 1,5,9,13, 2,6,10,14, 3,7,11,15, 16,20,24,28, 17,21,25,29, ...]
+               └─Block(0,0)┘ └─Block(0,1)┘ └─Block(0,2)┘ └─Block(0,3)┘ └─Block(1,0)─┘ └─Block(1,1)─┘
```
france5289: It looks like the alignment in this line is a bit off. For instance, the Block(0,0) label doesn't quite line up with its values 0,4,8,12. I noticed similar formatting shifts in a few other places throughout the document as well. It might need a little adjustment?

src/pack/README.md
```diff
@@ -0,0 +1,204 @@
...
+                                        Block(3,0)    Block(3,1)
+                                        ┌───────────┐ ┌───────────┐
+                                        │24 25 26 27│ │28 29 30 31│
+                                        └───────────┘ └───────────┘
+
+Memory layout: [0,1,2,3, 8,9,10,11, 16,17,18,19, 24,25,26,27, 4,5,6,7, 12,13,14,15, ...]
+               └Block(0,0)┘ └Block(1,0)─┘ └─Block(2,0)─┘ └─Block(3,0)─┘ └Block(0,1)┘ └─Block(1,1)─┘
+
+Strides: rs0=4 (row stride within block), cs0=1 (column stride within block)
+         rs1=4 (row stride between blocks), cs1=16 (column stride between blocks)
```
france5289, line 139: Regarding `rs1=4` and `cs1=16`, I'm trying to make sure I understand the stride calculation correctly. Based on the layout, I was expecting `rs1=8` and `cs1=4` instead. Could you help me clarify if I'm missing something or if these values might need a quick double-check?

ericlove: Because this is block-column-major, `rs1` is just the size of a block. The layout you described would be block-row-major.  The bocks on the "Packed Layout" side of the diagram are now shown in memory order, but the "Memory layout" below is correct.

I think the only questionable dimension is `rs0`, which really could be anything. Maybe it should be 1...

src/pack/README.md
ericlove: They don't have both row and stride parameters in this example, and there is a default block-row-major layout (otherwise this would be `Example: 2x2cbrc`).

src/pack/README.md
ericlove: They're supposed to be hard to read so that people don't look for them.

src/pack/README.md
ericlove: Good catch. Latest commit should fix.

src/pack/README.md
france5289: Thanks for the clarification, Eric! I completely missed that this section was referring to block-column-major. That clears up my confusion regarding the strides.
Also, I agree with you on `rs0`—setting it to 1 would definitely make more sense in this context.

myeh01:
> After this is merged, I think we may want to revise the naming convention for packed GEMM kernels themselves. In the pack_ing_ kernels, we indicate an output tensor's layout with `m0xn0<rc>b<rc>`, which is separate from the (single) type term (`_f32` etc.).
> 
> My idea is that we would would combine these two schemes: in the GEMM kernels, we would attach a specific layout suffix to each packed matrix type term, just after the `p`. So for example `skl_gemm_a1b01_i8_i8pc_i32_xsfvqdotq` would become `skl_gemm_a1b01_i8p4x1c_i32_xsfvqdotq`. And then we might want to further combine the two by renaming the packing kernel to `skl_pack_i84x1c_i8_zve32x`. These would then match with each other nicely.
> 
> For VQMACC, we'd have `skl_gemm_i8p8x4_i8p4x8_i32p4x4_xsfvqmaccqoq`, and for IREE-GEMM-SFMM, we'd have `skl_gemm_f32p1xtec_f32ptex1bc_f32ptexte_xsfmm32a32f` etc.
> 
> Thoughts? Michael Yeh (@myeh01) Hugues de Lassus (@hdelassus)

This sounds reasonable to me. Just wanted to clarify the examples (not sure if there were typos or if I did not understand correctly): the Xsfvqdotq kernel would be called `skl_gemm_a1b01_i8_i8p4x1c_i32_xsfvqdotq` and the packing kernel for B would be `skl_pack_i8_i8p4x1c_zve32x`; and the Xsfmm kernel would be `skl_gemm_f32ptex1c_f32p1xtebc_f32ptexte_xsfmm32a32f`.

ericlove:
> the Xsfvqdotq kernel would be called `skl_gemm_a1b01_i8_i8p4x1c_i32_xsfvqdotq` 

Yes that's right.

> and the packing kernel for B would be `skl_pack_i8_i8p4x1c_zve32x`;

Yeah I guess I had the argument order wrong in my head and didn't bother to check. Assume the suffixes are in the order of the arguments.

> and the Xsfmm kernel would be `skl_gemm_f32ptex1c_f32p1xtebc_f32ptexte_xsfmm32a32f`.

Yeah sorry I'm getting sloppy.  The A matrix should be (TE x 1) col-major and B should be (TE x 1) row major.

src/pack/README.md
myeh01: Sorry, what did you mean by fixing the padding value to the block size?

ericlove:
It sounds like this is mostly ready to be merged? I think we have agreed on the general API.  It would be good to unblock @myeh01 and @Harlen126 for work that builds on top of this.

hdelassus (approved):

Harlen126 (approved):

PR #66, opened 2026-04-01
[TEST] Port cvt test to new test structure

Harlen126:
This PR ports the cvt tests to the new testing framework introduced in https://github.com/sifive/skl/pull/56.

Unlike other SKL kernels, the conversion kernel supports various input/output data types. Therefore, the conversion test harness uses void pointers for input/output data instead of directly using the `SKL_TEST_BUFFER(TYPE)` feature provided by skl-test-driver.

Leveraging the feature introduced in #63 that allows each test to define its own "steps", various types of functions within a kernel can be tested in a single test suite.


test/cvt/cvt_ofp.c
```diff
@@ -0,0 +1,394 @@
...
+/**
+ * @brief Implementation of the cvt_ofp test harness.
+ *
+ * This file defines all harness functions _except_ `skl_test_execute`, which is
+ * defined in the test file (e.g. cvt_ofp_zvfofp8min.c).
+ */
+
+#include "cvt_ofp.h"
+#include "skl-test-driver.h"
+#include "skl-test.h"
```
Harlen126: The tests still include "skl-test.h" because the scalar reference implementations for conversion functions currently reside there. These will be moved to a separate header (e.g., "skl-ref.h") once "skl-test.h" is deprecated.

test/cvt/cvt_ofp.c
```diff
@@ -0,0 +1,394 @@
...
+
+static int init_input_bf16(skl_test_t *t, cvt_ofp_t *h) {
+  SKL_TEST_BUFFER(__bf16)
+  in = {.region = h->in_region,
+        .len = h->len,
+        .max = (__bf16)512.0f,
+        .min = (__bf16)-512.0f,
+        .mode = SKL_TEST_RANDOM};
+  SKL_TEST_BUF_CREATE(t, __bf16, &in);
+  h->in = in.data;
```
Harlen126: Using `SKL_TEST_BUFFER` as a local variable to access `SKL_TEST_BUF_CREATE`'s region allocation feature. This feels a bit awkward, but it avoids duplicating the allocation logic. Open to suggestions for a cleaner approach.

ericlove: @Aaron-Hutchinson should work with you to extend the `skl-test-driver.h` buffer creation macro to support these types. That would eliminate this layer of indirection.

test/cvt/cvt_ofp.h
```diff
@@ -0,0 +1,54 @@
...
+  float scale;
+  bool saturation;
+
+  // Buffer type for in,out
+  cvt_ofp_type_t in_type;
+  cvt_ofp_type_t out_type;
+
+  // Memory region for in,out
+  size_t in_region;
+  size_t out_region;
```
ericlove: I don't think your test harness should support regions directly, nor should any others inside the SKL repo since it has no notion of them.  That is a feature for individual tests in _other frameworks_ when they re-write the `test.c` file (which should be short so it can easily be replaced).

One way they might do that is to have their own `.init` function supplied via the `steps` field of this struct.

ericlove (commented):
@Harlen126 thank you so much for taking the time to write a test in this new framework! It is really very useful to see how people use it, so we can help refine it and the documentation to make it work better.

I will defer to @Aaron-Hutchinson who is now leading this effort to give detailed feedback and approval, but I have made a few comments below.

My high-level feedback is that it seems like the generic `cvt_ofp.c` harness ends up being quite complicated due to the proliferation of type-based dispatch logic.  I wonder if it's worth experimenting with an alternative that sacrifices this generality for shorter, type-specific `cvt_ofp_{type}.c` implementations.  Just a thought.

Aaron-Hutchinson:
> the generic `cvt_ofp.c` harness ends up being quite complicated due to the proliferation of type-based dispatch logic. I wonder if it's worth experimenting with an alternative that sacrifices this generality for shorter, type-specific `cvt_ofp_{type}.c` implementations.

I agree with this for a couple reasons: it makes shorter and simpler harnesses that are easier to understand, and more importantly I think we should strive to take a fairly uniform approach across the test directory to give some familiarity between the different suites. The intent in #56 was to have different harnesses for each datatype; we could debate that decision, but whatever approach is used I think should be applied consistently throughout the codebase.

test/cvt/cvt_ofp.c
Aaron-Hutchinson: Am I understanding correctly that this was not due to the `__bf16` type, but rather the generic typing of `void* in` in the harness struct? If that's the case then won't restricting the harness to a specific type solve this problem?

test/cvt/cvt_ofp.c
```diff
@@ -0,0 +1,394 @@
...
+  size_t count = len < 3 ? len : 3;
+  if (count > 0)
+    data[0] = (__bf16)nanf("");
+  if (count > 1)
+    data[1] = (__bf16)INFINITY;
+  if (count > 2)
+    data[2] = (__bf16)-INFINITY;
+}
+
+static int init_input_f32(skl_test_t *t, cvt_ofp_t *h) {
```
Aaron-Hutchinson: Can't `h` be derived from `t` as `t->harness`? What's the reason to accept it as a separate parameter?

test/cvt/cvt_ofp.c
Harlen126: Yes, the issue is generic typing, not __bf16 specifically.

The conversion kernel ([src/cvt/cvt_zvfofp8min.h](https://github.com/sifive/skl/blob/main/src/cvt/cvt_zvfofp8min.h)) contains multiple functions with different type combinations (F32→F8E4M3, BF16→F8E5M2, etc.). I wanted to test all these variants in a single test suite.

The current test framework doesn't support multiple harness struct types in a single test suite, so I used `void*` for input/output in the harness struct.

I could separate test suites by type, but that would require either:
1. Creating multiple test suites for a single kernel file
2. Splitting the kernel into separate type-specific files (e.g., cvt_f32_f8_zvfofp8min.h + cvt_bf16_f8_zvfofp8min.h)

Since the current convention seems to be one test suite per kernel, I'm leaning toward solution 2.

test/cvt/cvt_ofp.h
Harlen126: Thanks for the clear explanation. I'll remove the unnecessary region parameters from the harness.

test/cvt/cvt_ofp.c
Aaron-Hutchinson: Ah, I see. Yes I think approach 2 should be taken, for both test suite and source code consistency (at least, gemm uses kernel-specific headers and source code). But this is probably a question for @ericlove.

Aaron-Hutchinson (changes requested):
Sorry for the delay @Harlen126. I thought I had responded much earlier but my comments were still pending.

test/cvt/cvt_ofp.c
Harlen126: Good catch! You're right, `h` can be derived from `t->harness`. I'll fix this.
These helper functions were generated by Augment to reduce complexity, so I didn't review them carefully enough. Thanks for catching this!

src/cvt/cvt_bf16_f8_zvfofp8min_zvfbfmin.h
Aaron-Hutchinson: Should update the copyright year since this file is being changed.

src/cvt/README.md
```diff
@@ -1,4 +1,4 @@
-# OFP (*Open Compute Project* Floating Point) Conversion Kernels
+#OFP(*Open Compute Project *Floating Point) Conversion Kernels
```
Aaron-Hutchinson: The markdown doesn't render correctly with this change.

test/cvt/cvt_bf16_f8.c
```diff
@@ -0,0 +1,106 @@
...
+#include "skl-ref.h"
+#include <inttypes.h>
+#include <math.h>
+#include <stddef.h>
+#include <stdint.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+static void init_special_values_bf16(__bf16 *data, size_t len) {
```
Aaron-Hutchinson: This appears to only be called once and doesn't seem worth being defined as an independent function.

test/cvt/cvt_bf16_f8.c
```diff
@@ -0,0 +1,106 @@
...
+  h->out.len = h->len;
+  SKL_TEST_BUF_CREATE(t, __bf16, &h->in);
+  init_special_values_bf16(h->in.data, h->in.len);
+  if (h->out.len > 0) {
+    SKL_TEST_BUF_CREATE(t, uint8_t, &h->out);
+  }
+  if (h->steps.verify) {
+    h->ref = h->len > 0 ? malloc(h->len * sizeof(uint8_t)) : NULL;
+  }
+  t->status.init_status = SKL_TEST_PASS;
```
Aaron-Hutchinson: All of the `status` fields should be initialized by the driver with a `SKL_TEST_PASS` value, so the fields should only need to be updated when a failure occurs.

test/cvt/cvt_bf16_f8.c
```diff
@@ -0,0 +1,106 @@
...
+  uint8_t *out = h->out.data;
+  uint8_t *ref = h->ref;
+
+  for (size_t i = 0; i < h->len; ++i) {
+    ref[i] =
+        (h->out_type == F8E4M3)
+            ? skl_cvt_f32_f8e4m3((float)h->in.data[i] * h->scale, h->saturation)
+            : skl_cvt_f32_f8e5m2((float)h->in.data[i] * h->scale,
+                                 h->saturation);
+  }
```
Aaron-Hutchinson: Couple comments:
- `skl_cvt_f32_f8e*m*` are reference functions, right? If so aren't they missing a `_ref` suffix? Maybe this was missed in #68.
- Forgive me for not being familiar with the design of `cvt`. The signature of this reference function appears to be different from that of the SKL kernels, whereby the kernel operates elementwise whereas the ref operates on a single value. Could you give some insight on why this discrepancy exists? Shouldn't they align with each other?

test/cvt/cvt_bf16_f8.c
```diff
@@ -0,0 +1,106 @@
...
+  for (size_t i = 0; i < h->len; ++i) {
+    if (out[i] != ref[i]) {
+      SKL_TEST_LOG(t, SKL_TEST_LOG_ERROR,
+                   "result [%zu] (%d) != reference (%d)\n", i, out[i], ref[i]);
+      err_cnt += 1;
+    }
+  }
+
+  if (err_cnt) {
+    t->status.verify_status = SKL_TEST_FAIL;
```
Aaron-Hutchinson: I think you can omit the `else` case and move the `SKL_TEST_FAIL` up into the conditional inside the loop. Then there's no need for the `err_cnt` variable.

test/cvt/cvt_bf16_f8.c
```diff
@@ -0,0 +1,106 @@
...
+  t->status.report_status = SKL_TEST_PASS;
+}
+
+void cvt_bf16_f8_cleanup(skl_test_t *t) {
+  cvt_bf16_f8_t *h = (cvt_bf16_f8_t *)t->harness;
+
+  SKL_TEST_BUF_FREE(t, &h->in);
+  SKL_TEST_BUF_FREE(t, &h->out);
+  if (h->steps.verify)
+    free(h->ref);
```
Aaron-Hutchinson: Does this need a check for `len > 0`?

test/cvt/cvt_bf16_f8.h
Aaron-Hutchinson, line 1: Missing a `#pragma once`

test/cvt/cvt_bf16_f8_zvfofp8min_zvfbfmin.c
```diff
@@ -0,0 +1,175 @@
...
+#if !defined(__riscv_zvfofp8min) || !defined(__riscv_zvfbfmin)
+#error This file requires the Zvfofp8min and Zvfbfmin extensions
+#endif
+
+/**
+ * @brief Test cases for Conversion from BF16 to OFP8.
+ */
+
+#define TEST_BF16_F8E4M3                                                       \
+  CVT_BF16_F8_DEFAULTS, .out_type = F8E4M3, .saturation = false,               \
```
Aaron-Hutchinson: It might simplify this file a lot to just specify the `.saturation` value within each test, and check the value inside the execute functions. Then I think half of these macros and `execute`s can be removed. But I leave this up to your judgment.

test/cvt/cvt_f4_f8.c
```diff
@@ -0,0 +1,100 @@
...
+
+  uint8_t in_0 = in & 0xFU;
+  uint8_t in_1 = in >> 4U;
+  *out0 = skl_cvt_f32_f8e4m3(skl_cvt_f4e2m1_f32(in_0), false);
+  *out1 = skl_cvt_f32_f8e4m3(skl_cvt_f4e2m1_f32(in_1), false);
+}
+
+void cvt_f4_f8_init(skl_test_t *t) {
+  cvt_f4_f8_t *h = (cvt_f4_f8_t *)t->harness;
+  h->in.len = (h->len + 1) / 2, h->out.len = h->len;
```
Aaron-Hutchinson: What reason is there for using a comma operator here?

test/cvt/cvt_f4_f8.c
```diff
@@ -0,0 +1,100 @@
...
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+static void cvt_ofp4x2_f8e4m3(uint8_t in, uint8_t *out0, uint8_t *out1) {
+
+  uint8_t in_0 = in & 0xFU;
+  uint8_t in_1 = in >> 4U;
+  *out0 = skl_cvt_f32_f8e4m3(skl_cvt_f4e2m1_f32(in_0), false);
+  *out1 = skl_cvt_f32_f8e4m3(skl_cvt_f4e2m1_f32(in_1), false);
```
Aaron-Hutchinson: This function could use some explanatory comments for its purpose and what these magic values `0xFU` and `4U` are doing.

test/cvt/cvt_ofp.c
Aaron-Hutchinson, line 1: With the recent changes to use multiple harnesses and source files, is this file and its header now unnecessary?

test/CMakeLists.txt
```diff
@@ -69,3 +69,31 @@ skl_add_test(
   "xsfmm32a32f"
   ""
 )
+skl_add_test(
```
Aaron-Hutchinson, line 72: Same comment I posed to @peichialin in #78: we should sort these `skl_add_test` calls somehow.

Aaron-Hutchinson (changes requested):
Thanks for these huge improvements to the `cvt` tests @Harlen126! I reviewed the `cvt_bf16_f8` harness structure in detail, and skimmed the others. The harness code is looking much better with the recent changes!

As I'm not very familiar with the `cvt` kernels themselves, I'll have to leave it up to someone more qualified to review the cvt-specific test logic, as well as the proposed tests cases themselves.

test/cvt/cvt_f4_f8.c
Harlen126: That was originally a comma-separated declarator, but I forgot to change it when I refactored.
Thanks for the reminder.

test/cvt/cvt_bf16_f8_zvfofp8min_zvfbfmin.c
Harlen126: Good idea! I originally wanted to avoid dispatch overhead for benchmarking, but I think the impact is negligible. Let me merge them together! Thanks for the suggestion!

test/cvt/cvt_bf16_f8.c
Harlen126: These functions were originally written as utility functions for correctness testing. I didn't create scalar versions of the elementwise kernels because there's no real-world use case for OFP conversion without vector support—and as a result, there are no standard scalar conversion functions provided either.

Since these are now "ref" functions semantically, I can provide elementwise versions and rename them with the `_ref` suffix.

test/cvt/cvt_bf16_f8_zvfofp8min_zvfbfmin.c
Harlen126: Done~

test/cvt/cvt_bf16_f8.c
Harlen126: Add reference functions in https://github.com/sifive/skl/pull/66/changes/c98ef5094128bdcaf9cf0fd2ba38167279237a41.

test/cvt/cvt_f4_f8.c
Harlen126: Added comment in https://github.com/sifive/skl/pull/66/changes/c98ef5094128bdcaf9cf0fd2ba38167279237a41#diff-1df3c7473df07599767800edf239025900cdd0b446a854744f48878b5d8277d4R50.
Thanks.

Harlen126:
@Aaron-Hutchinson Thanks for your careful review!
I've addressed the points you mentioned.
Please let me know if it looks good!

ref/cvt/cvt_ofp4.h
```diff
@@ -0,0 +1,32 @@
...
+#if defined(__cplusplus)
+extern "C" {
+#endif
+
+/** @brief Convert a single OFP4 E2M1 value to IEEE FP32.
+ *
+ * @param in - Input value. The lowest 4 bits contain the OFP4 E2M1 value.
+ * @return The converted IEEE FP32 value.
+ */
+float skl_cvt_f4e2m1_f32(uint8_t in);
```
Aaron-Hutchinson, line 18: This should also have a `_ref` suffix right? This isn't necessarily relevant to this PR's goals though, so I'm fine with doing a more comprehensive update for this in a separate PR.

ref/cvt/cvt_ofp8.h
```diff
@@ -4,44 +4,157 @@
...
- * as an 8-bit unsigned integer.
- * @param is_sat - Whether to saturate the output to the maximum representable
- * value of the same sign when the input is out of range.
+ * @param in - Input IEEE FP32 value.
+ * @param is_sat - If true, infinite results are clamped to the maximum finite
+ * value of the same sign; if false, infinity is preserved.
+ * @return The converted OFP8 E5M2 value, represented as an 8-bit unsigned
+ * integer.
  */
 uint8_t skl_cvt_f32_f8e5m2(float in, bool is_sat);
```
Aaron-Hutchinson, line 46: With the componentwise functions added, I wonder if there is any reason to keep these single value functions now. Isn't this function equivalent to calling `skl_cvt_f32_f8e5m2_ref(&in, &in, 1.f, 1)`?

Aaron-Hutchinson (approved):
Looks great! Thanks again @Harlen126!

ref/cvt/cvt_ofp4.h
Harlen126: I kept the names as-is, as I haven't thought of a better way to distinguish between the "single value" and "elementwise" versions.
Since it functions more as a shared utility for other kernel tests rather than a pure reference function, I'm not sure if adding a `_ref` suffix is appropriate. Any ideas?

ref/cvt/cvt_ofp8.h
Harlen126: The "single value" versions function as utilities that can be called by other reference code (e.g., `skl_cvt_f8e4m3_f32()` is called within `ref/gemm/gemm_f8e4m3rc_f8e4m3rc_f32rc.c`).
Furthermore, not every scalar function has a corresponding componentwise version. For instance, there is no SKL conversion kernel for `f8e4m3` to `f32` array conversion.

Aaron-Hutchinson: > The "single value" versions function as utilities that can be called by other reference code

If the single value version is shared between multiple reference kernels it could just be marked as a `static` helper function. If not, I think it could just be inlined into the elementwise reference function.

> not every scalar function has a corresponding componentwise version

The scalar function could be converted into an elementwise version right? I think it's not an issue if there is no corresponding SKL library function.

ref/cvt/cvt_ofp4.h
Aaron-Hutchinson: In line with my other comment, I think a longer term goal could be to just remove the single value versions entirely, but this doesn't need to happen in this PR.

ref/cvt/cvt_ofp8.h
Harlen126: Got it, I'll address that in a separate PR. Thanks!

PR #68, opened 2026-04-01
Move scalar sources into ref

hdelassus:
This patch moves scalar reference sources out of src/ and test/ and into a ref/ directory that is built as a standalone libskl-ref static library. The goal is to reduce dependencies on standard libc/libm functions (such as scalar math functions) for integration with software that must be buildable with freestanding implementations.

* Fix transpose kernel headers depending on skl-common.h but not including it.
* Fix missing stdint header in gemm_i8_i8_i32_zve32x_x390 kernel header.

| Review focus | Reviewer |
|---|---|
| `example/conv2d` | @peichialin |
| `{ref,test}/cvt` | @Harlen126  |
| `{src,ref,test}/depthwise_conv2d` | @peichialin |
| `{src,ref,test}/gemm` | @myeh01 @Aaron-Hutchinson   |
| `{src,ref,test}/{logistic,silu,softmax}` | @ebavier |
| `{src,ref,test}/transpose` |  @Aaron-Hutchinson @Harlen126 @myeh01 |
| `.github`, `CMakeLists.txt`, `include/`, `README.md` | **all** |


Aaron-Hutchinson:
There are a few scalar functions that were previously moved to `skl-test.h` that I think should go into `libskl-ref` as well, but this could be done in a follow up PR if desired.

ref: https://github.com/sifive/skl/pull/66#discussion_r3019952496

ebavier:
I believe some of the kernel `README.md` mention or discuss scalar reference functions.  Do we want to review those to ensure its clear where those reference functions live?

hdelassus:
> There are a few scalar functions that were previously moved to `skl-test.h` that I think should go into `libskl-ref` as well, but this could be done in a follow up PR if desired.
> 
> ref: [#66 (comment)](https://github.com/sifive/skl/pull/66#discussion_r3019952496)

Thanks for the reference, I had forgotten about these and can move them as part of this PR too.

hdelassus:
> I believe some of the kernel `README.md` mention or discuss scalar reference functions. Do we want to review those to ensure its clear where those reference functions live?

I think it'd be better for clarity indeed. Let me see what I can do.

hdelassus:
> There are a few scalar functions that were previously moved to `skl-test.h` that I think should go into `libskl-ref` as well, but this could be done in a follow up PR if desired.
> 
> ref: [#66 (comment)](https://github.com/sifive/skl/pull/66#discussion_r3019952496)

Addressed in d26c6c1

README.md
```diff
@@ -48,8 +48,9 @@ Users are expected to choose the appropriate kernel variant for their target har
 
 ### Scalar Reference Routines
 
-To explain the semantics of each kernel, SKL provides a set of scalar reference routines that implement the same functionality as the vectorized kernels, but in pure scalar C code.
-These are affixed with the `_scalar` suffix, and are intended as documentation and for use in correctness testing and debugging, but are not intended to provide high performance, even on scalar hardware.
+To explain the semantics of each kernel, SKL separately provides a set of scalar reference routines that implement the same functionality as the vectorized kernels, but in pure scalar C code.
+These are affixed with the `_scalar` suffix, built into `libskl-ref.a`, and provided by the `skl-ref.h` header.
```
Aaron-Hutchinson: This feels a bit strange to me now. With this change, wouldn't it make more sense for the functions in the reference library to use a `_ref` suffix instead of `_scalar`? I know changing to `_ref` would involve a lot of work, but this mismatch is a bit jarring and `_ref` seems more accurate.

ebavier: I'd be down with this suggestion.  Especially if we're not going to any lengths to prevent compiler autovectorization there.

hdelassus: Good points. If you're fine with adding this change in this PR I'll do it.

src/depthwise_conv2d/depthwise_conv2d_f16_f16_f16_scalar.h
```diff
@@ -4,37 +4,33 @@
...
  * (Height-Width-Channel) data layout, while the filter tensor is formatted in
  * HWIM (Height-Width-Input Channel-Depth Multiplier) layout.
  */
 
 #include <stddef.h>
 
-#if !defined(__riscv_zfh)
-#error This file requires the Zfh extension
-#endif
-
```
hdelassus, line 31: @peichialin Can you please review that removing these lines is fine? Soft-float should take over in case Zfh is not enabled.

src/depthwise_conv2d/depthwise_conv2d_f16_f16_f16_scalar.c
```diff
@@ -1,16 +1,12 @@
 // Copyright 2025 SiFive, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
-#if !defined(__riscv_zfh)
-#error This file requires the Zfh extension
-#endif
-
```
hdelassus, line 7: @peichialin Here as well.

ref/cvt/cvt_ofp8.c
```diff
@@ -0,0 +1,166 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+#include "skl-common.h"
+#include <math.h>
+#include <stddef.h>
+#include <stdint.h>
+
+// NOLINTBEGIN(*)
+// TODO(fix linting errors)
+SKL_FUNC float skl_cvt_f8e4m3_f32(uint8_t in) {
```
hdelassus: @Harlen126 Can you please check that this new source file and associated header are fine? I had to disable linting, at least temporarily, because of too many warnings.

Harlen126: You can cherry-pick 9a708c05b8613c9f34bdfdb724ebb18162711971 which shall fix the linting errors.

include/skl-common.h
```diff
@@ -106,3 +106,10 @@
 SKL_FUNC_UTIL void skl_instruction_schedule_barrier(void) {
   __asm__ volatile("" ::: "memory");
 }
+
+/** @brief Reinterprets an unsigned 32-bit integer as an IEEE FP32. */
+static inline float skl_u32_as_float(uint32_t x) {
+  float y;
+  __builtin_memcpy(&y, &x, sizeof(float));
+  return y;
+}
```
Harlen126: Is it weird to have `skl_u32_as_float` in "skl-common.h" when it's only used by ref code currently?
Since it's a simple wrapper, it might be better to duplicate this utility function directly in the ref files (e.g., `ref/cvt/cvt_f4e2m1_f8e4m3.c` and `ref/cvt/cvt_ofp8.c`).

src/depthwise_conv2d/depthwise_conv2d_f16_f16_f16_scalar.h
peichialin: Thanks for addressing the issue. I agree with removing it with soft-float support.

ref/cvt/cvt_ofp8.c
hdelassus: Thank you so much @Harlen126 , I cherry-picked your changes

include/skl-common.h
hdelassus: I agree it's kind of weird. I've duplicated it across the OFP8/OFP4 conversion reference kernels.

ref/gemm/gemm_bf16rc_bf16rc_f32rc.c
Aaron-Hutchinson: Should probably update copyright year here, and in the other files that were moved.

ref/gemm/gemm_f8e4m3rc_f8e4m3rc_f32rc.h
```diff
@@ -0,0 +1,56 @@
...
+ * The entries of A and B are 8-bit floating point numbers in E4M3 format,
+ * type-punned as 8-bit unsigned integers.
+ *
+ * Matrices may be in row-major or column-major order, depending on the strides.
+ * For row-major matrices, the column stride is 1, and the row stride is the
+ * leading dimension. For column-major matrices, the reverse is true.
+ *
+ * @note
+ * This function is for API documentation purposes only, and should not be used
+ * for performance applications.
```
Aaron-Hutchinson, line 45: These notes may not be necessary now, but I'm fine with leaving them.

ref/gemm/gemm_f8e5m2rc_f8e5m2rc_f32rc.c
```diff
@@ -0,0 +1,27 @@
...
+        acc += skl_cvt_f8e5m2_f32(a[ii * rsa + kk * csa]) *
+               skl_cvt_f8e5m2_f32(b[kk * rsb + jj * csb]);
+      }
+      c[ii * rsc + jj * csc] = beta * c[ii * rsc + jj * csc] + alpha * acc;
+    }
+  }
+}
+
+#if defined(__cplusplus)
+} // extern "C"
```
Aaron-Hutchinson: I don't see the matching opening brace for this, or why it's necessary in the `.c` file.

CMakeLists.txt
```diff
@@ -23,11 +24,49 @@ set(SKL_COMPILE_OPTIONS
...
 )
 
-target_include_directories(${SKL_LIBRARY} PUBLIC
-    ${CMAKE_CURRENT_SOURCE_DIR}/include
+target_include_directories(${SKL_LIBRARY} PUBLIC include)
+target_compile_options(${SKL_LIBRARY} PRIVATE ${SKL_COMPILE_OPTIONS})
+
+# Build the reference library
+
+add_library(${REF_LIBRARY}
```
Aaron-Hutchinson: What do others think about moving the reference logic here into `ref/CMakeLists.txt`? Since the test logic is to be moved into `test/CMakeLists.txt`, I think it would be nice to eventually have the top level `CMakeLists.txt` be exclusively for building source files.

CMakeLists.txt
```diff
@@ -23,11 +24,49 @@ set(SKL_COMPILE_OPTIONS
...
+  STATIC
+  EXCLUDE_FROM_ALL
+    ref/cvt/cvt_f4e2m1_f8e4m3.c
```
Aaron-Hutchinson: We will probably eventually want to be able to conditionally compile these based on extension availability similar to `src`, but that's not a feature this PR needs to tackle.

Aaron-Hutchinson (changes requested):
Thanks for doing this massive amount of work @hdelassus!

ref/gemm/gemm_f8e5m2rc_f8e5m2rc_f32rc.c
hdelassus: Thanks for catching this!

CMakeLists.txt
hdelassus: In 7033cb13bfc4a9eb96d0a87cf92c837a04418936 I did just that. Let me know if that's close to what you are thinking.

Aaron-Hutchinson: Looks great to me!

Aaron-Hutchinson (dismissed):
LGTM, thanks again!

peichialin (approved):
LGTM~

Harlen126 (approved):
Thanks for the refactoring and cleanup!
LGTM!

PR #69, opened 2026-04-06
[Infra] Cleanup `get-riscv-extensions.cmake` extension detection

Aaron-Hutchinson:
This performs some cleanup of `get-riscv-extensions.cmake`:

- Replace the numerous `if (${SKL_PREPROCESSOR_OUTPUT} MATCHES "__riscv_{ext}")` checks with a   call to a common macro taking 3 arguments: the extension name for use in later CMakeLists.txt, the string to search for in preprocessor output, and the option(s) to pass to QEMU.
- Adds a `assert_extension_valid` function to be able to guard against typos when specifying extensions in other parts of the infrastructure.


cmake/get-riscv-extensions.cmake
```diff
@@ -23,68 +23,46 @@ if(${SKL_PREPROCESSOR_OUTPUT} MATCHES __riscv_v_min_vlen)
...
+detect_riscv_extension("zvfh" "zvfh" "zvfh=true")
+detect_riscv_extension("zvfbfwma" "zvfbfwma" "zvfbfwma=true")
+detect_riscv_extension("zvfbfmin" "zvfbfmin" "zvfbfmin=true")
+detect_riscv_extension("zve32f" "zve32f" "zve32f=true")
+detect_riscv_extension("zve64d" "zve64d" "zve64d=true")
+detect_riscv_extension("zve32x" "zve32x" "zve32x=true")
+detect_riscv_extension("zvfofp4min" "zvfofp4min" "x-zvfofp4min=true")
+detect_riscv_extension("zvfofp8min" "zvfofp8min" "x-zvfofp8min=true")
+
+detect_riscv_extension("xsfmmbase" "xsfmmbase" "x-xsfmmbase=true")
```
Aaron-Hutchinson: @myeh01 I replaced `xsfmm` with `xsfmmbase` here for consistency with the other cases, but I defer to your judgment on whether this actually makes sense.

cmake/get-riscv-extensions.cmake
```diff
@@ -23,68 +23,46 @@ if(${SKL_PREPROCESSOR_OUTPUT} MATCHES __riscv_v_min_vlen)
...
+
+function(assert_extension_valid EXT)
+  if(NOT ${EXT} IN_LIST VALID_RISCV_EXTENSIONS)
+    message(FATAL_ERROR "Provided RISC-V extension name is not valid: " ${EXT})
+  endif()
+endfunction()
+
+
+detect_riscv_extension("zfh" "zfh " "zfh=true") # The space is necessary to distinguish from zfhmin
+detect_riscv_extension("zvfh" "zvfh" "zvfh=true")
```
hdelassus: Could you also add `Zfhmin` and `Zvfhmin`?
I think for `Zvfh` here you also need the additional space to distinguish from `zvfhmin`.

hdelassus (changes requested):
LGTM but a nit for `Zvfh` detection.

cmake/get-riscv-extensions.cmake
```diff
@@ -23,68 +23,46 @@ if(${SKL_PREPROCESSOR_OUTPUT} MATCHES __riscv_v_min_vlen)
...
-endif()
+macro(detect_riscv_extension EXT_NAME EXT_PREPROCESSOR_STR QEMU_CPU_OPTION)
+  # Determine whether the given RISC-V extension is present in the current
+  # environment by checking EXT_PREPROCESSOR_STR against C preprocessor output
+  # and populate QEMU options.
+
+  # *** This is a macro because it modifies lists in the parent scope. ***
+
+  list(APPEND VALID_RISCV_EXTENSIONS "${EXT_NAME}")
+  if(${SKL_PREPROCESSOR_OUTPUT} MATCHES "__riscv_${EXT_PREPROCESSOR_STR}")
```
ebavier: Let's put single spaces on each side of the match string, use `${EXT_NAME}` and get rid of the `EXT_PREPROCESSOR_STR`.

Aaron-Hutchinson: Great idea! Done.

Aaron-Hutchinson:
I renamed the file to `detect_riscv_extensions.cmake` for accuracy and to better match the new macro name.

cmake/get-riscv-extensions.cmake
Aaron-Hutchinson: Added zfhmin and zvfhmin. Addressed the space issue with Eric's solution below.

hdelassus (approved):
LGTM!

cmake/get-riscv-extensions.cmake
myeh01: LGTM!

PR #71, opened 2026-04-08
[GEMM] Remove CLP-tuned Xsfvqdotq GEMM kernels, __builtin_memcpy

myeh01:
This PR removes the CLP-tuned Xsfvqdotq GEMM kernel and uses `ntl.p1` when available in the 6xm4 kernel to get a single schedule that performs well out of both the memory port and CLP. `__builtin_memcpy` is also replaced with `skl_memcpy` in the unaligned Xsfvqdotq kernels to avoid introducing a dependency on libc.


src/gemm/xsfvqdotq/gemm_i8rcp_i8pc_i32_xsfvqdotq.c
```diff
@@ -1196,50 +614,43 @@ SKL_FUNC_PRIVATE void skl_gemm_1xm4_aligned_i8rcp_i8pc_i32_xsfvqdotq(
...
-    skl_gemm_1xm4_i8rcp_i8pc_i32_xsfvqdotq_kernel =
-        &skl_gemm_1xm4_i8rcp_i8pc_i32_xsfvqdotq;
+  size_t m_idx = 0;
+  for (; m_idx + m0 - 1 < m; m_idx += m0) {
+    int32_t *c_write = c + m_idx * rsc;
+    const int8_t *a_tile_ptr = a_pack + m_idx * rsa1;
+    for (size_t n_idx = 0; n_idx < n; n_idx += n0) {
+      const int8_t *b_tile_ptr = b_pack + k0 * n_idx;
+      const size_t n_tile = n0 <= n - n_idx ? n0 : n - n_idx;
+      skl_gemm_1xm4_unaligned_i8rcp_i8pc_i32_xsfvqdotq(
```
JerryShih, line 715: for the unaligned a_buffer case, will it still be better to use 1xm4 instead of 6xm4_unaligned impl?

myeh01: I didn't measure the performance of the unaligned kernel, but it likely won't get good performance even if we try to optimize it. It's provided so that users have a kernel that works correctly for any A matrix, so the implementation was simplified to use a 1xm4 tiling.

JerryShih (dismissed):

src/gemm/xsfvqdotq/gemm_i8rcp_i8pc_i32_xsfvqdotq.h
```diff
@@ -70,18 +70,13 @@ extern "C" {
...
- *     c, 0, 0, rsc, 1      // c, rsc0, csc0, rsc1, csc1
- * );
+ * skl_gemm_i8rc_i8rc_i32rc_ref(m, n, k, alpha, a, rsa, 1, b, rsb, 1, beta,
+ *                              c, rsc, 1);
  * ```
  *
  * This kernel uses the SiFive Xsfvqdotq extension for vector quad widening 4D
- * dot product operations to achieve high performance on 8-bit integer data. The
+ * dot product operations to achieve high performance on 8-bit integer data.
+ * When A_pack is 4-byte aligned and rsa1 and csa1 are multiples of 4, the
```
ericlove, line 79: I think we should add a stronger note about the impact of alignment: we _really_ think users will want to pay the cost of copying into an aligned buffer in almost all cases, but provide a 1-register version as a failsafe.

ericlove (dismissed):
I think this is good! A welcome simplification, and a sensible handling of misalignment.

src/gemm/xsfvqdotq/gemm_i8rcp_i8pc_i32_xsfvqdotq.h
myeh01: Updated.

myeh01:
Need a reapproval after the latest update.

JerryShih (approved):

PR #72, opened 2026-04-08
[Infra] Export compile commands

Aaron-Hutchinson:
This sets `CMAKE_EXPORT_COMPILE_COMMANDS` to `ON` in order to assist in debugging and to allow tools like `clangd` to read the generated `compile_commands.json` file.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,6 +2,7 @@ cmake_minimum_required(VERSION 3.21)
 
 project(SKL)
 
+set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 set(DEFAULT_CMAKE_BUILD_TYPE "Release")
 if(NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES)
   message(STATUS "No build type selected, default to ${DEFAULT_CMAKE_BUILD_TYPE}")


hdelassus:
Is it a variable that should be cached, so that users can override it?

Aaron-Hutchinson:
I think since this is an environment variable it cannot be cached. But I just realized I could just do `-DCMAKE_EXPORT_COMPILE_COMMANDS=ON`, so if the global `set` is an issue I'm fine with adjusting my build flow to pass this option.

hdelassus:
> since this is an environment variable it cannot be cached.

Do you mean you can do `export CMAKE_EXPORT_COMPILE_COMMANDS=ON; cmake -B build ...`? TIL.

Is it both an env variable (aka the default value if no other configuration) and subsequently a `CACHE` variable? See [this](https://cmake.org/cmake/help/latest/envvar/CMAKE_EXPORT_COMPILE_COMMANDS.html#envvar:CMAKE_EXPORT_COMPILE_COMMANDS):
> This is a CMake [Environment Variable](https://cmake.org/cmake/help/latest/manual/cmake-language.7.html#cmake-language-environment-variables). Its initial value is taken from the calling process environment.
> The default value for [CMAKE_EXPORT_COMPILE_COMMANDS](https://cmake.org/cmake/help/latest/variable/CMAKE_EXPORT_COMPILE_COMMANDS.html#variable:CMAKE_EXPORT_COMPILE_COMMANDS) when there is no explicit configuration given on the first run while creating a new build tree. On later runs in an existing build tree the value persists in the cache as [CMAKE_EXPORT_COMPILE_COMMANDS](https://cmake.org/cmake/help/latest/variable/CMAKE_EXPORT_COMPILE_COMMANDS.html#variable:CMAKE_EXPORT_COMPILE_COMMANDS).

Aaron-Hutchinson:
I meant that you can do `cmake -DCMAKE_EXPORT_COMPILE_COMMANDS=ON .`.

Based on the link you gave it sounds like it's cached automatically with a default value taken from the environment variable of the same name, so using `set` with `CACHE` has no effect.

Aaron-Hutchinson:
I'm going to close this in favor of the above local solution. Thanks for the discussion @hdelassus!

PR #74, opened 2026-04-10
[Infra] Move examples build logic into `example`

Aaron-Hutchinson:
This does some more cleanup of the top level `CMakeLists.txt`:

- Moved the build logic for examples into `example/CMakeLists.txt`, as was/is being done for `ref` and `test`. I did some revision to the `add_skl_example` function:
  - I didn't understand the reason that the original code translated the input source list into... a list? The comment mentions backward compatibility but it's not clear to me what problems this was solving. I've removed the loop in question in the current commit.
  - Added a `REQUIRED_EXTENSIONS` argument, similar to the other `add_skl_*` functions.
  - Adjusted formatting.
- Adds a `SKL_BUILD_EXAMPLES` variable so that examples can be conditionally built with a command line option.
  - Ideally `SKL_BUILD_TESTS`, `SKL_BUILD_BENCHMARKS`, and `SKL_BUILD_EXAMPLES` should all be independent, indicating to build the corresponding objects. However the one example we currently have uses `SKL_BUILD_TESTS` and `SKL_BUILD_BENCHMARKS` and is dependent upon one of them being set. I've wrapped the `add_skl_example` calls with the appropriate condition in order to solve this problem another day.
- Replaced `SKL_ENABLE_TESTS` and `SKL_ENABLE_BENCHMARKS` in the new test files with `SKL_BUILD_TESTS` and `SKL_BUILD_BENCHMARKS` for simplicity in the top level `CMakeLists.txt`.
- Moved the `set` calls from `test/CMakeLists.txt` into `CMakeLists.txt` and added one for `SKL_BUILD_EXAMPLES`.


peichialin:
Hi @Aaron-Hutchinson ,

The `test` and `example` are different in the aspects shown below, but now they share the same (legacy) harness for verficiation of correctness (controlled by `SKL_BUILD_TESTS`) and performance profiling (controlled by `SKL_BUILD_BENCHMARKS`).

My proposal is renaming these two options and adding a new option to build examples on demand.
* `SKL_BUILD_TESTS` --> Used for adding the `test` folder into compilation target.
* `SKL_BUILD_EXAMPLES` (New) --> Used for adding the `example` folder into compilation target.
*  `SKL_ENABLE_VERIFICATION` (New, it was originally included in `SKL_BUILD_TESTS`) --> Used for enabling the verification of correctness 
* `SKL_ENABLE_PROFILING` (Renamed from `SKL_BUILD_BENCHMARKS`) --> Used for enabling the performance profiling

The follow-up questions:
* If we would like to let `example` cases reuse the new `skl-test-driver.{h,c}` implementation, shall we rename them into something like `skl-harness-driver.{h,c}` to avoid the confusion?
* The reason why skl_add_example accepts a list of source files -- The algorithm sources are put in the subdirectory of `example`, not linked to a prebuilt static library. So I passed in the test source along with the source of the algorithms (e.g. conv2d implementation) to make them be built together.

## Summary: `test/` vs `example/`

| Aspect | `test/` | `example/` |
|--------|---------|------------|
| **Tests what** | Individual SKL kernel APIs | Composed algorithms using SKL kernels |
| **Scope** | Single function from `skl.h` | Multi-step algorithm/workflow |
| **Reference** | Scalar implementation of same operation | Reference implementation of complete algorithm |
| **Purpose** | Validate SKL library correctness | Demonstrate how to use SKL library |
| **Example** | `skl_gemm_f32_f32_f32_zve32f_x390()` | `conv2d()` using im2row + `skl_gemm_*()` |
| **Audience** | SKL developers/CI | SKL users/integrators |

Welcome to any feedback!

Aaron-Hutchinson:
@peichialin Thanks for the thorough suggestion! I think it makes sense, and I updated the logic to use `SKL_BUILD_TESTS`, `SKL_BUILD_EXAMPLES`, `SKL_ENABLE_VERIFICATION`, and `SKL_ENABLE_PROFILING` as you suggested. Please have a look at the changes.

Note that this means `SKL_BUILD_BENCHMARKS` will eventually be deprecated, but is still currently used by legacy test harnesses. I've updated the example to be independent of it already.

Aaron-Hutchinson:
Requesting @myeh01 to review the changes to CI being made here.

peichialin:
@Aaron-Hutchinson
The overall changes of examples LGTM! However, if it still depends on `test/skl-test.h` in short-term stage, then we have to replace the usage of `ENABLE_TEST` and `ENABLE_BENCHMARK` by `SKL_ENABLE_VERIFICATION` and `SKL_ENABLE_PROFILING`.
@Aaron-Hutchinson @ericlove 
BTW, after all "tests" are refactored to use new test harness, then "examples" will be the only user of the legacy "skl-test.h". The examples are now designed as self-contained programs with a main functon. I believe it has high readibility without being implemented under complex test harness, but has restriction of preparing multiple testcases in a single program in the same coding style. If the intention is demonstrating the capbiliaty of using single/multiple SKL function(s) to form a more complex algorithm, but not claiming a high test coverage, then keeping it in the original form should be acceptable. In this case, we can consider to move the legacy "skl-test.h" under examples if there is no other user outside of "examples".
Migrating to new test harness can also be considered if we'd like to avoid maintaining duplicated implememtations of test/example harness. But it won't block the open souce schedule.

Aaron-Hutchinson:
@peichialin Based on our group discussion yesterday, I believe we agreed that each example should be self contained. So the example should not utilize the new test driver, and a follow up PR should be made to somehow remove the dependence on `skl-test.h`.

.github/workflows/ci.yml
```diff
@@ -63,35 +63,44 @@ jobs:
           egrep "\.(c|h)$" | \
           xargs -I{} bash -c "if [ -f {} ] ; then clang-format --dry-run -style=file --Werror {} ; fi"
 
-    - name: Configure and build (tests and benchmarks)
+    - name: Configure and build (tests and examples with profiling and verification)
       run: |
         cmake -B ${{ github.workspace }}/build \
               -DCMAKE_TOOLCHAIN_FILE=cmake/riscv.cmake \
-              -DSKL_BUILD_TESTS=ON -DSKL_BUILD_BENCHMARKS=ON
+              -DSKL_BUILD_TESTS=ON \
```
myeh01, line 71: Don't we still want to set `-DSKL_BUILD_BENCHMARKS` to `ON` or `OFF`? The benchmark part of the remaining legacy tests won't be tested in CI.

But after all tests have been converted to the new format, I think we'll only need to run one test with all four flags enabled.

Aaron-Hutchinson: Yeah, that's a good point. I added it back using the same value as the original code.

Oh, I guess before we were only testing all combinations since the harnesses sometimes branched based on both values simultaneously. But in the new setup they're independent (or at least that's the intent).

myeh01 (dismissed):
CI and GEMM changes LGTM.

.github/workflows/ci.yml
```diff
@@ -63,35 +63,47 @@ jobs:
           egrep "\.(c|h)$" | \
           xargs -I{} bash -c "if [ -f {} ] ; then clang-format --dry-run -style=file --Werror {} ; fi"
 
-    - name: Configure and build (tests and benchmarks)
+    - name: Configure and build (tests and examples with profiling and verification)
```
hdelassus: We have not enabled any profiling, have we?

example/conv2d/main.c
```diff
@@ -124,7 +124,7 @@ static int check_error(const char *name, const float *res, const float *ref,
 
 int main(void) {
   int ret = 0; // return value
-#if defined(ENABLE_BENCHMARK)
+#if defined(SKL_ENABLE_PROFILING)
```
hdelassus: OK I see. I don't think profiling is the right word for what the code does. Benchmarking seemed more appropriate.

CMakeLists.txt
```diff
@@ -154,52 +178,9 @@ function(skl_add_test_legacy TEST_NAME TEST_SRC OPTS)
...
-  target_include_directories(example-${EXAMPLE_NAME} PRIVATE test src example)
-  add_test(
-    NAME example-${EXAMPLE_NAME}
-    COMMAND
-    ${SKL_TESTER} ${SKL_TESTER_OPTIONS} $<TARGET_FILE:example-${EXAMPLE_NAME}>
-  )
-endfunction()
-
 if (SKL_BUILD_TESTS OR SKL_BUILD_BENCHMARKS)
   enable_testing()
```
hdelassus, line 172: Delete as well now?

hdelassus:
How about these changes?

* `SKL_ENABLE_VERIFICATION` -> `SKL_ENABLE_VALIDATION` (less loaded than verification)
* `SKL_ENABLE_PROFILING` -> `SKL_ENABLE_BENCHMARKING`

Aaron-Hutchinson:
@hdelassus Sounds fine to me, I made the change

CMakeLists.txt
Aaron-Hutchinson: Do you mean the `enable_testing()` line? I think it's still needed in case only `SKL_BUILD_BENCHMARKS` is set.

.github/workflows/ci.yml
```diff
@@ -63,35 +63,47 @@ jobs:
...
+    - name: Configure and build (tests and examples with benchmarking and validation)
       run: |
         cmake -B ${{ github.workspace }}/build \
               -DCMAKE_TOOLCHAIN_FILE=cmake/riscv.cmake \
-              -DSKL_BUILD_TESTS=ON -DSKL_BUILD_BENCHMARKS=ON
+              -DSKL_BUILD_BENCHMARKS=ON \
+              -DSKL_BUILD_TESTS=ON \
+              -DSKL_BUILD_EXAMPLES=ON \
+              -DSKL_ENABLE_BENCHMARKING=ON \
+              -DSKL_ENABLE_VALIDATION=ON
```
ebavier: What's the difference between "enabling" benchmarks and validation, and "building" benchmarks and validation?

CMakeLists.txt
```diff
@@ -127,6 +127,30 @@ endif()
 
 set(SKL_TEST_LIBS skl-ref m)
 
+set(SKL_BUILD_TESTS OFF CACHE STRING "Enable building of validation tests and/or benchmarks.")
```
ebavier: I don't like this.  The only purpose seems to be to control whether things in the `test` directory are configured.  Can this not simply be determined by `enable_tests || enable_benchmarking`?

CMakeLists.txt
```diff
@@ -127,6 +127,30 @@ endif()
...
+message(STATUS "SKL_BUILD_TESTS: ${SKL_BUILD_TESTS}")
+
+set(SKL_BUILD_EXAMPLES OFF CACHE STRING "Enable building of examples.")
```
ebavier: Similar here.  I don't think examples ultimately should contain either "tests" or benchmarks.  Save for the test of whether the examples compile and run successfully; and to that end, I don't think being able to disable their build is useful.

ebavier (changes requested):
I can appreciate the namesake change of moving relevant cmake logic into the `example` directory.

But the terminology changes ("verification") seem unnecessary to me, since they provide no additional clarification to the concepts.

CMakeLists.txt
Aaron-Hutchinson: `SKL_BUILD_TESTS` and `SKL_BUILD_EXAMPLES` control whether the `test` and `example` directories are entered at all during the build. `SKL_ENABLE_VALIDATION` and `SKL_ENABLE_BENCHMARKING` control the type of tests to run and what the example does. The `BUILD` and `ENABLE` variables have been separated to be able to conditionally build examples and tests independently.

CMakeLists.txt
Aaron-Hutchinson: The current example we have does both testing and benchmarking, and being able to disable it is useful for reducing build times during development.

ebavier: How much time?

Aaron-Hutchinson: It takes several seconds to run the existing examples, which would get worse as more are added. Also it doesn't really make sense to have to build the entire test suite if you just wanted to run the example.

What exactly is the issue with allowing these to be build independently?

CMakeLists.txt
hdelassus: Yes, but you are right!

CMakeLists.txt
ebavier: A general aversion to proliferation of configuration options.

Is `ctest` regex filtering not enough to target test runs to what you're interested in?

Could we instead add a custom cmake target for building just the examples?

hdelassus:
Noting that the root `README.md` would need to be updated (either in this PR or later)

CMakeLists.txt
Aaron-Hutchinson: I appreciate your concern about keeping the options in check, and thanks for offering these alternatives. I think a custom target for `tests` and `examples` plus a regex filter might be a sufficient solution. I'll revert most of the variable changes being made here and address custom targets in a separate PR to keep the scope here contained.

ebavier: >  I'll revert most of the variable changes being made here and address custom targets in a separate PR to keep the scope here contained.

Great, this looks much better.  I think there are some adjustments to be made in `ci.yml` still?

Aaron-Hutchinson: Not in this PR I think. Once all of the legacy tests are updated the `SKL_BUILD_TESTS` and `SKL_BUILD_BENCHMARKS` options can be removed.

ebavier: Oh, I see, right, the legacy tests will still use for a bit the "build" macros. 👍

ebavier (approved):

Aaron-Hutchinson:
@hdelassus Thanks for noticing this. I'd rather address the README in a separate PR since it needs more revision than just these CMake variable names. It was never updated when the test driver was added so it's pretty out of date at this point.

PR #75, opened 2026-04-11
[GEMM] Optimize large widening f16 x390 kernel

Aaron-Hutchinson:
This effectively rewrites the `skl_gemm_4xm4x1_f16_f16_f32_zvfh_x390` kernel to provide better performance. The new version uses inline assembly with a 6xm4x12 tile size (so the `vfwmacc.vf`s are still done at LMUL=2) and preloads the A and B matrix values a few iterations in advance by making the scalar loads non-blocking with the ntl.p1 instruction.

My benchmarking showed a substantial performance boost, with the new version giving up to 80% improvement over the original kernel in certain cases.

Note that the `__asm__` statement for the main loop used an extremely long assembler template string, for which I had to locally suppress the following compiler warning:
```
error: string literal of length 6475 exceeds maximum length 4095 that ISO C99 compilers are required to support [-Werror,-Woverlength-strings]
```

### DIFF
--- a/src/gemm/rvv/gemm_f16_f16_f32_zvfh_x390.c
+++ b/src/gemm/rvv/gemm_f16_f16_f32_zvfh_x390.c
@@ -10,6 +10,12 @@
 
 #include "skl-common.h"
 
+#if defined(__riscv_zihintntl)
+#define NTL_P1 "ntl.p1 \n\t"
+#else
+#define NTL_P1
+#endif
+
 /**
  * @brief RVV float16 vector-matrix multiply with float32 output for row-major
  * matrices, tuned for X390.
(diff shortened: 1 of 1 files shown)


hdelassus (dismissed):
Thanks for this work! I can't say I haven't missed anything, but overall it looks good to go for me. 👌

src/gemm/rvv/gemm_f16_f16_f32_zvfh_x390.c
```diff
@@ -219,125 +225,777 @@ skl_gemm_1xm8x16_f16_f16_f32_zvfh_x390(size_t n, size_t k, float alpha,
...
+              [acc2] "+&vr"(acc2),
+              [acc3] "+&vr"(acc3),
+              [acc4] "+&vr"(acc4),
+              [acc5] "+&vr"(acc5),
+              [b_addr] "+&r"(b_addr),
+              [a_addr_1] "=r" (a_addr_1),
+              [a_addr_2] "=r" (a_addr_2),
+              [a_addr_3] "=r" (a_addr_3),
+              [a_addr_4] "=r" (a_addr_4),
+              [a_addr_5] "=r" (a_addr_5)
```
hdelassus: Shouldn't all of these be early-clobbered?

Aaron-Hutchinson: Good catch!

hdelassus (approved):
LGTM!

PR #77, opened 2026-04-14
[Test][GEMM] Update Xsfvqdotq GEMM tests

myeh01:
This PR updates the Xsfvqdotq tests and benchmarks to the style introduced in #56.


Aaron-Hutchinson:
It looks like this branch might need a rebase after the merge base changed? The diff looks to be including changes from #71.

test/gemm/gemm_i8rcp_i8pc_i32_xsfvqdotq.c
```diff
@@ -0,0 +1,215 @@
...
+#include "skl-test-driver.h"
+#include "skl_test_gemm.h"
+#include <inttypes.h>
+#include <stddef.h>
+#include <stdint.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+static void skl_gemm_i8rcp_i8pc_i32_xsfvqdotq_ref(
```
Aaron-Hutchinson: Shouldn't we just put this in `skl-ref` so that the kernel docstring can refer to it instead of having the `k % 4 == 0` restriction?
https://github.com/sifive/skl/blob/1188c2f9816ce15a7cbe1b5c56e9330fc995178d/src/gemm/xsfvqdotq/gemm_i8rcp_i8pc_i32_xsfvqdotq.h#L41-L51

test/gemm/gemm_f32rcprc_f32rcprc_f32rcprc.c
```diff
@@ -20,6 +20,7 @@
 #include <stdlib.h>
 #include <string.h>
 
+// NOLINTBEGIN(readability-function-cognitive-complexity)
```
Aaron-Hutchinson, line 25: Why did this need to be added?

test/gemm/xsfvqdotq/skl_gemm_i8rcp_i8pc_i32_xsfvqdotq.c
```diff
@@ -0,0 +1,106 @@
+// Copyright 2026 SiFive, Inc.
+// SPDX-License-Identifier: Apache-2.0
```
Aaron-Hutchinson: This will need an update after #88 merges.

Aaron-Hutchinson (commented):
Looks good overall. Just a couple questions.

test/gemm/gemm_f32rcprc_f32rcprc_f32rcprc.c
myeh01: The "cognitive complexity" of this function exceeded the threshold.
https://clang.llvm.org/extra/clang-tidy/checks/readability/function-cognitive-complexity.html
I recently encountered a similar warning for the packed f32 GEMM test harness (maybe a toolchain update caused this?), so that will need to be updated too. We could also increase the threshold instead.

test/gemm/gemm_i8rcp_i8pc_i32_xsfvqdotq.c
myeh01: Maybe @hdelassus or @ericlove can give their thoughts? This kernel has an unusual API that's a mix of the GEMM and packed GEMM APIs and is not quite correctly described with `gemm_i8rcp_i8pc_i32`, even for the vectorized routine. Or maybe we could add something like `any_k` in the specialization field of the name to indicate this, so we would have
```
skl_gemm_any_k_i8rcp_i8pc_i32_xsfvqdotq
skl_gemm_any_k_i8rcp_i8pc_i32_ref
```
Let me know what you think.

test/gemm/gemm_f32rcprc_f32rcprc_f32rcprc.c
Aaron-Hutchinson: Right, I was more asking about why it was an issue here but not in #81. Maybe the change to `SKL_TEST_BUF_CREATE` pushed it over the limit?

I'm generally not in favor of selectively disabling lint checks or increasing thresholds instead of just addressing the underlying issue causing the warning.

test/skl-test-driver.h
```diff
@@ -423,7 +423,7 @@ typedef enum {
       }                                                                        \
       break;                                                                   \
     case SKL_TEST_SEQ: {                                                       \
-      TYPE step = ((BUF)->max - (BUF)->min) / (BUF)->len;                      \
+      TYPE step = (BUF)->len ? ((BUF)->max - (BUF)->min) / (BUF)->len : 0;     \
```
Aaron-Hutchinson, line 428: ~I can't find the motivation for this. Is a step of 0 being used somewhere?~

Sorry, I misread it. I guess this is to call `SKL_TEST_BUF_CREATE` when `len` is 0.

myeh01: Yeah the linter complained about possible division by 0.

test/gemm/gemm_f32rcprc_f32rcprc_f32rcprc.c
myeh01: Yeah, it looks like the check on whether `(BUF)->len` was 0 pushed it over.

test/skl-test-driver.h
Aaron-Hutchinson: I'm still trying to figure out why this was an issue in this PR but not #81. Does it only complain when `SKL_TEST_SEQ` is used somewhere? I haven't encountered this error yet myself.

test/gemm/gemm_i8rcp_i8pc_i32_xsfvqdotq.h
```diff
@@ -0,0 +1,61 @@
+// Copyright 2026 SiFive, Inc.
```
pvchupin: Should be updated as well

test/gemm/gemm_i8rcp_i8pc_i32_xsfvqdotq.c
```diff
@@ -0,0 +1,215 @@
+// Copyright 2026 SiFive, Inc.
```
pvchupin: should be updated

test/skl-test-driver.h
myeh01: I think I see now. For float types, the division becomes floating point division, which allows division by zero. For int types, division by zero is undefined.

Aaron-Hutchinson: Oh! That makes sense, but it raises a new question on why clang-tidy didn't have an issue with the transpose PR #84. Are unsigned types treated differently here? I can't find anything that would suggest that to be the case.

myeh01: I think it only complains if there is a branch that explicitly sets the length to 0. In the transpose tests, the buffer lengths are just computed as
```
h->a.len = h->m * h->rsa;
h->at.len = h->n * h->rsat;
```
whereas in the GEMM tests, the array length is explicitly set to 0 if one of the matrix dimensions is 0.

Aaron-Hutchinson: Ah, I see. Thanks for sorting out this explanation.

test/gemm/gemm_i8rcp_i8pc_i32_xsfvqdotq.c
Aaron-Hutchinson: What was the conclusion on this? I see that the reference function that was initially added is gone now.

myeh01: Still under discussion.

PR #78, opened 2026-04-15
[Test] Convert depthwise conv2d test harness

peichialin:
[Key changes]
* Create three testharnesses for i8/f16/f32. Each defines the members and fundemantal functions (init, verify, report, cleanup) for specialized kernel tests.
* To offer the flexibility of adding new specializations based on custom extensions in the future, we maintain the current tests based on standard rvv extension in a `rvv` subfolder.
* More tests are added to increase the test coverage: different combinations of strides, dilations and depthwise_conv2d.
* The verification of f16/f32 are refined by the same bound analysis used in GEMM.

[Open for discussions]
* We defined several public APIs for a single data type to bring in the specializations for different attributes (filter size, strides, dilations, depth multiplier). The users were asked to pick the one they really need, so they can restrict the code size. An alternative approach is to provide a generic API which is responsible for dispatching to the private specializations in the runtime. That will include all specializations during compilation, but can simplify the user code (and the test code as well).


ref/depthwise_conv2d/depthwise_conv2d_f64_f64_f64.h
```diff
@@ -0,0 +1,63 @@
...
+
+#pragma once
+
+/**
+ * @file depthwise_conv2d_f64_f64_f64.h
+ * @brief Reference float64 depthwise convolution 2D kernels.
+ *
+ * This header provides reference implementations of depthwise convolution 2D
+ * operations for float64 input/filter/output data. This implementation is
+ * primarily used for computing error bounds in test harnesses.
```
Aaron-Hutchinson: > This implementation is primarily used for computing error bounds in test harnesses.

I think it's likely safe to just omit this statement.

test/depthwise_conv2d/depthwise_conv2d_f16_f16_f16.h
```diff
@@ -0,0 +1,49 @@
...
+
+  // Derived parameters & buffers (private to the test harness)
+  struct {
+    double *input_abs;    // Absolute values of input (for bound computation)
+    double *filter_abs;   // Absolute values of filter (for bound computation)
+    _Float16 *ref_output; // Reference output
+    double *bound;        // Per-element error bound
+  } ctx;
+
+  uint32_t use_specialization;
```
Aaron-Hutchinson: Nit: `use_specialization` appears to be a configurable parameter, so maybe should be placed above the "derived parameters" comment.

test/CMakeLists.txt
```diff
@@ -69,3 +69,24 @@ skl_add_test(
   "xsfmm32a32f"
   ""
 )
+skl_add_test(
```
Aaron-Hutchinson, line 71: How should we sort the various `skl_add_test` calls? Maybe alphabetically by the first argument `TEST_NAME`?

Aaron-Hutchinson (commented):
Thanks for doing these conversions! The test structure matches my expectation exactly and looks great. I'm not qualified to review the details of the depthwise conv2d specifics, so I'll defer to others to provide feedback on those portions.

Very exciting seeing the improvements to the test suite coming to fruition!

test/depthwise_conv2d/rvv/depthwise_conv2d_i8_i8_i32_zve32x.c
```diff
@@ -0,0 +1,264 @@
...
+#ifdef SKL_ENABLE_BENCHMARKS
+  // Benchmark tests
+{BENCH, 
+   .input_height = 16, .input_width = 16, .input_channel = 512, 
+   .filter_height = 3, .filter_width = 3, 
+   .output_height = 14, .output_width = 14, .output_channel = 512, 
+   .depth_multiplier = 1, .stride_height = 1, .stride_width = 1, 
+   .dilation_height_factor = 1, .dilation_width_factor = 1,
+   .input_zero_point = 0
+  },
```
Harlen126: Do you intend to only benchmark the non-specialized one?

test/depthwise_conv2d/depthwise_conv2d_f16_f16_f16.c
```diff
@@ -1,155 +1,177 @@
...
+  SKL_TEST_BUF_FREE(t, &h->filter);
+  SKL_TEST_BUF_FREE(t, &h->output);
 
-  return ret > 0;
+  if (h->steps.verify) {
+    free(h->ctx.input_abs);
+    free(h->ctx.filter_abs);
+    free(h->ctx.ref_output);
+    free(h->ctx.bound);
+  }
```
Harlen126, line 182: Maybe free them only when the length is greater than "0".

peichialin:
@Aaron-Hutchinson @Harlen126 
I updated PR based on your comments, please see if the changes address them. Thanks!

Harlen126 (approved):
LGTM!

Aaron-Hutchinson (approved):
LGTM, thanks!

PR #79, opened 2026-04-15
[Softmax] Improves Xsfvfexp32e performance for "small" 2D sizes.

ebavier:
New "mvec" kernels improve performance in cases where N, the number of matrix columns, is less than 2*__riscv_vsetvlmax_e32m8().

* src/softmax/softmax_f32_xsfvfexp32e.c (skl_softmax_2d_mvec_f32_xsfvfexp32e): New high-level dispatch.
(skl_softmax_2d_f32_xsfvfexp32e): Use it when N and M are appropriately sized.

### DIFF
--- a/src/softmax/softmax_f32_xsfvfexp32e.c
+++ b/src/softmax/softmax_f32_xsfvfexp32e.c
@@ -1,4 +1,4 @@
-// Copyright 2025 SiFive, Inc.
+// Copyright 2026 SiFive, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
 #if !defined(__riscv_xsfvfexp32e)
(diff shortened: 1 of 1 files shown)


src/softmax/softmax_f32_xsfvfexp32e.c
```diff
@@ -195,3 +196,740 @@ SKL_FUNC void skl_softmax_2d_f32_xsfvfexp32e(float *s, const size_t rss,
...
+
+  vfloat32m1_t R = __riscv_vfmul_vv_f32m1(r, r, vl);
+  vfloat32m1_t p89a = __riscv_vfmacc_vv_f32m1(c8, r, p9a, vl);
+
+  vfloat32m1_t S = __riscv_vfmul_vv_f32m1(R, R, vl);
+  vfloat32m1_t p2345 = __riscv_vfmacc_vv_f32m1(p23, R, p45, vl);
+  vfloat32m1_t p6789a = __riscv_vfmacc_vv_f32m1(p67, R, p89a, vl);
+
+  vfloat32m1_t p = __riscv_vfmacc_vv_f32m1(p2345, S, p6789a, vl);
+  p = __riscv_vfmacc_vv_f32m1(r, R, p, vl);
```
hdelassus, line 612: I'm curious if the choice of an Estrin evaluation scheme was motivated by something in particular vs a Horner scheme?

hdelassus (approved):
Wow, that was quite complex to review for me, but FWIW it looks good to me. I assume tests for both variants will be added by #83 ?

ebavier:
> Wow, that was quite complex to review for me, but FWIW it looks good to me.

I thought it might be tough, thanks for looking over!

> I assume tests for both variants will be added by #83 ?

More extensive tests, yes.  Until then, `test/softmax/softmax_2d_f32.c` covers tests for the 2D functions.

src/softmax/softmax_f32_xsfvfexp32e.c
ebavier: Because this function is evaluated at LMUL=1, leaving plenty of space for many temporaries, the Estrin evaluation scheme was chosen to hopefully use any architectural parallelism available.

PR #80, opened 2026-04-21
[Test] Port exp_f32 tests to new driver

hdelassus:
This converts the FP32 exponential tests to use the new test driver, and lays the foundation to convert other unary functions tests using similar patterns. To match the style of the existing tests style, I have split the original tests into one for each extension used by the different kernels. The `test/math/unary_f32.[ch]` sources provide common FP32 unary function harness types and functions that can be shared between different kernels.

* CMakeLists.txt: Remove legacy test.
* include/skl-ref.h: Add FP32 and FP16 reference kernels.
* ref/CMakeLists.txt: Add sources to the build.
* ref/exp/exp_f16.c,
* ref/exp/exp_f16.h,
* ref/exp/exp_f32.c,
* ref/exp/exp_f32.h: New files.
* test/CMakeLists.txt: Add new tests.
* test/README.md: Fix typos.
* test/exp/exp_f32.c: Delete.
* test/exp/exp_f32_xsfvfexp32e.c,
* test/exp/exp_f32_xsfvfexpa.c,
* test/exp/exp_f32_zve32f.c: New tests.
* test/math/unary_f32.c,
* test/math/unary_f32.h: Generic harness for FP32 unary functions.
* test/skl-test-driver.h: Add utility functions for test harnesses.


test/exp/exp_f32_xsfvfexp32e.c
```diff
@@ -0,0 +1,55 @@
...
+static skl_test_suite_t suite = {.name = "skl_exp_f32_xsfvfexp32e",
+                                 .num_tests = sizeof(tests) / sizeof(tests[0]),
+                                 .test_size = sizeof(unary_f32_t),
+                                 .tests = tests};
+
+static void execute(skl_test_t *t) {
+  unary_f32_t *h = (unary_f32_t *)t->harness;
+  unary_func_t func = h->func;
+  func(h->ctx.b, h->a.data, h->a.len);
+}
```
ebavier: It seems like this function is not used?  The `unary_f32_execute` function from the harness is in `steps`.

test/exp/exp_f32_xsfvfexp32e.c
```diff
@@ -0,0 +1,55 @@
...
+
+unary_f32_t tests[] = {
+#if defined(__riscv_xsfvfexp32e)
+#if defined(SKL_ENABLE_BENCHMARKS)
+    FUNCTION_BENCHMARKS(skl_exp_2p398u0alt64ainf_f32_xsfvfexp32e,
+                        -0x1.fffffep5f, 0x1.fffffep5f),
+    FUNCTION_BENCHMARKS(skl_exp_5p32u_f32_xsfvfexp32e, MIN, MAX),
+#endif
+#if defined(SKL_ENABLE_TESTS)
+    FUNCTION_TESTS(skl_exp_2p398u0alt64ainf_f32_xsfvfexp32e, -0x1.fffffep5f,
```
ebavier: We don't know here what scalar reference function will be run.  And since `skl_abs_error_ulp` compares `float` inputs, the error returned will most often be an integer (unless values cross a binade).  I think this means that the tolerance needs to be something like `floor(u_f + u_r)` if `u_f` is the max ulp error of the function under test and `u_r` is the max ulp error of the reference function.  E.g. if the reference `expf` is from Newlib, we need `floor(2.398 + 0.911) -> 3`?

test/math/unary_f32.c
```diff
@@ -0,0 +1,92 @@
...
+
+  float ulp = 0.f;
+  for (size_t i = 0; i < h->a.len; ++i) {
+    float Y = h->ctx.b[i];
+    float y = h->ctx.ref[i];
+    float err = skl_abs_error_ulp_f32(Y, y);
+    if (err > max)
+      SKL_TEST_LOG(t, SKL_TEST_LOG_ERROR,
+                   "[%4d]: %15.6a %15.6a : %.4g ulp [bound = %.4g ulp]\n", i, Y,
+                   y, err, max);
```
ebavier: - Would it make any sense to limit the number of error messages printed?
- In the case of a failure is there enough information printed to isolate the test case?
- Do we know if `printf` will always have support for `%f` and `%g`?

test/math/unary_f32.c
```diff
@@ -0,0 +1,92 @@
...
+  unary_f32_t *h = (unary_f32_t *)t->harness;
+  // Allocate buffers
+  SKL_TEST_BUF_CREATE(t, float, &h->a);
+  h->ctx.b = malloc(h->a.len * sizeof(float));
+  if (!h->ctx.b)
+    t->status.init_status = SKL_TEST_FAIL;
+  if (h->steps.verify) {
+    h->ctx.ref = malloc(h->a.len * sizeof(float));
+    if (!h->ctx.ref)
+      t->status.init_status = SKL_TEST_FAIL;
```
ebavier: How about checking `h->ctx.b` and `h->ctx.ref` at the end, so that that `t->status.init_status` is set just once?

test/exp/exp_f32_xsfvfexpa.c
```diff
@@ -0,0 +1,49 @@
...
+#if defined(__riscv_xsfvfexpa)
+#if defined(SKL_ENABLE_BENCHMARKS)
+    FUNCTION_BENCHMARKS(skl_exp_1u_f32_xsfvfexpa, MIN, MAX),
+    FUNCTION_BENCHMARKS(skl_exp_1p0002ugen5d639eP6s0_f32_xsfvfexpa,
+                        -0x1.5d639ep6f, +89.f),
+#endif
+#if defined(SKL_ENABLE_TESTS)
+    FUNCTION_TESTS(skl_exp_1u_f32_xsfvfexpa, MIN, MAX, 1.0f),
+    FUNCTION_TESTS(skl_exp_1p0002ugen5d639eP6s0_f32_xsfvfexpa, -0x1.5d639ep6f,
+                   +89.f, 1.0002f),
```
ebavier: Maybe `MAX` here in place of `+89.f`?

test/exp/exp_f32_xsfvfexp32e.c
```diff
@@ -0,0 +1,55 @@
...
+
+#define MIN (-104.f)
+#define MAX (+89.f)
+
+#define FUNCTION_TESTS(FUN, MIN, MAX, ULP)                                     \
+  {TEST,           .ref_func = skl_exp_f32_ref, .func = (FUN), .a.min = (MIN), \
+   .a.max = (MAX), .ctx.max_err = (ULP)}
+
+#define FUNCTION_BENCHMARKS(FUN, MIN, MAX)                                     \
+  {BENCH, .func = (FUN), .a.min = (MIN), .a.max = (MAX)}
```
ebavier: These macros are nice.  Too bad they're replicated in each extension's source.
Previous `exp` tester lumped all extensions together, selectively enabling parts depending both on availability and configuration macros.

test/exp/exp_f32_xsfvfexp32e.c
```diff
@@ -0,0 +1,55 @@
...
+
+unary_f32_t tests[] = {
+#if defined(__riscv_xsfvfexp32e)
```
ebavier: Based on the guard at the top of the file, we know here that `__riscv_xsfvfexp32e` _will_ be defined.

test/math/unary_f32.h
```diff
@@ -0,0 +1,63 @@
...
+ *  - Expected maximum numerical error of the result
+ *  - The kernel function to test
+ */
+
+#include "skl-test-driver.h"
+#include <stddef.h>
+#include <stdint.h>
+
+// Type of SKL unary functions
+typedef void (*unary_func_t)(float *, const float *, size_t);
```
ebavier: For consistency, maybe best to have `f32` somewhere in this function type's name?   Otherwise the name might clash at some point with an `f16` unary function type.

test/math/unary_f32.c
```diff
@@ -0,0 +1,92 @@
...
+}
+
+void unary_f32_verify(skl_test_t *t) {
+  unary_f32_t *h = (unary_f32_t *)t->harness;
+
+  // The ulp tolerance for this test.
+  float max = h->ctx.max_err;
+
+  // Compute the reference result
+  skl_exp_f32_ref(h->ctx.ref, h->a.data, h->a.len);
```
ebavier: Do we need to check whether input data is still intact (e.g. not clobbered by an out-of-bounds write) before computing a reference result?  Or will that be left to ASAN tests to check?

test/math/unary_f32.c
```diff
@@ -0,0 +1,92 @@
+// Copyright 2026 SiFive, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+/**
+ * @brief Implementation of the exp_f32 test harness.
```
ebavier:
```suggestion
 * @brief Generic test harness for f32 unary element-wise functions.
```

test/math/unary_f32.c
```diff
@@ -0,0 +1,92 @@
...
+ *
+ * This file defines all harness functions _except_ `skl_test_execute`, which is
+ * defined in the test file (e.g. skl_exp_f32_zve32f.c).
```
ebavier:
```suggestion
```

ref/exp/exp_f32.h
```diff
@@ -0,0 +1,29 @@
...
+
+/**
+ * @brief Reference FP32 Exponential function.
+ *
+ * @param out - Array of output elements.
+ * @param in - Array of input elements.
+ * @param n - Number of elements to process.
+ *
+ * Computes the exponential function e^x.
+
```
ebavier:
```suggestion
 *
```

ref/exp/exp_f16.h
```diff
@@ -0,0 +1,29 @@
...
+
+/**
+ * @brief Reference FP16 Exponential function.
+ *
+ * @param out - Array of output elements.
+ * @param in - Array of input elements.
+ * @param n - Number of elements to process.
+ *
+ * Computes the exponential function e^x.
+
```
ebavier:
```suggestion
 *
```

ebavier (changes requested):

test/math/unary_f32.h
```diff
@@ -0,0 +1,63 @@
...
+  // Configurable parameters
+  unary_func_t ref_func; // The reference function
+  unary_func_t func;     // The function to test
+
+  // Buffer generation settings
+  SKL_TEST_BUFFER(float) a;
+
+  // Derived parameters and buffers (private to the test harness)
+  struct {
+    float *b;      // result buffer
```
Aaron-Hutchinson: Wouldn't it be clearer to rename `a` and `b` to `in` and `out`? I thought we had agreed to standardize these names at some point.

test/skl-test-driver.h
Aaron-Hutchinson, line 1: The functions being added here aren't relevant to the driver interface, so I don't think they belong in this file. Arguably some of the existing content shouldn't be placed there either. Maybe we should make a new `skl-test-utils.h` header, or re-use the existing `skl-test.h`. What do you guys think?

test/elementwise/unary_f32.c
Aaron-Hutchinson, line 1: The directory name `math` seems a bit too general. How about `elementwise` or just `unary`? Not sure what else you planned to go in there.

test/CMakeLists.txt
```diff
@@ -89,3 +89,24 @@ skl_add_test(
   "xsfmm32a32f"
   ""
 )
+skl_add_test(
```
Aaron-Hutchinson, line 120: I think the consensus among the PRs is to sort these calls alphabetically by the first argument.

test/exp/exp_f32_xsfvfexp32e.c
hdelassus: Oops, that's a leftover from refactoring. Thanks for catching it!

test/exp/exp_f32_xsfvfexp32e.c
hdelassus: Updated. Probably need the same for other source files.

test/math/unary_f32.c
hdelassus: * In some cases, I would think so. Do you think we should have a hardcoded maximum? 10 sounds reasonable to me, but we could also make it a configurable value in each test harness.
* Technically, I think so (as long as the PRNG seed is constant), but practically, not so much. Printing the input would be valuable, I think.
* I think `%f` and `%g` are supported with all C89-compliant libcs. The `%a` specifier might not be supported in some Newlib builds ([since 2007](https://sourceware.org/git/?p=newlib-cygwin.git;a=commitdiff;h=0962fe9178a40c8410d5f1497bb1966f0db66d6d), those built without `_WANT_IO_C99_FORMATS`), but I don't know if there is a way to know this at runtime.

ebavier: E.g. when multiple `exp` functions are run in a suite, is it possible to know which function an error belongs to?

test/exp/exp_f32_xsfvfexp32e.c
hdelassus: I'd like to ask @Aaron-Hutchinson what he thinks of this. We could certainly share more code among each test, or merge different kernel tests as it was done before, but I don't know if that would match the intent of the new test driver?

test/math/unary_f32.c
hdelassus: I would think OOB reads or writes are best left for ASAN to check? I'm open to discussion.

test/math/unary_f32.h
hdelassus: I'm a bit confused because the FP32 GEMM test harness uses `a`, `b`, `c`, and `ref_c`. Did I miss something?

test/elementwise/unary_f32.c
hdelassus: `elementwise` sounds fine.
@ebavier any thoughts or suggestions?

test/skl-test-driver.h
hdelassus: Yes, I agree. A new file would perhaps be best, as `skl-test.h` is eventually going to be deleted. If we go that way, I'd prefer to address this as a small refactoring in a subsequent PR.

test/elementwise/unary_f32.c
ebavier: `elementwise` seems good to me, then we could conceivably have `binary_f32.c` etc at some point.

test/exp/exp_f32_xsfvfexp32e.c
Aaron-Hutchinson: I think the tests for each RISC-V extension should be kept in separate files, as the CMake infrastructure kind of expects it that way, and it would be more consistent with other test directories. I'm totally fine with reusing code through a shared header, although the driver header is probably not the right place for it.

test/math/unary_f32.h
Aaron-Hutchinson: Sorry I wasn't clear. GEMM uses a, b, c because they are the standard names for that kernel (originating from BLAS I believe). I vaguely recall an internal, informal conversation about parameter names for elementwise functions, which I think proposed to use `in` and `out` in the API, and it seems logical for the test here to match those names.

test/skl-test-driver.h
Aaron-Hutchinson: I'm fine with that. This likely also has crossover with the discussion in #74, where some examples may also want to reuse some of these utility functions. We can probably address all of this at once.

test/math/unary_f32.h
ebavier: We have:
```
  in /  out : exp, logistic, silu
 src /  dst : gelu
pSrc / pDst : cvt
```
so there's not yet a single pair of names that would match all.

Aaron-Hutchinson: Oh. I guess this was some cleanup that we never addressed then. We should probably come up with a standard and enforce it throughout.

test/math/unary_f32.c
hdelassus: I added the tested function name to each test report, so that both tests and benchmarks identify which function was executed.

test/math/unary_f32.h
hdelassus: I used `in` / `out` for now and we'll do a refactoring pass once we have a naming convention.

test/elementwise/unary_f32.c
hdelassus: Renamed it.

test/elementwise/unary_f32.c
```diff
@@ -0,0 +1,94 @@
...
+
+#define INFO(fmt, ...) SKL_TEST_LOG(t, SKL_TEST_LOG_INFO, fmt, __VA_ARGS__)
+
+  INFO("Function: %s\n", h->func_name);
+  if (h->steps.verify) {
+    INFO("N: %zd\n", h->in.len);
+    INFO("Max ulp: %.3f\n", h->ctx.max_err);
+  }
+
+  if (h->steps.warmup) {
```
Aaron-Hutchinson: I guess this condition wrapper is trying to detect whether performance reporting is desired? What if you want to measure performance without cache warming?

Aaron-Hutchinson (dismissed):
The harness structure looks great to me. I defer to @ebavier's review for the exp specifics.

test/elementwise/unary_f32.c
hdelassus: Is there a way in place in the test driver to determine if a test is actually a benchmark?
I don't know if we would actually want to measure performance without a warmup run, but I'm ok with using a better check.

test/CMakeLists.txt
hdelassus: Sorted.

test/elementwise/unary_f32.c
ebavier: How about checking for `h->steps.verify == NULL` or `!= NULL`?  Maybe not intended that way, but I assume most benchmarks will not be verified, whereas all tests will?

hdelassus: LGTM!

Aaron-Hutchinson: No, the driver itself has no concept of test vs benchmark since a single "test" could in theory perform both verification and performance measurement.

In the GEMM harness I just always printed the performance information to avoid this problem altogether. For GEMM we tune around a cold cache, so there it makes sense to not assume a warmup. Not sure if it's different here.

Aaron-Hutchinson (approved):

ebavier (approved):
LGTM now! Things came together pretty well, thanks for working through the feedback!

PR #81, opened 2026-04-21
[GEMM] Add packed f32 GEMM harness

myeh01:
This PR adds a packed f32 GEMM test harness using the new format and converts all packed f32 GEMM tests. The non-packed f32 GEMM harness has been removed, and all tests using it have been updated to use this packed harness (with block size 1x1x1).


test/gemm/gemm_f32rcprc_f32rcprc_f32rcprc.c
```diff
@@ -1,187 +1,80 @@
...
-  SKL_TEST_REQUIRE(status, csc0 == 1);
-  SKL_TEST_REQUIRE(status, alpha == 1.f);
-  SKL_TEST_REQUIRE(status, beta == 0.f || beta == 1.f);
-  if (status) {
-    exit(status);
+void gemm_f32rcprc_f32rcprc_f32rcprc_init(skl_test_t *t) {
+  gemm_f32rcprc_f32rcprc_f32rcprc_t *h =
+      (gemm_f32rcprc_f32rcprc_f32rcprc_t *)t->harness;
+
+  if (h->m1 == 0 || h->k1 == 0) {
```
Aaron-Hutchinson: Should this also check if `m0` and `k0` are 0? Looks like we'll have overflow in the other branch in those cases.

test/gemm/gemm_f32rcprc_f32rcprc_f32rcprc.c
```diff
@@ -199,80 +92,222 @@ int check_error(void) {
...
-      (size_t)CSB1, roundoff_scaling * fabs((double)BETA), bound, RSC0, CSC0,
-      (size_t)RSC1, (size_t)CSC1);
+      h->m0, h->n0, h->k0, h->m1, h->n1, h->k1,
+      roundoff_scaling * fabs((double)h->alpha), h->ctx.a_wide, h->rsa0,
+      h->csa0, h->rsa1, h->csa1, h->ctx.b_wide, h->rsb0, h->csb0, h->rsb1,
+      h->csb1, roundoff_scaling * fabs((double)h->beta), h->ctx.bound, h->rsc0,
+      h->csc0, h->rsc1, h->csc1);
+
+  // Compute the reference result using h->ctx.ref_c
+  // h->ctx.ref_c contains the original C values (copied in skl_test_init)
```
Aaron-Hutchinson, line 170: > h->ctx.ref_c contains the original C values (copied in skl_test_init)

There's some redundancy between this comment and the one above. Should probably omit one of them.

test/gemm/gemm_f32rcprc_f32rcprc_f32rcprc.c
```diff
@@ -199,80 +92,222 @@ int check_error(void) {
...
+  // After this call, h->ctx.ref_c will contain the reference result
+  skl_gemm_f32rcprc_f32rcprc_f32rcprc_ref(
+      h->m0, h->n0, h->k0, h->m1, h->n1, h->k1, h->alpha, h->a_pack.data,
```
Aaron-Hutchinson: The harness pointer dereferences make the whole `verify` function a bit difficult to read. I'm not sure if it's actually an improvement, but one idea is to unpack some/all of the harness parameters at the start (e.g. `size_t m0 = h->m0`) to reduce the clutter. I'll let you decide if this is worth doing.

test/gemm/gemm_f32rcprc_f32rcprc_f32rcprc.c
```diff
@@ -199,80 +92,222 @@ int check_error(void) {
...
-            printf("result [%zu, %zu, %zu, %zu] (%f) != reference (%f)\n", i0,
-                   j0, i1, j1, test_c[idx], ref_c[idx]);
-            return 1;
+              i1 * rsc1_adj + j1 * csc1_adj + i0 * rsc0_amb + j0 * csc0_amb;
+          // NaNs always compare as false, so do a bitwise comparison.
+          int64_t initial = 0;
+          int64_t result = 0;
+          memcpy(&initial, &(h->ctx.c_pack_copy[idx]),
+                 sizeof(*(h->ctx.c_pack_copy)));
+          memcpy(&result, &(h->c_pack.data[idx]), sizeof(*(h->c_pack.data)));
```
Aaron-Hutchinson: Since this logic is reused a few times maybe it's worth extracting it into a helper function:
```
static bool is_bitwise_eq(float x, float y) { ... }
```

test/gemm/rvv/skl_gemm_f32_f32_f32_zve32f_x390.c
```diff
@@ -10,98 +10,111 @@
...
+                                 .test_size =
+                                     sizeof(gemm_f32rcprc_f32rcprc_f32rcprc_t),
                                  .tests = tests};
 
 static void execute(skl_test_t *t) {
-  const gemm_f32rc_f32rc_f32rc_t *h = (gemm_f32rc_f32rc_f32rc_t *)t->harness;
+  const gemm_f32rcprc_f32rcprc_f32rcprc_t *h =
+      (gemm_f32rcprc_f32rcprc_f32rcprc_t *)t->harness;
 
-  SKL_TEST_REQUIRE(t, execute_status, h->csa == 1);
```
myeh01, line 86: Do we want to place these checks in `execute`? I think it's ok for tests, but we probably don't want them in benchmarks.

test/gemm/gemm_f32rcprc_f32rcprc_f32rcprc.c
```diff
@@ -199,80 +92,222 @@ int check_error(void) {
...
+      }
+      if (i1 == h->m1 - 1 && j1 == h->n1 - 1) {
+        bsc1 = c_block_min_len;
+      }
+      for (size_t i = c_block_min_len; i < bsc1; ++i) {
+        size_t idx = i1 * rsc1_adj + j1 * csc1_adj + i;
+        // NaNs always compare as false, so do a bitwise comparison.
+        int64_t initial = 0;
+        int64_t result = 0;
+        memcpy(&initial, &(h->ctx.c_pack_copy[idx]),
```
Aaron-Hutchinson: What's the reason for using a separate array `c_pack_copy` throughout this verification instead of just `ref_c`? Wouldn't the intermediate array elements be the same between these arrays?

test/gemm/gemm_f32rcprc_f32rcprc_f32rcprc.c
```diff
@@ -199,80 +92,222 @@ int check_error(void) {
...
+  // Check each block.
+  for (size_t i1 = 0; i1 < h->m1; ++i1) {
+    for (size_t j1 = 0; j1 < h->n1; ++j1) {
+      // Check the first c_block_min_len elements.
+      for (size_t i0 = 0; i0 < m0_amb; ++i0) {
+        for (size_t j0 = 0; j0 < n0_amb; ++j0) {
+          if (layoutc0_row_major && j0 % csc0_adj == 0 &&
+              j0 < h->n0 * csc0_adj) {
+            continue;
+          }
```
Aaron-Hutchinson: If I'm understanding right this is here to skip over the array entries corresponding to the actual matrix that are supposed to change as part of the kernel computation. The last condition `j0 < h->n0 * csc0_adj` looks like it's ensuring this check only applies within the bounds corresponding to the matrix (non-leading dimension values). In that case should `n0` be `n0 - 1` because `(n0 - 1) * csc0` would be the index of the final value in this row?

test/gemm/gemm_f32rcprc_f32rcprc_f32rcprc.c
```diff
@@ -199,80 +92,222 @@ int check_error(void) {
...
+                         h->ctx.bound[idx]);
+            t->status.verify_status = SKL_TEST_FAIL;
+            return;
+          }
+        }
+      }
+    }
+  }
+
+  /* Check for clobbered elements. */
```
Aaron-Hutchinson, line 194: All of the logic for checking for clobbered elements should depend only on the size of the datatype and not the datatype itself. Due to this, how feasible do you think it is to generalize this code and extract it into a separate shared file? That way we only need to write and maintain one version of it.

test/gemm/rvv/skl_gemm_f32_f32_f32_zve32f_x390.c
Aaron-Hutchinson: `init` probably makes more sense.

Aaron-Hutchinson: Oh but `init` is shared. In that case we either put them in `main`, which would have to error before calling the into driver, or have each test file define its own `init` wrapper that calls into the main harness `init`.

test/gemm/gemm_f32rcprc_f32rcprc_f32rcprc.c
myeh01: They should be, but I thought it would be better to compare to an untouched copy of the original data. Don't really have a strong opinion either way.

test/gemm/gemm_f32rcprc_f32rcprc_f32rcprc.c
myeh01: The matrix entries correspond to `j0 * csc0` for `j0 = 0, 1, ..., n0 - 1`, right? So I think it's correct as is.

test/gemm/gemm_f32rcprc_f32rcprc_f32rcprc.c
myeh01: I tried writing it in a way so that it could just be copy and pasted into other test harnesses. It just assumes the datatype is <=64 bits (for the memcpy, which copies into `int64_t`). Besides that, I think only the error message printf format specifiers potentially need to be changed. The memcpy is redundant for integer types, but it'll still work correctly.

test/gemm/gemm_f32rcprc_f32rcprc_f32rcprc.c
myeh01: I'll check these with `SKL_TEST_REQUIRE`s (wherever they end up going) because I don't think anyone should be using blocks with a zero dimension.

test/gemm/gemm_f32rcprc_f32rcprc_f32rcprc.c
Aaron-Hutchinson: I agree it should work just fine with minimal adjustments. I'm more concerned about the large amount of code duplication among all the coming harnesses, especially involving pretty complicated logic.

myeh01: Oh, sorry, did not read your original comment carefully. Let me see what I can do.

myeh01: I'm not seeing a good way to make this into a single function. If we remove the error print out, then I think it's possible by casting the arrays to `char *` and adjusting the strides accordingly. But, having the print out is helpful for debugging. Another option is to make this into a macro, or just leave it as is.

test/gemm/rvv/skl_gemm_f32_f32_f32_zve32f_x390.c
myeh01: I'm not sure that either is great. If we put it in `init`, we'll need to reduplicate all the array initialization logic in each test suite. If we put it `main`, we'll need to create a `skl_test_t` struct for each test instance to pass to `SKL_TEST_REQUIRE`. The latter will update the status field of each struct, but these structs will be separate from the ones created by `skl_test_driver_run_suite`. I don't know if we want to add another parameter check step to `skl_test_steps_t`.

test/gemm/gemm_f32rcprc_f32rcprc_f32rcprc.c
Aaron-Hutchinson: Regarding the error message, it's just the format string that's causing an issue right? Can we just print the raw value as an int in hex? I'm also not sure it's super useful to know the actual value being clobbered, so maybe just printing the index and omitting the value altogether would be fine.

test/gemm/rvv/skl_gemm_f32_f32_f32_zve32f_x390.c
Aaron-Hutchinson: Sorry I wasn't very clear. I meant to do something like:
```
static void init(skl_test_t *t) {
  SKL_TEST_REQUIRE(t, init_status, h->csa1 == 1);
  SKL_TEST_REQUIRE(t, init_status, h->csb1 == 1);
  SKL_TEST_REQUIRE(t, init_status, h->csc1 == 1);
  if(t->status.init_status != SKL_TEST_PASS)
    return;
  gemm_f32rcprc_f32rcprc_f32rcprc_init(t);  
}
```
with `steps.init = init`. Would this work?

myeh01: Oh, right. Yes this looks great.

test/gemm/gemm_f32rcprc_f32rcprc_f32rcprc.c
myeh01: Yeah, maybe the index is enough.

test/gemm/skl_test_gemm.h
Aaron-Hutchinson, line 1: Nit: the other gemm filenames use underscore separators, so should this one also? I guess we've been somewhat inconsistent with this overall.

test/gemm/xsfmm/skl_gemm_a1b01_f32pc_f32cp_f32rcp_xsfmm32a32f.c
```diff
@@ -0,0 +1,222 @@
...
+gemm_f32rcprc_f32rcprc_f32rcprc_t tests[] = {
+#ifdef SKL_ENABLE_BENCHMARKS
+    // Benchmark tests
+    {BENCH, .m1 = 2, .n1 = 2, .k1 = 2048, .alpha = 1.f, .beta = 0.f},
+    {BENCH, .m1 = 2, .n1 = 2, .k1 = 2048, .alpha = 1.f, .beta = 1.f},
+#endif // SKL_ENABLE_BENCHMARKS
+
+#ifdef SKL_ENABLE_TESTS
+    // Verification tests - comprehensive coverage for Xsfmm A1B01
+    {TEST, .m1 = 7, .n1 = 1, .k1 = 0, .alpha = 1.f, .beta = 0.f},
```
Aaron-Hutchinson, line 68: `alpha = 1.f` is a strict requirement for this kernel right? We could move that assignment into the `TEST` and `BENCH` definitions to simplify this list a bit. Same for the other xsfmm file.

test/gemm/skl_test_gemm.h
```diff
@@ -0,0 +1,213 @@
...
+  if (m0 > 1 && n0 > 1) {
+    if (rs0 >= cs0) {
+      SKL_TEST_REQUIRE(t, init_status, rs0 >= (n0 - 1) * cs0 + 1);
+    } else {
+      SKL_TEST_REQUIRE(t, init_status, cs0 >= (m0 - 1) * rs0 + 1);
+    }
+  }
+
+  size_t block_min_len = (m0 - 1) * rs0 + (n0 - 1) * cs0 + 1;
+  if (m1 > 1 && n1 > 0) {
```
Aaron-Hutchinson, line 67: Is there a situation where `n1 == 0` makes sense? I'm confused why the `n1 > 0` check is here, or otherwise why there isn't a `SKL_TEST_REQUIRE(t, init_status, n1 > 0)`.

test/gemm/skl-test-gemm.h
```diff
@@ -0,0 +1,213 @@
...
+ * @param n1 - Number of block-columns in C_pack and ref_c.
+ * @param c_pack - Pointer to matrix C_pack.
+ * @param rsc0 - Row stride within each block of C_pack and ref_c in elements.
+ * @param csc0 - Column stride within each block of C_pack and ref_c in
+ * elements.
+ * @param rsc1 - Row stride between blocks of C_pack and ref_c in elements.
+ * @param csc1 - Column stride between blocks of C_pack and ref_c in elements.
+ * @param ref_c - Pointer to matrix ref_c.
+ *
+ * C_pack and ref_c are matrices stored in arrays with c_pack_len elements of
```
Aaron-Hutchinson: In the context of this function I'm not sure there's a need to distinguish between "result" and "reference". It might simplify the exposition and variable naming to just call them e.g. `x` and `y`.

test/gemm/xsfmm/skl_gemm_a1b01_f32pc_f32cp_f32rcp_xsfmm32a32f.c
```diff
@@ -0,0 +1,222 @@
...
+ */
+
+#define TEST                                                                   \
+  GEMM_F32RCPRC_F32RCPRC_F32RCPRC_DEFAULTS,                                    \
+      .steps = {                                                               \
+          .init = init,                                                        \
+          .warmup = NULL,                                                      \
+          .execute = execute,                                                  \
+          .verify = gemm_f32rcprc_f32rcprc_f32rcprc_verify,                    \
+          .report = NULL,                                                      \
```
Aaron-Hutchinson: I know this wasn't due to this PR but can we call report for tests too? When a test fails none of the parameters in the `INFO` lines are being printed. Maybe we can just omit the instructions, cycles, and other performance info when `verify != NULL`.

Aaron-Hutchinson (commented):
Thanks for the updates! I think this is really close to ready.

test/gemm/skl_test_gemm.h
myeh01: Sure, I can change it. I used hyphens because `skl-test-driver.{c,h}` use it, but I'm fine either way.

test/gemm/skl_test_gemm.h
myeh01: While having `m0`, `n0`, or `k0` be zero makes no sense, `m1`, `n1` and `k1` should be allowed to be zero. This is analogous to having `m`, `n`, or `k` be zero in a normal GEMM. If `m1 == 0` or `n1 == 0`, then `rs1` and `cs1` can be arbitrary, so we can only impose a requirement on these strides when `m1 > 0` and `n1 > 0`.

test/gemm/gemm_f32rcprc_f32rcprc_f32rcprc.c
```diff
@@ -199,80 +144,97 @@ int check_error(void) {
...
-                    (size_t)CSB1, BETA, c, RSC0, CSC0, (size_t)RSC1,
-                    (size_t)CSC1);
+void gemm_f32rcprc_f32rcprc_f32rcprc_report(skl_test_t *t) {
+  gemm_f32rcprc_f32rcprc_f32rcprc_t *h =
+      (gemm_f32rcprc_f32rcprc_f32rcprc_t *)t->harness;
+
+  size_t maccs = h->m1 * h->n1 * h->k1 * h->m0 * h->n0 * h->k0;
+  float mpc = (float)maccs / (float)t->counters.cycles;
+
+#define INFO(fmt, ...) SKL_TEST_LOG(t, SKL_TEST_LOG_INFO, fmt, __VA_ARGS__)
```
Aaron-Hutchinson: Related to my other comment, maybe most of this printing can be factored out into a shared function(s) now that we have `skl-test-gemm.h`. Maybe one for printing the dimension parameters, one for the performance info, and leave type-dependent fields (alpha, beta, ...) up to individual harnesses. Just brainstorming, but this would reduce code duplication more.

test/gemm/xsfmm/skl_gemm_a1b01_f32pc_f32cp_f32rcp_xsfmm32a32f.c
myeh01: What if we had different report functions for tests and benchmarks? One would just print out information on the input matrices and other parameters like alpha, beta, etc., and the other would additionally print out performance data. Though maybe not a common use case, we could still allow users to verify benchmark results and print out performance information.

Aaron-Hutchinson: Sounds good to me

test/gemm/xsfmm/skl_gemm_a1b01_f32pc_f32cp_f32rcp_xsfmm32a32f.c
myeh01: We will add support for general alpha and beta in this kernel soon, so I will leave it as is.

test/gemm/skl_test_gemm.h
myeh01: Changed to `skl_test_gemm.h`.

test/gemm/skl-test-gemm.h
myeh01: Renamed these to `c_pack0` and `c_pack1`.

test/gemm/xsfmm/skl_gemm_a1b01_f32pc_f32cp_f32rcp_xsfmm32a32f.c
myeh01: Ok, I've updated the printing functions and put some of this code in `skl_test_gemm.h`. Let me know what you think.

test/gemm/skl_test_gemm.h
```diff
@@ -0,0 +1,246 @@
...
+               csb1);
+  SKL_TEST_LOG(t, SKL_TEST_LOG_INFO,
+               "RSC0: %zd, CSC0: %zd, RSC1: %zd, CSC1: %zd\n", rsc0, csc0, rsc1,
+               csc1);
+}
+
+static inline void gemm_rcprc_rcprc_rcprc_report_perf(skl_test_t *t,
+                                                      bool warmup, size_t maccs,
+                                                      size_t cycles,
+                                                      size_t instret) {
```
Aaron-Hutchinson: I'd prefer this just accepted the original `counters` object so that the signature and call sites don't have to be updated if additional performance information is added to `skl_test_counters_t`.

test/gemm/skl_test_gemm.h
```diff
@@ -0,0 +1,246 @@
...
+    memcpy(&element1, &(c_pack1_cast[c_type_size * idx]), c_type_size);
+    if (element0 != element1) {
+      SKL_TEST_LOG(t, SKL_TEST_LOG_ERROR, "element [%zu] clobbered\n", idx);
+      t->status.verify_status = SKL_TEST_FAIL;
+      return;
+    }
+  }
+}
+
+static inline void gemm_rcprc_rcprc_rcprc_report_matrix_params(
```
Aaron-Hutchinson, line 247: Missing docstring here and on the other report function.

test/gemm/skl_test_gemm.h
```diff
@@ -0,0 +1,246 @@
+// Copyright 2026 SiFive, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
```
Aaron-Hutchinson, line 3: Should probably have a `@brief` docstring for this file. Probably doesn't need to be long.

Aaron-Hutchinson (commented):
A few more small requests and I think this is good to go!

myeh01:
@Aaron-Hutchinson Also, are we spelling it "MACs" or "MACCs"? I thought it was the latter, but in the original harness, the variable is called `maccs` while the output is `MACs/Cycle`.

Aaron-Hutchinson:
I don't think it matters here. You can pick one and make them consistent.

test/gemm/skl_test_gemm.h
```diff
@@ -0,0 +1,282 @@
...
+               csc1);
+}
+
+/**
+ * @brief Print out performance information for a packed GEMM test.
+ *
+ * @param t - Test context.
+ * @param warmup - Pointer to the test's warmup function.
+ * @param maccs - Number of multiply-accumulate operations performed.
+ * @param counters - The test's `counters` object.
```
Aaron-Hutchinson: Actually as this comment points out isn't `counters` accessible as just `t->counters`? Maybe it doesn't need to be a parameter at all.

I think `warmup` is ok as a separate parameter to avoid the dependence on the data layout since the harness type is unknown here.

Aaron-Hutchinson (approved):
Look great! Thanks for pushing this through!

PR #83, opened 2026-04-23
[Test] Ports softmax_f32 and softmax_2d_f32 to new driver.

ebavier:
This adds a new test harness for Softmax and new tests for softmax_f32 and softmax_2d_f32 using that harness.

* test/softmax/softmax_f32.h: New file; softmax harness declarations.
* test/softmax/softmax_f32.c: Replace old tester with implementation for new harness.
* test/softmax/skl_softmax_f32.c, test/softmax/skl_softmax_2d_f32.c: New tests using skl-test-driver and softmax harness.
* test/CMakeLists.txt (skl_softmax_f32, skl_softmax_2d_f32): Declare new tests.
* test/softmax/softmax_2d_f32.c: Delete old 2d test.
* CMakeLists.txt (softmax_f32, softmax_2d_f32): Remove legacy tests.


test/softmax/skl_softmax_2d_f32.c
```diff
@@ -0,0 +1,105 @@
...
+#ifdef SKL_ENABLE_BENCHMARKS
+#ifdef __riscv_zve32f
+  VARIANT_BENCHMARKS(zve32f),
+#endif
+#ifdef __riscv_xsfvfexpa
+  VARIANT_BENCHMARKS(xsfvfexpa),
+#endif
+#ifdef __riscv_xsfvfexp32e
+  VARIANT_BENCHMARKS(xsfvfexp32e),
+#endif
```
Aaron-Hutchinson: My preference would be to split the tests for different extensions into separate files. That way every file in `src` has an associated file in `test`, which makes it very clear to users where a given kernel is being tested. This would also eliminate the awkward file naming of `softmax_f32.c` vs `skl_softmax_f32.c`, and allow skipping the compilation of this test when none of these extensions are present.

test/softmax/skl_softmax_2d_f32.c
```diff
@@ -0,0 +1,105 @@
...
+  skl_softmax_2d_f32_t fn = (skl_softmax_2d_f32_t)(h->func);
+  fn(h->ctx.s, h->rss, h->a.data, h->rsa, h->beta, h->m, h->n);
+}
+
+int main(void) {
+  // Set execution step
+  for (size_t i = 0; i < suite.num_tests; i++) {
+    tests[i].steps.execute = execute;
+    if (tests[i].steps.verify == NULL)
+      tests[i].steps.warmup = execute;
```
Aaron-Hutchinson: What's the reason to set these here instead of including them as part of the above macro definitions?

test/softmax/softmax_f32.c
```diff
@@ -1,44 +1,38 @@
...
+  h->ctx.s = slen ? malloc(slen * sizeof(float)) : NULL;
+
+  if (h->steps.verify) {
+    // Allocate reference buffer
+    h->ctx.S = h->n ? malloc(h->n * sizeof(double)) : NULL;
+  }
+}
+
+/** Reference softmax producing FP64 results for high accuracy */
+static void softmax_f32_f64(double *out, const float *in, float beta,
```
Aaron-Hutchinson, line 57: Should this just be added to `skl-ref`?

test/softmax/softmax_f32.h
```diff
@@ -0,0 +1,47 @@
...
+
+#include "skl-test-driver.h"
+#include <stddef.h>
+
+typedef struct {
+  // Test function pointers for various steps
+  // *** This field must be placed first within this struct ***
+  skl_test_steps_t steps;
+
+  // Configurable parameter (Softmax function and arguments)
```
hdelassus:
```suggestion
  // Configurable parameters (Softmax function and arguments)
```

test/softmax/softmax_f32.c
```diff
@@ -1,44 +1,38 @@
...
+    h->a.len = 0;
+  } else {
+    h->a.len = (h->m - 1) * h->rsa + h->n;
+    slen = (h->m - 1) * h->rss + h->n;
+  }
+
+  // Allocate and initialize input buffer
+  SKL_TEST_BUF_CREATE(t, float, &h->a);
+  // Allocate output buffer, avoiding malloc(0)
+  h->ctx.s = slen ? malloc(slen * sizeof(float)) : NULL;
```
hdelassus, line 30: Seeing this was used in other harnesses, I was wondering why we would want to avoid `malloc(0)`. What is the problem?

test/softmax/softmax_f32.c
```diff
@@ -49,54 +43,101 @@ static void reference_softmax_f32(float *out, const float *in, float beta,
...
+}
+
+void softmax_f32_verify(skl_test_t *t) {
+  softmax_f32_t *h = (softmax_f32_t *)t->harness;
+  // NOLINTBEGIN(*-isolate-declaration)
+  const size_t M = h->m, N = h->n;
+  const float min = h->a.min, max = h->a.max;
+  const float beta = h->beta;
+  size_t errs = 0;
+  // NOLINTEND(*-isolate-declaration)
```
hdelassus: Should we remove `*-isolate-declaration` from our linter configuration? 😅

test/softmax/softmax_f32.c
```diff
@@ -49,54 +43,101 @@ static void reference_softmax_f32(float *out, const float *in, float beta,
...
+
+  // Expected error is error of an N-element summation, i.e. `N u`,
+  // plus error introduced by stabilization and beta-scaling.  The
+  // latter subtraction and multiplication would alone introduce just
+  // `3 u` error, but this is inflated to `3β(min-max)` when passed
+  // through the exponential.
+  double u = 0x1p-24; // "unit round-off"
+  double tol = (3 * fmax(beta * fabsf(max - min), 1) + (double)N) * u;
+  tol = tol / (1 + tol); // relative to reference
```
hdelassus, line 92: Why `tol / (1 + tol)`?

test/softmax/softmax_f32.c
```diff
@@ -1,44 +1,38 @@
...
-#if !defined(BETA)
-#define BETA 1.0 // Exponential scaling factor
-#endif
-
-#include "skl-ref.h"
-#include "skl-test.h"
-#include "skl.h"
-#include <inttypes.h>
-#if defined(ENABLE_TEST)
+#include "softmax/softmax_f32.h"
```
hdelassus, line 6: Missing copyright?

test/softmax/skl_softmax_f32.c
```diff
@@ -0,0 +1,91 @@
+#include "skl-ref.h" // NOLINT(misc-include-cleaner)
```
hdelassus: ditto.

test/softmax/skl_softmax_2d_f32.c
```diff
@@ -0,0 +1,105 @@
+#include "skl-ref.h" // NOLINT(misc-include-cleaner)
```
hdelassus: ditto.

test/softmax/skl_softmax_2d_f32.c
ebavier: `softmax_f32.c` is the test harness for both 1D and 2D Softmax functions.  `skl_softmax_f32.c` is the tester (? do we have a name for the test-driver and harness client?) for 1D Softmax.

I'd prefer to keep everything together to avoid having to duplicate both harness infrastructure as well as test-case generation macros many times over.

In the case of none of the vector extensions being present, tests will be run for the (scalar) reference function, which IMO is useful given issues we've had in the past.

hdelassus (commented):
Thanks for working on porting these. Some early feedback...

test/softmax/skl_softmax_2d_f32.c
ebavier: Right, they probably could be set in the macros.

test/softmax/softmax_f32.c
ebavier: Likely not, imo, since it does not correspond to any kernel proper.

test/softmax/softmax_f32.c
ebavier: At the very least, clang's static analyzer will issue an error about "potential malloc(0)".  Though, `malloc(0)` is properly defined by POSIX, but its possible there are some bugs in the wild?

test/softmax/softmax_f32.c
ebavier: Maybe!

test/softmax/skl_softmax_2d_f32.c
Aaron-Hutchinson: I don't think it's necessary to duplicate the harness, similar to reusing the `unary_f32.c` harness introduced in #80 for the various exp functions. I'm only requesting to split the "[test suite](https://github.com/sifive/skl/blob/main/test/README.md#test-driver-framework)" so that it roughly matches the `src` directory, as AFAIK this is what other PRs have done and should provide familiarity to users.

As for the macros, maybe they can be placed somewhere that they can be shared among the test suites.

ebavier: Do we have a home for tests of reference functions?

Aaron-Hutchinson: I don't think so, as this is the first time I've seen this come up since the test refactor began. I would say put it in the same directory, maybe as `softmax_f32_ref.c` as if "ref" were the extension.

Sorry for being out of the loop on this. What's the reason to test the reference function? If it needs to be tested, doesn't that defeat the purpose of it being a reference?

ebavier: Because untested code is the exact sort of place a stupid bug would tend to lurk.

The situation is a bit different here from e.g. `exp`, where "reference" is understood in the sense of "we measure correctness relative to the reference kernel," and is more like "the semantics are to be understood as outlined in the reference".  I think both are valid form of reference.  

In `exp` the reference function is tested by cross-reference with the vector kernels (and currently also by virtue of "external" testing, the `expf` either coming from glibc or newlib).  In Softmax tests, I'd rather not use the reference kernel as a ground-truth reference point, since the accuracy is rather worse than the vector kernels, which would put us in danger of relaxing error bounds beyond usefulness.  So, the reference function needs another form of more direct testing.

test/softmax/softmax_f32.c
hdelassus: Yeah I think it's also properly defined in C99: it returns either `NULL` or a value that can be safely passed to `free`. Fair point about clang-tidy though... we could disable the [checker](https://clang.llvm.org/docs/analyzer/checkers.html#optin-portability-unixapi) as well.

test/softmax/softmax_f32.c
ebavier: The initial calculation of `tol` based on `beta`, `min`, `max`, and `N` should be properly understood as a absolute error tolerance, but since this later calculates per-element _relative_ error, we need this `tol / (1 + tol)`.  In most cases this doesn't change much but does make the tolerance tighter.

hdelassus:
Noticed a typo in the PR title: softmax_3d -> 2d 😁

test/CMakeLists.txt
```diff
@@ -157,3 +157,59 @@ skl_add_test(
   "zve32x"
   ""
 )
+skl_add_test(
```
Aaron-Hutchinson, line 237: Could we put these above the transpose tests to keep them roughly alphabetical?

test/softmax/softmax_f32.c
```diff
@@ -54,54 +66,99 @@ static void reference_softmax_f32(float *out, const float *in, float beta,
...
+  softmax_f32_t *h = (softmax_f32_t *)t->harness;
+
+#define INFO(fmt, ...) SKL_TEST_LOG(t, SKL_TEST_LOG_INFO, fmt, __VA_ARGS__)
+
+  INFO("Function: %s\n", h->name);
+  INFO("M: %zd, N: %zd\n", h->m, h->n);
+  if (h->rss != h->n || h->rsa != h->n)
+    INFO("RSS: %zd, RSA: %zd\n", h->rss, h->rsa);
+  INFO("Beta: %g\n", h->beta);
+  if (h->steps.verify != NULL) {
```
Aaron-Hutchinson, line 145: In case it appeals to you, an alternative @myeh01 found for GEMM is to use two separate report functions for tests and benchmarks, and specify the appropriate one in `steps`: 
https://github.com/sifive/skl/blob/c282ee944ea94178c5472af812d90970c42ffd8f/test/gemm/gemm_f32rcprc_f32rcprc_f32rcprc.c#L202-L222

Aaron-Hutchinson (commented):
Thanks for refactoring this to be better aligned with the overall structure! I'm satisfied with the big picture and will defer to @hdelassus to review the finer softmax details.

test/softmax/softmax_f32.h
```diff
@@ -0,0 +1,121 @@
...
+  {__VA_ARGS__, .beta = 1.0f},                                                 \
+  {__VA_ARGS__, .beta = 0x1.0a2b24p0f}
+
+#define DOMAIN_TESTS(...)                                                      \
+  BETA_TESTS(__VA_ARGS__, .a.min =  -1, .a.max =   +1),                        \
+  BETA_TESTS(__VA_ARGS__, .a.min = +40, .a.max = +100)
+
+#define ALIGN(N) (((N) / 128) * 128 + 128)
+
+#define NAT_TESTS(N,...)                                                       \
```
hdelassus, line 86: What does `NAT` mean or stand for?

test/softmax/softmax_f32.h
```diff
@@ -0,0 +1,121 @@
...
+      .report = softmax_f32_report,                                            \
+      .cleanup = softmax_f32_cleanup,                                          \
+  }
+
+#define TEST                                                                   \
+  .a.mode = SKL_TEST_RANDOM, BASIC_STEPS, .steps.verify = softmax_f32_verify
+
+#define BENCHMARK .a.mode = SKL_TEST_SEQ, BASIC_STEPS, .steps.verify = NULL
+
+// clang-format off
```
hdelassus, line 52: Why this? Now there's a mix of no whitespace after commas as well as whitespace after commas.

test/softmax/skl_softmax_f32_zve32f.c
```diff
@@ -0,0 +1,37 @@
...
+
+#include "skl-test-driver.h"
+#include "skl.h"
+#include "softmax/softmax_f32.h"
+
+/**
+ * @brief Test cases for 1D Softmax with Zve32f extension
+ */
+
+// clang-format off
```
hdelassus, line 18: Was clang-format doing something ugly in these instances?

hdelassus (approved):
LGTM, just cosmetic/questions

test/softmax/skl_softmax_f32_zve32f.c
ebavier: It was pushing the `VARIANT_...` bits far to the right to align with the opening `{`, which I thought made it visually difficult to track the `#ifdef`s etc.

test/softmax/softmax_f32.h
ebavier: Otherwise `clang-format` wants to put the member assignments on the same line, which I thought made it difficult to quickly see which things were being set by each macro.

With the leading double `__` I didn't catch the missing spaces. 😺 . Should we bother fixing?

test/softmax/softmax_f32.h
ebavier: Not `Nat` in the functional language "natural-number" sense 😸 .  But "natural" in the sense of "not-aligned", as in the other case.

test/softmax/softmax_f32.h
hdelassus: I'll leave it up to you. :)

PR #85, opened 2026-04-28
[Logistic] Adds Xsfvfexp32e kernels.

ebavier:
Inline assembly avoids pessimistic register spills from compilers, lowers memory traffic, and improves performance.

* src/logistic/logistic_f32_xsfvfexp32e.h, src/logistic/logistic_f32_xsfvfexp32e.c: New files
* include/skl.h [__riscv_xsfvfexp32e]: Include header.
* CMakeLists.txt: Add src.
* test/logistic/logistic_f32.c [RUN_XSFVFEXP32E]: Test it.

### DIFF
--- a/src/logistic/logistic_f32_xsfvfexp32e.c
+++ b/src/logistic/logistic_f32_xsfvfexp32e.c
@@ -25,7 +25,7 @@ SKL_FUNC void skl_logistic_5u_f32_xsfvfexp32e(float *out, const float *in,
       "\t   vle32.v     v8,  (%[in])     \n"
       "\t   vfsgnj.vf   v8,  v8, %[b]    \n"
       "\t   vfmax.vf    v8,  v8, %[b]    \n"
-      /* 1. Split x = a + e */
+      /* 1. Split x = a - e */
       "\t   vfmul.vf   v16,  v8, %[S]    \n"
       "\t   vfsub.vv   v24, v16, v8      \n"
       /* 2. Compute o = exp(a) and p = exp(e) */
--- a/src/logistic/logistic_f32_xsfvfexp32e.c
+++ b/src/logistic/logistic_f32_xsfvfexp32e.c
@@ -47,9 +47,9 @@ SKL_FUNC void skl_logistic_5u_f32_xsfvfexp32e(float *out, const float *in,
       "\t   vfmul.vv   %[y], %[y], %[z]       \n"
       "\t   vse32.v    %[y], (%[out])"
       : [vl] "=&r"(vl),
-	[w] "=vr"(w), [x] "=vd"(x), [y] "=vd"(y), [z] "=vd"(z)
+        [w] "=vr"(w), [x] "=vd"(x), [y] "=vd"(y), [z] "=vd"(z)
       : [n] "r"(n), [in] "r"(in), [out] "r"(out),
-	[b] "f"(b), [S] "f"(S), [one] "f"(1.0f)
+        [b] "f"(b), [S] "f"(S), [one] "f"(1.0f)
       : "vtype", "vl", "memory");
   // clang-format on
 }
--- a/src/logistic/logistic_f32_xsfvfexp32e.c
+++ b/src/logistic/logistic_f32_xsfvfexp32e.c
@@ -31,7 +31,7 @@ SKL_FUNC void skl_logistic_5u_f32_xsfvfexp32e(float *out, const float *in,
       /* 2. Compute o = exp(a) and p = exp(e) */
       "\t   sf.vfexp.v v16, v16          \n"
       "\t   sf.vfexp.v v24, v24          \n"
-      "\t   vmsgt.vi    v0,  v8, 0       \n"
+      "\t   vmslt.vi    v0,  v8, 0       \n"
       "\t   vfadd.vv    v8, v16, v24     \n"
       /* 3. Assemble numerator: n = x>0 ? p : o */
       "\t   vmerge.vvm v16, v16, v24, v0 \n"
--- a/src/logistic/logistic_f32_xsfvfexp32e.c
+++ b/src/logistic/logistic_f32_xsfvfexp32e.c
@@ -1,0 +1,53 @@
+// Copyright 2026 SiFive, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+#if !defined(__riscv_xsfvfexpa)
+#error This file requires the Xsfvfexpa extension
+#endif
+
+#include "skl-common.h"
+#include <riscv_vector.h>
+#include <sifive_vector.h>
+#include <stddef.h>
+
+// NOLINTNEXTLINE(readability-non-const-parameter)
+SKL_FUNC void skl_logistic_5u_f32_xsfvfexp32e(float *out, const float *in,
+                                              size_t N) {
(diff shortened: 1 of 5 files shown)


src/logistic/logistic_f32_xsfvfexp32e.c
```diff
@@ -0,0 +1,53 @@
+// Copyright 2026 SiFive, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+#if !defined(__riscv_xsfvfexpa)
+#error This file requires the Xsfvfexpa extension
+#endif
```
hdelassus: Looks like the wrong extension

src/logistic/logistic_f32_xsfvfexp32e.c
```diff
@@ -0,0 +1,53 @@
...
+
+#include "skl-common.h"
+#include <riscv_vector.h>
+#include <sifive_vector.h>
+#include <stddef.h>
+
+// NOLINTNEXTLINE(readability-non-const-parameter)
+SKL_FUNC void skl_logistic_5u_f32_xsfvfexp32e(float *out, const float *in,
+                                              size_t N) {
```
hdelassus: I believe `N` is lowercased in pretty much every other kernel?

src/logistic/logistic_f32_xsfvfexp32e.c
```diff
@@ -0,0 +1,53 @@
...
+      /* 4. Approximate 1 / (o + p) */
+      "\t   vfrec7.v   v24, v8 \n"
+      "\t   vfnmsac.vv  v0, v8, v24      \n"
+      "\t   vfmadd.vv  v24, v0, v24      \n"
+      "\t   vfmul.vv    v0, v0, v0       \n"
+      "\t   vfmadd.vv  v24, v0, v24      \n"
+      /* 5. Finally, n * r ~ n / d */
+      "\t   vfmul.vv   v16, v16, v24     \n"
+      "\t   vse32.v    v16, (%[out])"
+      : [vl] "=r"(vl)
```
hdelassus: Nit: I think `vl` should have an early-clobber constraint, do you agree?

hdelassus (dismissed):
Thanks for adding this. I guess I'm fine with explicit register allocation in inline assembly, especially if the compiler couldn't figure it out.

src/logistic/logistic_f32_xsfvfexp32e.c
ebavier: Yes 👍 .  I think I capitalized it when I was trying to get an intrinsics version to work, where I wanted to use `n` for the numerator variable. 😸 I'll lower it.

src/logistic/logistic_f32_xsfvfexp32e.c
ebavier: Maybe?  This would be to prevent the compiler from putting `vl` in the same register as one of the other inputs?  In this case, I would think the compiler would notice that the inputs are used in every loop trip and would avoid materializing them every time, possibly into a register shared with `vl`.  I guess it's possible though.  I'll add the `&`.

hdelassus: Yeah, I don't think the compiler would actually clobber another xreg, but it's more a matter of not leaving UB unchecked.

Now that I'm thinking about it, this snippet would also need to mark all the explicitly-allocated & written regs as clobbers (v0, v8, ...).

ebavier: Is marking vector registers necessary if there are no other vector uses in the function though?  The ABI assumes all vector registers are clobbered at function boundaries?

hdelassus: Actually, v0, v1, v2, ..., v32. 🫠

hdelassus: > Is marking vector registers necessary if there are no other vector uses in the function though? The ABI assumes all vector registers are clobbered at function boundaries?

Sounds logical. Does the compiler have the same constraint when it inlines a function?

hdelassus (dismissed):

src/logistic/logistic_f32_xsfvfexp32e.c
topperc: > Sounds logical. Does the compiler have the same constraint when it inlines a function?

No. If the function gets inlined we could keep vector registers live across it's body.

topperc: Also interprocedural register allocation can inspect the callee to see if the vector registers are really clobbered. If they aren't, the caller could hold a vector register across the call.

ebavier: Ok, so my implementation here could cause problems if inlined, which is a thing we want to be able to do.

ebavier: What I've now done here is to declare 4 `vfloat32m8_t` variables that are used as output variables of the asm block.  This seems to allow the compiler to notice the used registers.

src/logistic/logistic_f32_xsfvfexp32e.c
```diff
@@ -0,0 +1,55 @@
...
+      "\t   vfrec7.v   %[z], %[x] \n"
+      "\t   vfnmsac.vv %[w], %[x], %[z]       \n"
+      "\t   vfmadd.vv  %[z], %[w], %[z]       \n"
+      "\t   vfmul.vv   %[w], %[w], %[w]       \n"
+      "\t   vfmadd.vv  %[z], %[w], %[z]       \n"
+      /* 5. Finally, n / d ~ n * r */
+      "\t   vfmul.vv   %[y], %[y], %[z]       \n"
+      "\t   vse32.v    %[y], (%[out])"
+      : [vl] "=&r"(vl),
+        [w] "=vr"(w), [x] "=vd"(x), [y] "=vd"(y), [z] "=vd"(z)
```
hdelassus, line 50: All of these should be marked early-clobbers too, right?

hdelassus: Also `+` instead of `=`?

hdelassus: Ah no, I believe `=` is fine since their values aren't read from outside the block.

ebavier: I don't think so? Since there are no input operands that are vector registers?  Unless early-clobber is required to ensure the compiler actually reserves independent register groups for each declared variable?  From my experiments, at least clang does this already, e.g. attempting to pass another `vfloat32m8_t` "vd" or "vr" emits a "cannot allocate registers" error.

hdelassus: Well, there _are_ vector registers that are input operands in the asm template, right?
It's true that the VRF is fully allocated, but I am not sure that this doesn't make the early clobber constraint unnecessary for correctness?
@topperc We probably need your advice again 😅

hdelassus: On second thought, I think you're right: there is no vreg that is an input (from outside the block) to the asm statement.

hdelassus (approved):
LGTM!

PR #88, opened 2026-04-30
[NFC] Update the License to MIT

pvchupin:
Updating the whole project codebase to the MIT license before public release. This is the ultimate decision on the project license. While git history may show this project being under Apache 2 license that we considered previously it was never released under it.

* Change LICENSE
* Update CONTRIBUTING doc
* Update all source headers
* Update years to range
* Added header to makefiles

### DIFF
--- a/example/conv2d/conv2d_nhwc_f32.c
+++ b/example/conv2d/conv2d_nhwc_f32.c
@@ -1,4 +1,4 @@
-// Copyright (c) 2026-Present SiFive, Inc. All rights reserved.
+// Copyright (c) 2026 SiFive, Inc. All rights reserved.
 // Licensed under the MIT License.
 // See LICENSE file in the project root for full license information.
 // SPDX-License-Identifier: MIT
--- a/example/conv2d/conv2d_nhwc_f32.c
+++ b/example/conv2d/conv2d_nhwc_f32.c
@@ -1,5 +1,6 @@
 // Copyright (c) 2026 SiFive, Inc. All rights reserved.
-// Licensed under the MIT License. See LICENSE file in the project root for full license information.
+// Licensed under the MIT License.
+// See LICENSE file in the project root for full license information.
 
 /**
  * @file conv2d_nhwc_f32.c
--- a/example/conv2d/conv2d_nhwc_f32.c
+++ b/example/conv2d/conv2d_nhwc_f32.c
@@ -1,5 +1,5 @@
-// Copyright 2026 SiFive, Inc.
-// SPDX-License-Identifier: Apache-2.0
+// Copyright (c) 2026 SiFive, Inc. All rights reserved.
+// Licensed under the MIT License. See LICENSE file in the project root for full license information.
 
 /**
  * @file conv2d_nhwc_f32.c
--- a/example/conv2d/conv2d_nhwc_f32.c
+++ b/example/conv2d/conv2d_nhwc_f32.c
@@ -1,6 +1,7 @@
-// Copyright (c) 2026 SiFive, Inc. All rights reserved.
+// Copyright (c) 2026-Present SiFive, Inc. All rights reserved.
 // Licensed under the MIT License.
 // See LICENSE file in the project root for full license information.
+// SPDX-License-Identifier: MIT
 
 /**
  * @file conv2d_nhwc_f32.c
--- a/example/conv2d/conv2d_nhwc_f32.h
+++ b/example/conv2d/conv2d_nhwc_f32.h
@@ -1,6 +1,7 @@
-// Copyright (c) 2026 SiFive, Inc. All rights reserved.
+// Copyright (c) 2026-Present SiFive, Inc. All rights reserved.
 // Licensed under the MIT License.
 // See LICENSE file in the project root for full license information.
+// SPDX-License-Identifier: MIT
 
 #pragma once
 
--- a/example/conv2d/conv2d_nhwc_f32.h
+++ b/example/conv2d/conv2d_nhwc_f32.h
@@ -1,4 +1,4 @@
-// Copyright (c) 2026-Present SiFive, Inc. All rights reserved.
+// Copyright (c) 2026 SiFive, Inc. All rights reserved.
 // Licensed under the MIT License.
 // See LICENSE file in the project root for full license information.
 // SPDX-License-Identifier: MIT
--- a/example/conv2d/conv2d_nhwc_f32.h
+++ b/example/conv2d/conv2d_nhwc_f32.h
@@ -1,5 +1,5 @@
-// Copyright 2026 SiFive, Inc.
-// SPDX-License-Identifier: Apache-2.0
+// Copyright (c) 2026 SiFive, Inc. All rights reserved.
+// Licensed under the MIT License. See LICENSE file in the project root for full license information.
 
 #pragma once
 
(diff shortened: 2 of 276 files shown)


example/conv2d/conv2d_utils.h
```diff
@@ -1,5 +1,6 @@
 // Copyright 2025 SiFive, Inc.
-// SPDX-License-Identifier: Apache-2.0
+// Licensed under the MIT License.
+// See LICENSE file in the project root for full license information.
```
Aaron-Hutchinson: In many files the first line was not updated consistently: the year remains as 2025, and they're missing the `(c)` and `All rights reserved.` text.

pvchupin: Good catch! Thanks!

example/conv2d/conv2d_nhwc_f32.c
```diff
@@ -1,5 +1,6 @@
-// Copyright 2026 SiFive, Inc.
-// SPDX-License-Identifier: Apache-2.0
```
ebavier, line 2: Why drop the `SPDX-License-Identifier` format?

example/conv2d/conv2d_nhwc_f32.c
```diff
@@ -1,5 +1,6 @@
...
+// Copyright (c) 2026 SiFive, Inc. All rights reserved.
```
ebavier, line 1: The additional "All rights reserved" seems unnecessarily verbose, especially if it is left out of the copyright statement in `LICENSE.txt`, which would seem to be the place with the most weight.

Aaron-Hutchinson:
How is it determined what the appropriate year for the license is? Is the year only updated to the current year when the actual C code changes? This should maybe be stated somewhere in our documentation.

example/conv2d/conv2d_nhwc_f32.c
pvchupin: I'm checking if it makes sense to update LICENSE.txt with the same.

example/conv2d/conv2d_nhwc_f32.c
pvchupin: Not sure how actually it's going to be used, but can return I think. Let me check, there are few ids for MIT.

pvchupin:
> How is it determined what the appropriate year for the license is? Is the year only updated to the current year when the actual C code changes? This should maybe be stated somewhere in our documentation.

I'll check for recommendation.

example/conv2d/conv2d_nhwc_f32.h
```diff
@@ -1,5 +1,6 @@
-// Copyright 2026 SiFive, Inc.
-// SPDX-License-Identifier: Apache-2.0
+// Copyright (c) 2026 SiFive, Inc. All rights reserved.
+// Licensed under the MIT License.
+// See LICENSE file in the project root for full license information.
```
Aaron-Hutchinson, line 3: These exact lines (possibly modulo the year) are mandatory for all C source files right? If that's the case there should be a step in CI that checks for this in all `.c` and `.h` files. Otherwise it's too easy to forget about it.

This would also make reviewing this PR pretty effortless.

pvchupin: Yes. We should have it in all source files, including makefiles and scripts. I absolutely agree on the CI check! 
@hdelassus can you help? If trivial effort I can wait and we can run and I can fix everything. Otherwise we can follow up on the CI later.

pvchupin:
> How is it determined what the appropriate year for the license is? Is the year only updated to the current year when the actual C code changes? This should maybe be stated somewhere in our documentation.

Ok, so recommendation is a range: [YYYY contribution started]-Present. Should be straightforward now. And we don't need to update really.

example/conv2d/conv2d_nhwc_f32.h
pvchupin: or maybe it's a question to @myeh01

Aaron-Hutchinson:
These files were missed or only partially updated:
```
test/exp/exp_bf16.c
test/exp/exp_f16.c
test/gelu/gelu_f32.c
test/logistic/logistic_f16.c
test/silu/silu_f16.c
test/skl-test-driver.c
test/softmax/softmax_2d_f32.c
test/softmax/softmax_bf16.c
test/softmax/softmax_f16.c
test/softmax/softmax_f32.c
```

Aaron-Hutchinson:
Also if CMake files are included then the `cmake` directory was missed.

pvchupin:
Thanks @Aaron-Hutchinson on highlighting these. Updated.

ebavier:
> > How is it determined what the appropriate year for the license is? Is the year only updated to the current year when the actual C code changes? This should maybe be stated somewhere in our documentation.
> 
> Ok, so recommendation is a range: [YYYY contribution started]-Present. Should be straightforward now. And we don't need to update really.

Literally "Present"?  I've never seen that in another project.  More often [YYYY contribution started]-[YYYY last modification].

pvchupin:
> > > How is it determined what the appropriate year for the license is? Is the year only updated to the current year when the actual C code changes? This should maybe be stated somewhere in our documentation.
> > 
> > 
> > Ok, so recommendation is a range: [YYYY contribution started]-Present. Should be straightforward now. And we don't need to update really.
> 
> Literally "Present"? I've never seen that in another project. More often [YYYY contribution started]-[YYYY last modification].

Yeah, me too... But that's my understanding, and better than specific year that we have to update.

example/conv2d/conv2d_nhwc_f32.h
myeh01: @pvchupin I can make a follow up PR for CI later.

Aaron-Hutchinson (dismissed):
LGTM

example/conv2d/conv2d_nhwc_f32.h
hdelassus: Thanks @myeh01 for opening #92

hdelassus (dismissed):
LGTM!

example/conv2d/conv2d_nhwc_f32.c
```diff
@@ -1,5 +1,7 @@
...
```
Aaron-Hutchinson, line 1: Is the start date supposed to be when the file was first created? Some files in the repo might have been created last year but already had their copyright date updated to 2026 prior to this PR.

If a file gets renamed is it considered a new file so that the start year resets? It's not very clear to me how these years are determined, or how much it even matters.

pvchupin: Since these files have never been released under this license before original year could be either 2025 or 2026, it doesn't matter much. We can reset everything to 2026 and that would be OK.
General rules are:
Range is [original contribution year]-[modification year].
When file is created: [current year] should be used.
When file is changed/renamed/copied [modification year] should be added if it's different from [original year].

ebavier: Copyright typically applies to the content, so a rename would retain the header of the original.  Similar with e.g. a function moved from one file to another, the copyrights would be merged into the destination.

hdelassus:
Now that #85 was merged, this PR should probably update the license in `src/logistic`.

pvchupin:
Thanks @hdelassus, rebased and added new files

hdelassus (approved):

PR #90, opened 2026-05-01
[Test][GEMM] Convert `f64_f64_f64` harness and tests

Aaron-Hutchinson:
This converts the non-widening f16 GEMM harness into the new format introduced in #56. Only one kernel currently uses this harness:
- `skl_gemm_f64_f64_f64_zve64d_x390`

The f128 reference function was moved from `skl-test.h` directly into the harness, since with packed format this should be the only harness using it.


test/gemm/gemm_f64rcprc_f64rcprc_f64rcprc.c
```diff
@@ -0,0 +1,274 @@
...
+
+#include "gemm_f64rcprc_f64rcprc_f64rcprc.h"
+#include "skl-ref.h"
+#include "skl-test-driver.h"
+#include "skl_test_gemm.h"
+#include <inttypes.h>
+#include <math.h>
+#include <stddef.h>
+#include <stdint.h>
+#include <stdio.h>
```
myeh01:
```suggestion
#include <math.h>
#include <stddef.h>
```

test/gemm/gemm_f64rcprc_f64rcprc_f64rcprc.c
```diff
@@ -0,0 +1,274 @@
...
+          cp_block[ii0 * rsc0 + jj0 * csc0] =
+              beta * cp_block[ii0 * rsc0 + jj0 * csc0] + alpha * acc;
+        }
+      }
+    }
+  }
+}
+
+void gemm_f64rcprc_f64rcprc_f64rcprc_verify(skl_test_t *t) {
+  /* Compute the reference matrix output. */
```
myeh01:
```suggestion
```

test/gemm/gemm_f64rcprc_f64rcprc_f64rcprc.c
```diff
@@ -0,0 +1,274 @@
...
+    for (size_t j1 = 0; j1 < n1; ++j1) {
+      for (size_t i0 = 0; i0 < m0; ++i0) {
+        for (size_t j0 = 0; j0 < n0; ++j0) {
+          size_t idx = i1 * rsc1 + j1 * csc1 + i0 * rsc0 + j0 * csc0;
+          if (fabsl((long double)c_pack[idx] - (long double)ref_c[idx]) >
+              bound[idx]) {
+            SKL_TEST_LOG(
+                t, SKL_TEST_LOG_ERROR,
+                "result [%zu, %zu, %zu, %zu] (%.10f) != reference (%.10f) "
+                "[bound = %.20Lf]\n",
```
myeh01:
```suggestion
                "result [%zu, %zu, %zu, %zu] (%f) != reference (%f) "
                "[bound = %.8Le]\n",
```
Something like this might give a cleaner output.

Aaron-Hutchinson: I made this change because not enough digits were being printed to see the difference between the result and reference.

myeh01: Right, so using the `e` format specifier will remove all the leading zeros.

Edit: specifically for the bound. The `%.10f` for the result and reference are probably ok.

Aaron-Hutchinson: Oh, I misunderstood your suggestion. Thanks for clarifying.

myeh01 (approved):
LGTM

PR #92, opened 2026-05-05
[CI] Check license headers

myeh01:
This PR adds a script by @ericlove to check each file in the repo for a correct license header if needed and adds this check to CI.

Currently based on #88, but will rebase after it merges.

### DIFF
--- a/.github/check_license_headers.py
+++ b/.github/check_license_headers.py
@@ -8,7 +8,7 @@
 Script to check for SiFive copyright and MIT license headers in source files.
 This script verifies that each relevant file contains the following header:
 
-    Copyright (c) {year_str} SiFive, Inc. All rights reserved.
+    Copyright (c) (YYYY-)YYYY SiFive, Inc. All rights reserved.
     Licensed under the MIT License.
     See LICENSE file in the project root for full license information.
     SPDX-License-Identifier: MIT
@@ -83,30 +83,22 @@ def get_comment_style(filepath: Path) -> str:
     return ''
 
 
-def get_copyright_year(filepath: Path) -> str:
+def get_year_modified(filepath: Path) -> str:
     current_year = datetime.now().strftime('%Y')
-    date_created = subprocess.check_output(['git', 'log', '--pretty=format:%cI', '--follow', filepath]).decode('utf-8').rsplit('\n', 1)[-1]
     date_modified = subprocess.check_output(['git', 'log', '-1', '--pretty=format:%cI', filepath]).decode('utf-8')
-    year_created = date_created.split('-')[0] or current_year
     year_modified = date_modified.split('-')[0] or current_year
-
-    if year_created != year_modified:
-        year_str = f"{year_created}-{year_modified}"
-    else:
-        year_str = f"{year_modified}"
-
-    return year_str
+    return year_modified
 
 
 def generate_license_pattern(filepath: Path) -> str:
     """Generate the appropriate license header pattern for a file."""
     comment_prefix = get_comment_style(filepath)
-    year_str = get_copyright_year(filepath)
+    year_modified = get_year_modified(filepath)
     header_lines = [
-        f"{comment_prefix} Copyright (c) {year_str} SiFive, Inc. All rights reserved.\n",
-        f"{comment_prefix} Licensed under the MIT License.\n",
-        f"{comment_prefix} See LICENSE file in the project root for full license information.\n",
-        f"{comment_prefix} SPDX-License-Identifier: MIT\n",
+        rf"{comment_prefix} Copyright \(c\) ((\d{{4}})-)?{year_modified} SiFive, Inc\. All rights reserved\.\n",
+        rf"{comment_prefix} Licensed under the MIT License\.\n",
+        rf"{comment_prefix} See LICENSE file in the project root for full license information\.\n",
+        rf"{comment_prefix} SPDX-License-Identifier: MIT\n",
         ""
     ]
     return ''.join(header_lines)
@@ -132,11 +124,17 @@ def check_file_header(filepath: Path) -> List[str]:
             header_text = ''.join(header_lines)
 
             license_pattern = generate_license_pattern(filepath)
-            if not license_pattern in header_text:
+            match = re.search(license_pattern, header_text)
+            if not match:
                 issues.append("Missing or incorrect header. Should be:\n" + license_pattern)
+            else:
+                start_year = match.groups()[1]
+                year_modified = get_year_modified(filepath)
+                if start_year and (int(start_year) < 2025 or int(start_year) >= int(year_modified)):
+                    issues.append(f"Invalid start year. Should be >= 2025 and < {year_modified}.")
 
     except Exception as e:
-        issues.append(f"Error reading file: {e}")
+        issues.append(f"Error: {e}")
 
     return issues
 
--- a/.github/check_license_headers.py
+++ b/.github/check_license_headers.py
@@ -28,10 +28,11 @@ Exit codes:
     1 - One or more files are missing headers
 """
 
-import sys
-import re
 import argparse
+import re
 import subprocess
+import sys
+from datetime import datetime
 from pathlib import Path
 from typing import List, Tuple
 
@@ -83,10 +84,11 @@ def get_comment_style(filepath: Path) -> str:
 
 
 def get_copyright_year(filepath: Path) -> str:
+    current_year = datetime.now().strftime('%Y')
     date_created = subprocess.check_output(['git', 'log', '--pretty=format:%cI', '--follow', filepath]).decode('utf-8').rsplit('\n', 1)[-1]
     date_modified = subprocess.check_output(['git', 'log', '-1', '--pretty=format:%cI', filepath]).decode('utf-8')
-    year_created = date_created.split('-')[0]
-    year_modified = date_modified.split('-')[0]
+    year_created = date_created.split('-')[0] or current_year
+    year_modified = date_modified.split('-')[0] or current_year
 
     if year_created != year_modified:
         year_str = f"{year_created}-{year_modified}"
--- a/.github/check_license_headers.py
+++ b/.github/check_license_headers.py
@@ -34,7 +34,7 @@ import subprocess
 import sys
 from datetime import datetime
 from pathlib import Path
-from typing import List, Tuple
+from typing import List
 
 # Comment styles for different file types
 # The script defaults to '' for types not listed here
--- a/.github/check_license_headers.py
+++ b/.github/check_license_headers.py
@@ -57,8 +57,9 @@ EXCLUDE_PATTERNS = [
     r'.*\.md$',
     '.clang-tidy',
     '.clang-format',
+    r'.*CODEOWNERS$',
     r'.*Doxyfile$',
-    '.github/',
+    '.github/workflows/',
     '.gitignore',
     'LICENSE.txt',
 ]
@@ -155,7 +156,7 @@ def main():
     )
     args = parser.parse_args()
 
-    repo_root = Path(__file__).parent.resolve()
+    repo_root = Path(__file__).resolve().parent.parent
 
     files_skipped = 0
     files_ok = []
--- a/.github/check_license_headers.py
+++ b/.github/check_license_headers.py
@@ -32,6 +32,7 @@ import argparse
 import re
 import subprocess
 import sys
+from codecs import decode
 from datetime import datetime
 from pathlib import Path
 from typing import List
@@ -90,18 +91,34 @@ def get_year_modified(filepath: Path) -> str:
     return year_modified
 
 
-def generate_license_pattern(filepath: Path) -> str:
-    """Generate the appropriate license header pattern for a file."""
+def generate_license(filepath: Path, generate_pattern: bool = False) -> str:
+    """
+    Generate the appropriate license header for a file.
+
+    If generate_pattern is True, returns a regex for the license header.
+    Otherwise, returns the license header itself.
+    """
     comment_prefix = get_comment_style(filepath)
     year_modified = get_year_modified(filepath)
+    if generate_pattern:
+        esc = "\\"
+        start_year = r"((\d{4})-)?"
+    else:
+        esc = ""
+        start_year = "(YYYY-)"
+
     header_lines = [
-        rf"{comment_prefix} Copyright \(c\) ((\d{{4}})-)?{year_modified} SiFive, Inc\. All rights reserved\.\n",
-        rf"{comment_prefix} Licensed under the MIT License\.\n",
-        rf"{comment_prefix} See LICENSE file in the project root for full license information\.\n",
+        rf"{comment_prefix} Copyright {esc}(c{esc}) {start_year}{year_modified} SiFive, Inc{esc}. All rights reserved{esc}.\n",
+        rf"{comment_prefix} Licensed under the MIT License{esc}.\n",
+        rf"{comment_prefix} See LICENSE file in the project root for full license information{esc}.\n",
         rf"{comment_prefix} SPDX-License-Identifier: MIT\n",
-        ""
     ]
-    return ''.join(header_lines)
+    header = "".join(header_lines)
+
+    if not generate_pattern:
+        header = decode(header, 'unicode-escape')
+
+    return header
 
 
 def check_file_header(filepath: Path) -> List[str]:
@@ -123,10 +140,10 @@ def check_file_header(filepath: Path) -> List[str]:
                 header_lines.append(line)
             header_text = ''.join(header_lines)
 
-            license_pattern = generate_license_pattern(filepath)
+            license_pattern = generate_license(filepath, True)
             match = re.search(license_pattern, header_text)
             if not match:
-                issues.append("Missing or incorrect header. Should be:\n" + license_pattern)
+                issues.append("Missing or incorrect header. Should be:\n" + generate_license(filepath))
             else:
                 start_year = match.groups()[1]
                 year_modified = get_year_modified(filepath)
--- a/.github/check_license_headers.py
+++ b/.github/check_license_headers.py
@@ -115,7 +115,7 @@ def check_file_header(filepath: Path) -> List[str]:
     Check if a file has the required copyright and license header.
 
     Returns:
-        Tuple of (has_valid_header, list_of_issues)
+        List of issues
     """
     issues = []
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,7 +32,7 @@ jobs:
         HEAD_SHA=${{ github.event.pull_request.head.sha }}
         BASE_SHA=${{ github.event.pull_request.base.sha }}
         MERGE_BASE_SHA=$(git merge-base $HEAD_SHA $BASE_SHA)
-        git diff --name-only $MERGE_BASE_SHA $HEAD_SHA | xargs ./check_license_headers.py
+        git diff --name-only $MERGE_BASE_SHA $HEAD_SHA | xargs ./.github/check_license_headers.py
 
     - name: Setup QEMU
       run: |
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,6 +27,10 @@ jobs:
       with:
         fetch-depth: 0
 
+    - name: Check license headers 
+      run: |
+        ./check_license_headers.py
+
     - name: Setup QEMU
       run: |
         gh release download $SKL_TAG -p "$QEMU_TARBALL"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,9 +27,12 @@ jobs:
       with:
         fetch-depth: 0
 
-    - name: Check license headers 
+    - name: Check license headers
       run: |
-        ./check_license_headers.py
+        HEAD_SHA=${{ github.event.pull_request.head.sha }}
+        BASE_SHA=${{ github.event.pull_request.base.sha }}
+        MERGE_BASE_SHA=$(git merge-base $HEAD_SHA $BASE_SHA)
+        git diff --name-only $MERGE_BASE_SHA $HEAD_SHA | xargs ./check_license_headers.py
 
     - name: Setup QEMU
       run: |
--- a/check_license_headers.py
+++ b/check_license_headers.py
@@ -1,0 +1,474 @@
+#!/usr/bin/env python3
+# Copyright (c) 2025-Present SiFive, Inc. All rights reserved.
+# Licensed under the MIT License.
+# See LICENSE file in the project root for full license information.
+# SPDX-License-Identifier: MIT
+
+"""
+Script to check for SiFive copyright and MIT license headers in source files.
+
+This script verifies that each relevant file contains:
+1. A copyright notice with "SiFive, Inc." and years 2025-Present or 2026-Present
+2. MIT License reference
+3. Proper SPDX-License-Identifier
+
+Usage:
+    # Check all files in the repository
+    ./check_license_headers.py
+
+    # Check specific files
+    ./check_license_headers.py file1.c file2.h
+
+    # Show file type summary
+    ./check_license_headers.py --summary
+
+    # Show verbose output (lists all valid files)
+    ./check_license_headers.py --verbose
+
+    # Show example headers
+    ./check_license_headers.py --examples
+
+    # Fix files with missing headers
+    ./check_license_headers.py --fix
+
+    # Fix specific files only
+    ./check_license_headers.py --fix file1.c file2.h
+
+Exit codes:
+    0 - All files have proper headers
+    1 - One or more files are missing headers
+"""
+
+import os
+import sys
+import re
+import argparse
+from pathlib import Path
+from typing import List, Tuple, Set
+
+# Expected copyright patterns (allowing for 2025 or 2026)
+COPYRIGHT_PATTERNS = [
+    r"Copyright\s+\(c\)\s+202[56]-Present\s+SiFive,?\s+Inc\.\s+All\s+rights\s+reserved\.",
+]
+
+# Expected license reference
+LICENSE_REFERENCE = r"Licensed\s+under\s+the\s+MIT\s+License\."
+
+# Expected SPDX identifier
+SPDX_IDENTIFIER = r"SPDX-License-Identifier:\s*MIT"
+
+# File extensions to check and their comment styles
+FILE_TYPES = {
+    # C/C++ style comments
+    '.c': ('//', '//'),
+    '.h': ('//', '//'),
+    '.cpp': ('//', '//'),
+    '.hpp': ('//', '//'),
+    # Python style comments
+    '.py': ('#', '#'),
+    # CMake style comments
+    '.cmake': ('#', '#'),
+    'CMakeLists.txt': ('#', '#'),
+}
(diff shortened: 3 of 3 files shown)


Aaron-Hutchinson (changes requested):
Thanks for taking on this work @myeh01. I'd like for this to be simplified and cleaned up dramatically instead of just committing the AI slop. High level feedback to start with:
- The feature set can be reduced significantly. I don't think this tool needs to be modifying any files. Just printing a list of all files in the repo with an invalid license should be all we need.
- Why do we need to print examples or generate headers?
- Due to the above I don't see a reason that any command line arguments would be needed, so the `argparse` logic can be removed.
- I don't think it's necessary to exclude specific directories or files. Shouldn't it be sufficient to just traverse the whole repo and check files whose filename suffix matches `.c`, `.h`, `CMakeLists.txt`, etc?
- The valid start year for the license check looks to be hardcoded as 2025 or 2026, which will break next year. The end year is "Present", which there was also discussion around changing.

I'm not personally opposed to Python, but this is the first instance of it in this repository and I'm not sure what the implications of that are. Do we need to ensure a common version / virtual environment is being used for compatibility with this script?

check_license_headers.py
```diff
@@ -0,0 +1,474 @@
+#!/usr/bin/env python3
+# Copyright (c) 2025-Present SiFive, Inc. All rights reserved.
```
hdelassus: 2026?

check_license_headers.py
```diff
@@ -0,0 +1,474 @@
...
+# Expected copyright patterns (allowing for 2025 or 2026)
+COPYRIGHT_PATTERNS = [
+    r"Copyright\s+\(c\)\s+202[56]-Present\s+SiFive,?\s+Inc\.\s+All\s+rights\s+reserved\.",
+]
+
+# Expected license reference
+LICENSE_REFERENCE = r"Licensed\s+under\s+the\s+MIT\s+License\."
+
+# Expected SPDX identifier
+SPDX_IDENTIFIER = r"SPDX-License-Identifier:\s*MIT"
```
hdelassus: Nit: `\s+`
Reference: https://spdx.dev/learn/handling-license-info/#how

check_license_headers.py
```diff
@@ -0,0 +1,474 @@
...
+    '.c': ('//', '//'),
+    '.h': ('//', '//'),
+    '.cpp': ('//', '//'),
+    '.hpp': ('//', '//'),
+    # Python style comments
+    '.py': ('#', '#'),
+    # CMake style comments
+    '.cmake': ('#', '#'),
+    'CMakeLists.txt': ('#', '#'),
+}
```
hdelassus: It would be nice to have a comment explaining here what goes in each tuple so as to not need to scroll down to understand how this works, e.g. when adding a new file type.

check_license_headers.py
```diff
@@ -0,0 +1,474 @@
...
+
+# Directories to exclude
+EXCLUDE_DIRS = {
+    '.git',
+    'build',
+    'doc/html',
+    '__pycache__',
+    '.pytest_cache',
+}
```
hdelassus: I think we could limit excluded directories to those that are tracked by git, so nothing already here, really.

check_license_headers.py
```diff
@@ -0,0 +1,474 @@
...
+EXCLUDE_FILES = {
+    '.clang-tidy',
+    '.gitignore',
+    'LICENSE.txt',
+    'CHANGELOG.md',
+    'README.md',
+    'CONTRIBUTING.md',
+    'CONTRIBUTORS.md',
+    '.github/CODEOWNERS',
+}
```
hdelassus: I think all of those could be integrated into the exclude patterns below.
Also: missing `.clang-format`?

check_license_headers.py
```diff
@@ -0,0 +1,474 @@
...
+# See LICENSE file in the project root for full license information.
+# SPDX-License-Identifier: MIT
+""")
+
+
+def main():
+    """Main function to check all files in the repository."""
+    parser = argparse.ArgumentParser(
+        description="Check for SiFive copyright and MIT license headers in source files.",
+        epilog="Note: Copyright year can be either 2025-Present or 2026-Present"
```
hdelassus: Not accounting for 2027 and beyond.

check_license_headers.py
```diff
@@ -0,0 +1,474 @@
...
+        '--fix',
+        action='store_true',
+        help='Automatically add missing headers to files'
+    )
+    parser.add_argument(
+        '--year',
+        choices=['2025', '2026'],
+        default='2025',
+        help='Copyright year to use when fixing files (default: 2025)'
+    )
```
hdelassus: Hmm, I'm not sure about those limited choices.

check_license_headers.py
```diff
@@ -0,0 +1,474 @@
...
+import os
+import sys
+import re
+import argparse
+from pathlib import Path
+from typing import List, Tuple, Set
+
+# Expected copyright patterns (allowing for 2025 or 2026)
+COPYRIGHT_PATTERNS = [
+    r"Copyright\s+\(c\)\s+202[56]-Present\s+SiFive,?\s+Inc\.\s+All\s+rights\s+reserved\.",
```
hdelassus: Needs to account for future years.

check_license_headers.py
```diff
@@ -0,0 +1,474 @@
...
+    Generate the appropriate copyright header for a file.
+
+    Args:
+        filepath: Path to the file
+        year: Starting copyright year (2025 or 2026)
+
+    Returns:
+        String containing the copyright header
+    """
+    comment_prefix, _ = get_comment_style(filepath)
```
hdelassus: Now I have no idea what the tuple's second member is supposed to mean since it's simply ignored. I can only guess it might have been for comment suffixes, but then the hardcoded values are incorrect.

check_license_headers.py
```diff
@@ -0,0 +1,474 @@
...
+
+
+def get_comment_style(filepath: Path) -> Tuple[str, str]:
+    """Get the comment style for a file."""
+    if filepath.name == 'CMakeLists.txt':
+        return FILE_TYPES['CMakeLists.txt']
+    return FILE_TYPES.get(filepath.suffix, (None, None))
+
+
+def generate_header(filepath: Path, year: str = "2025") -> str:
```
hdelassus: No need to put a default year if the year option already has a default?

myeh01:
> Thanks for taking on this work Michael Yeh (@myeh01). I'd like for this to be simplified and cleaned up dramatically instead of just committing the AI slop. High level feedback to start with:
> 
> * The feature set can be reduced significantly. I don't think this tool needs to be modifying any files. Just printing a list of all files in the repo with an invalid license should be all we need.
> * Why do we need to print examples or generate headers?
> * Due to the above I don't see a reason that any command line arguments would be needed, so the `argparse` logic can be removed.
> * I don't think it's necessary to exclude specific directories or files. Shouldn't it be sufficient to just traverse the whole repo and check files whose filename suffix matches `.c`, `.h`, `CMakeLists.txt`, etc?
> * The valid start year for the license check looks to be hardcoded as 2025 or 2026, which will break next year. The end year is "Present", which there was also discussion around changing.
> 
> I'm not personally opposed to Python, but this is the first instance of it in this repository and I'm not sure what the implications of that are. Do we need to ensure a common version / virtual environment is being used for compatibility with this script?

Thanks, I agree this script could be simplified. Wanted to open this to get a discussion started. I was originally thinking we might be able to use a shell script to do this. What do others thinks?

.github/workflows/ci.yml
```diff
@@ -27,6 +27,10 @@ jobs:
       with:
         fetch-depth: 0
 
+    - name: Check license headers 
+      run: |
+        ./check_license_headers.py
```
hdelassus: I believe we could `git ls-files | xargs ./check...`. Right now, the script doesn't ignore any file that is passed as positional arguments even if it is matching one or more of the exclude rules, but that could be fixed.

check_license_headers.py
```diff
@@ -0,0 +1,474 @@
...
+    'CHANGELOG.md',
+    'README.md',
+    'CONTRIBUTING.md',
+    'CONTRIBUTORS.md',
+    '.github/CODEOWNERS',
+}
+
+# Exclude files matching these patterns
+EXCLUDE_PATTERNS = [
+    r'.*README\.md$',
```
hdelassus: No need for this one, since all markdown files are ignored.

ebavier:
> What do others thinks?

+1 for shell script.

check_license_headers.py
```diff
@@ -0,0 +1,474 @@
...
+    """
+    Check if a file has the required copyright and license header.
+    
+    Returns:
+        Tuple of (has_valid_header, list_of_issues)
+    """
+    issues = []
+    
+    try:
+        with open(filepath, 'r', encoding='utf-8', errors='ignore') as f:
```
hdelassus: Do you think we should keep `errors='ignore'` or leave it as the None default, meaning the same as `'strict'`?

Aaron-Hutchinson:
I originally also thought a shell script would be best to not introduce unnecessary dependencies. But I imagine we might want to use Python for something in the future anyway, if not for this then something else.

I'm in favor of whatever gives the simplest script. Python is pretty simple, but I'm not experienced enough with shell scripting to see how involved this task would be there.

check_license_headers.py
```diff
@@ -0,0 +1,474 @@
...
+                if i >= 20:
+                    break
+                header_lines.append(line)
+            
+            header_text = ''.join(header_lines)
+            
+            # Check for copyright notice
+            copyright_found = False
+            for pattern in COPYRIGHT_PATTERNS:
+                if re.search(pattern, header_text, re.IGNORECASE):
```
hdelassus: Do we want to ignore the case?

check_license_headers.py
```diff
@@ -0,0 +1,474 @@
...
+                    copyright_found = True
+                    break
+            
+            if not copyright_found:
+                issues.append("Missing or incorrect SiFive copyright notice (should be '2025-Present' or '2026-Present')")
+            
+            # Check for MIT license reference
+            if not re.search(LICENSE_REFERENCE, header_text, re.IGNORECASE):
```
hdelassus: Do we want to ignore the case?

hdelassus (changes requested):
Thanks for opening this PR. I believe it's ok to add this AI-generated script to the repository, even if far from perfect. After all, it's "working" according to some definition of "working".

First, I have a few nits that I'd like to see fixed (maybe not all of them) before merging, though.

Second, perhaps the script should probably not be located at the root. Maybe in `.github/`?

I also noticed there were trailing whitespaces left behind:
```
$ git diff --check main
.github/workflows/ci.yml:30: trailing whitespace.
+    - name: Check license headers
check_license_headers.py:204: trailing whitespace.
+
check_license_headers.py:209: trailing whitespace.
+
check_license_headers.py:218: trailing whitespace.
+
check_license_headers.py:220: trailing whitespace.
+
check_license_headers.py:227: trailing whitespace.
+
check_license_headers.py:230: trailing whitespace.
+
check_license_headers.py:234: trailing whitespace.
+
check_license_headers.py:238: trailing whitespace.
+
check_license_headers.py:241: trailing whitespace.
+
```

pvchupin:
Posted another update to #88 with years updated. You can check the script again.

myeh01:
I have cleaned up the script. Please let me know what you think.

check_license_headers.py
```diff
@@ -0,0 +1,234 @@
...
+        if re.match(pattern, rel_path_str):
+            return False
+    return True
+
+
+def get_comment_style(filepath: Path) -> str:
+    """Get the comment style for a file."""
+    if filepath.name == 'CMakeLists.txt':
+        return FILE_TYPES[filepath.name]
+    return FILE_TYPES.get(filepath.suffix, '')
```
Aaron-Hutchinson: Rather than special-casing the `CMakeLists.txt`:

```py
for suffix in FILE_TYPES:
  if filepath.endswith(suffix):
    return FILE_TYPES[suffix]
return some_default
```

.github/check_license_headers.py
```diff
@@ -0,0 +1,234 @@
...
+]
+
+def should_check_file(filepath: Path, repo_root: Path) -> bool:
+    """Determine if a file should be checked."""
+    rel_path = filepath.relative_to(repo_root)
+    rel_path_str = str(rel_path)
+    for pattern in EXCLUDE_PATTERNS:
+        if re.match(pattern, rel_path_str):
+            return False
+    return True
```
Aaron-Hutchinson, line 76: Doesn't `FILE_TYPES` already determine what files should be checked? It seems like we can just omit `EXCLUDE_PATTERNS` altogether and just check if the filename suffix is a key in `FILE_TYPES`. I think this would also remove the need for the `repo_root` argument.

check_license_headers.py
```diff
@@ -0,0 +1,234 @@
...
+
+
+def get_copyright_year(filepath: Path) -> str:
+    year_from_date = lambda date: date.split(b'-')[0].decode('utf-8')
```
Aaron-Hutchinson: Some linters will complain about this. https://www.flake8rules.com/rules/E731.html

check_license_headers.py
```diff
@@ -0,0 +1,234 @@
...
+def generate_license_pattern(filepath: Path) -> str:
+    """Generate the appropriate license header pattern for a file."""
+    comment_prefix = get_comment_style(filepath)
+    year_str = get_copyright_year(filepath)
+    header_lines = [
+        rf"{comment_prefix} Copyright \(c\) {year_str} SiFive, Inc\. All rights reserved\.(\r)?\n",
+        rf"{comment_prefix} Licensed under the MIT License\.(\r)?\n",
+        rf"{comment_prefix} See LICENSE file in the project root for full license information\.(\r)?\n",
+        rf"{comment_prefix} SPDX-License-Identifier: MIT(\r)?\n",
+        ""
```
Aaron-Hutchinson: What's the reason for raw strings here? It seems like this would work:
```py
    return "\n".join([
        f"{comment_prefix} Copyright (c) {year_str} SiFive, Inc. All rights reserved.",
        f"{comment_prefix} Licensed under the MIT License.",
        f"{comment_prefix} See LICENSE file in the project root for full license information.",
        f"{comment_prefix} SPDX-License-Identifier: MIT",
        ""
    ])
```

.github/check_license_headers.py
```diff
@@ -0,0 +1,234 @@
...
+
+    try:
+        with open(filepath, 'r') as f:
+            # Read first 6 lines (header should be in this range)
+            header_lines = []
+            for i, line in enumerate(f):
+                if i >= 6:
+                    break
+                header_lines.append(line)
+            header_text = ''.join(header_lines)
```
Aaron-Hutchinson, line 141: Might be more Pythonic to do:
```py
with open(filepath, 'r') as f:
  lines = list(line for line in f)[:6]
  header_text = ''.join(lines)
```

check_license_headers.py
```diff
@@ -0,0 +1,234 @@
...
+            # Read first 6 lines (header should be in this range)
+            header_lines = []
+            for i, line in enumerate(f):
+                if i >= 6:
+                    break
+                header_lines.append(line)
+            header_text = ''.join(header_lines)
+
+            license_pattern = generate_license_pattern(filepath)
+            if not re.search(license_pattern, header_text):
```
Aaron-Hutchinson: Is the `re.search` necessary? Seems like the the `license_pattern` should be a substring of `header_text`, so maybe `license_pattern in header_text` is sufficient.

check_license_headers.py
```diff
@@ -0,0 +1,234 @@
...
+                issues.append("Missing or incorrect header. Should be: " + license_pattern)
+
+    except Exception as e:
+        issues.append(f"Error reading file: {e}")
+
+    return (len(issues) == 0, issues)
```
Aaron-Hutchinson: The return value is a bit odd.

1. The boolean seems unnecessary since the caller can just do `if issues` on the returned list.
2. From the above code the list only ever contains zero or one entries, so maybe just return the error string, and `''` or `None` if no error.

.github/check_license_headers.py
```diff
@@ -0,0 +1,234 @@
...
+def main():
+    """Main function to check given files in the repository."""
+    parser = argparse.ArgumentParser(
+        description="Check for SiFive copyright and MIT license headers in source files.",
+    )
+    parser.add_argument(
+        'files',
+        nargs='*',
+        help='Files to check'
+    )
```
Aaron-Hutchinson, line 168: Currently the script runs successfully with no arguments but checks no files, which isn't very useful. `files` should either be a required argument, or the default should be to check all files in the repo. My preference is the latter.

Another alternative: the script runs fast enough that maybe checking all files can just be the sole functionality, and the `files` argument can just be removed.

check_license_headers.py
```diff
@@ -0,0 +1,234 @@
...
+    1 - One or more files are missing headers
+"""
+
+import os
+import sys
+import re
+import argparse
+import subprocess
+from pathlib import Path
+from typing import List, Tuple, Set
```
Aaron-Hutchinson: `os` and `Set` appear unused.

.github/check_license_headers.py
```diff
@@ -0,0 +1,234 @@
...
+
+        if not filepath.is_file():
+            print(f"Warning: Not a file: {filepath}")
+            continue
+
+        try:
+            rel_path = filepath.relative_to(repo_root)
+        except ValueError:
+            print(f"Warning: File outside repository: {filepath}")
+            continue
```
Aaron-Hutchinson, line 204: What's the motivation in checking for this? We don't currently submodule anything, and the logic used here wouldn't account for that anyway. Might not be our intention to check files outside the repo, but I don't see the harm in it either and it's maybe not worth concerning ourselves with that use case.

check_license_headers.py
```diff
@@ -0,0 +1,234 @@
...
+
+        files_checked += 1
+
+        has_valid_header, issues = check_file_header(filepath)
+
+        if has_valid_header:
+            files_ok.append(rel_path)
+        else:
+            files_with_issues += 1
+            all_issues.append((rel_path, issues))
```
Aaron-Hutchinson: Might be able to clean this up a little after my suggestions to the `check_file_header` return value. Also looks like `file_checked == len(files_ok) + len(all_issies)` always holds.

Aaron-Hutchinson (commented):
Thanks for the heavy revision! It's looking much better.

check_license_headers.py
myeh01: Done

check_license_headers.py
myeh01: Fixed

check_license_headers.py
myeh01: Windows uses `\r\n` for newlines, so the end of each line includes an optional `\r`: `(\r)?\n`. If we don't need to support this, then we can just use regular strings and `in` instead of regexes and `re.search()`.

.github/check_license_headers.py
myeh01: Using `enumerate(f)` avoids reading in the whole file.

check_license_headers.py
myeh01: See above comment.

check_license_headers.py
myeh01: I've changed the return value to just a list of issues. I think it's fine to return an empty list if there are no errors.

check_license_headers.py
ebavier: No, we shouldn't need to support windows-encoded files.

.github/check_license_headers.py
myeh01: `files` is already a positional argument.

We only want to check files that are tracked by git. I liked @hdelassus's suggestion of using `git ls-files | xargs ./check_license_headers.py`. We could possibly put `git ls-files` into the script itself.

check_license_headers.py
myeh01: Removed

.github/check_license_headers.py
myeh01: I think it's fine to just leave it.

check_license_headers.py
myeh01: I've removed `files_checked` and `files_with_issues`. Assuming there are no files with `File not found`/`Not a file`/`File outside repository`, `files_checked` should equal `files_skipped + len(files_ok) + len(all_issues)`.

.github/check_license_headers.py
Aaron-Hutchinson: It's positional but the `nargs='*'` makes providing zero arguments valid. If none are provided then using `git ls-files` to generate a list of files to check might be a nice default behavior.

check_license_headers.py
myeh01: Updated

.github/check_license_headers.py
myeh01: Updated

check_license_headers.py
```diff
@@ -0,0 +1,234 @@
...
+    ]
+    return ''.join(header_lines)
+
+
+def check_file_header(filepath: Path) -> List[str]:
+    """
+    Check if a file has the required copyright and license header.
+
+    Returns:
+        Tuple of (has_valid_header, list_of_issues)
```
Aaron-Hutchinson: Docstring needs an update since the last changes

check_license_headers.py
```diff
@@ -0,0 +1,234 @@
...
+        if str(filepath).endswith(suffix):
+            return FILE_TYPES[suffix]
+    return ''
+
+
+def get_copyright_year(filepath: Path) -> str:
+    date_created = subprocess.check_output(['git', 'log', '--pretty=format:%cI', '--follow', filepath]).decode('utf-8').rsplit('\n', 1)[-1]
+    date_modified = subprocess.check_output(['git', 'log', '-1', '--pretty=format:%cI', filepath]).decode('utf-8')
+    year_created = date_created.split('-')[0]
+    year_modified = date_modified.split('-')[0]
```
Aaron-Hutchinson: The above git commands return an empty string on files that have never been committed. I'm seeing this result in a recommended license string containing no year. Maybe the easiest fix is something like:
```py
year_created = date_created.split('-')[0] or current_year
```

cmake/detect_riscv_extensions.cmake
```diff
@@ -1,4 +1,4 @@
-# Copyright (c) 2025-2026 SiFive, Inc. All rights reserved.
+# Copyright (c) 2026 SiFive, Inc. All rights reserved.
```
Aaron-Hutchinson: I guess we are somewhat at the mercy of what git determines the start year to be. I would consider this file to be originally created in 2025 under the name `get-riscv-extensions.cmake`, but in #69 it was renamed with significant enough changes that git considered it to be a new file. Unsure how much we care about these details. CC @pvchupin

hdelassus (dismissed):
LGTM. What do you think about my earlier suggestion to locate this script somewhere less "front and center", like `.github`?

cmake/detect_riscv_extensions.cmake
pvchupin: Good point. @myeh01 would it be possible to distinguish rename and add in the script?
For add we can just check that there is a current year.
For rename we check for second year, unless it's happening in the current year.
Alternatively we can also simplify checks and check only for the second year (or the only year) matching the current year and treat rename/add the same.

check_license_headers.py
myeh01: Fixed!

check_license_headers.py
myeh01: Fixed!

cmake/detect_riscv_extensions.cmake
myeh01: I don't really understand how `git log --follow` works. After rebasing, it shows that the new test harness from #97 `test/gemm/f8rcprc_f8rcprc_f32rcprc.c` was created in 2025. When I looked at it in more detail with `git log --follow -p test/gemm/gemm_f8rcprc_f8rcprc_f32rcprc.c`, it shows some changes related to `test/gemm/gemm_bf16rcprc_bf16rcprc_f32rcprc.c`, which is not really how we wanted to track it.
What do people think about just resetting the copyright year for all renames? Or does someone else who's more familiar with `git` know a better way to track file history including through renames?

Aaron-Hutchinson: What about simplifying these checks in the following way:
- Assert that the ending year is correct by ensuring it matches the year of the most recent commit that changed the file, or the current year if it's a new file. I think this is the current logic in the script.
- Assert that the starting year is correct (if present) by accepting any value between 2025 and the ending year determined above. This leaves it up to us to determine the start year and avoids dealing with file renames and git's logic.

myeh01: Yeah, this sounds reasonable.

myeh01:
> LGTM. What do you think about my earlier suggestion to locate this script somewhere less "front and center", like `.github`?

I have moved the script to `.github`.

cmake/detect_riscv_extensions.cmake
myeh01: Ok, see what you think.

.github/check_license_headers.py
```diff
@@ -0,0 +1,235 @@
...
+            for i, line in enumerate(f):
+                if i >= 6:
+                    break
+                header_lines.append(line)
+            header_text = ''.join(header_lines)
+
+            license_pattern = generate_license_pattern(filepath)
+            match = re.search(license_pattern, header_text)
+            if not match:
+                issues.append("Missing or incorrect header. Should be:\n" + license_pattern)
```
Aaron-Hutchinson: For incorrect headers I'm now seeing this message as the recommended correction:
```
  ✗ Missing or incorrect header. Should be:                                                                                                                                                                                                                                                           
// Copyright \(c\) ((\d{4})-)?2026 SiFive, Inc\. All rights reserved\.\n// Licensed under the MIT License\.\n// See LICENSE file in the project root for full license information\.\n// SPDX-License-Identifier: MIT\n
```
This is a bit hard to read, and ideally the user should be able to just copy-paste this text directly into their file with minimal changes to make the script happy. Could we prettify the message a bit by:
1. interpret the line breaks to split the message to multiple lines, as it would be written in the file.
2. replace the `((\d{4})-)?` with something more user friendly. Maybe just omit it?
3. remove the other `\` characters.
4. since there's not a unique acceptable message any more, maybe rephrase "Should be" as something like "Recommendation"?

.github/check_license_headers.py
```diff
@@ -0,0 +1,235 @@
...
+    1 - One or more files are missing headers
+"""
+
+import argparse
+import re
+import subprocess
+import sys
+from datetime import datetime
+from pathlib import Path
+from typing import List, Tuple
```
Aaron-Hutchinson: Looks like Tuple is unused now.

.github/check_license_headers.py
myeh01: I have updated the output to be more readable. The year part is printed out as (YYYY-){year_modified} so users can modify the start year or omit it accordingly.

Aaron-Hutchinson: Nice! Looks much better!

Aaron-Hutchinson (approved):
I think the current version is good enough to enable in CI. Thanks for going through many iterations of revising this @myeh01!

myeh01:
Forgot to limit the check to pre-merge only. Fixed here: https://github.com/sifive/skl/pull/120

PR #94, opened 2026-05-07
[Test] Port FP16 exponential test harness

hdelassus:
This series of commits ports the legacy FP16 exponential test harness to the new test driver.


Aaron-Hutchinson (commented):
Thanks for doing these conversions! Since this looks pretty close to the fp32 harness that was already merged I don't have much to say from a high level. I'll let @ebavier approve of the exp details.

test/elementwise/unary_f16.h
```diff
@@ -0,0 +1,80 @@
...
+  } ctx;
+} unary_f16_t;
+
+void unary_f16_init(skl_test_t *t);
+void unary_f16_execute(skl_test_t *t);
+void unary_f16_verify(skl_test_t *t);
+void unary_f16_report(skl_test_t *t);
+void unary_f16_cleanup(skl_test_t *t);
+
+#define UNARY_FP16_TEST_DEFAULTS .in = {.len = 65536, .mode = SKL_TEST_RANDOM}
```
ebavier: There are 63,490 non-NaN f16 values? Even fewer in the `exp` domain defined here (38,530).  Do we want to test all of them in `[min;max]`?  Possibly `SKL_TEST_SEQ` could be used if `init` punned the array type to `int16_t`?  With `SKL_TEST_RANDOM`, its likely to have many values tested repeatedly and many not tested at all.

hdelassus: Haha, yes, that was sloppy of me. I wanted to use `SKL_TEST_SEQ` and test everything since it's quite cheap with 16-bit types, but as you noted we'd need to type-pun the array type. Let me see what I can do.

hdelassus: I've reduced the length to 16384 for now, and we can implement exhaustive testing for all unary functions as part of another PR.

test/exp/exp_f16_xsfvfexpa_zvfh.c
```diff
@@ -0,0 +1,42 @@
...
+unary_f16_t tests[] = {
+#if defined(SKL_ENABLE_BENCHMARKS)
+    FUNCTION_BENCHMARKS(skl_exp_1u_f16_xsfvfexpa_zvfh, -1, 1),
+    FUNCTION_BENCHMARKS(skl_exp_1p132ugen37P3s0_f16_xsfvfexpa_zvfh, -1, 1),
+#endif
+#if defined(SKL_ENABLE_TESTS)
+    FUNCTION_TESTS(skl_exp_1u_f16_xsfvfexpa_zvfh, skl_exp_f16_ref, MIN, MAX,
+                   1.0f),
+    FUNCTION_TESTS(skl_exp_1p132ugen37P3s0_f16_xsfvfexpa_zvfh, skl_exp_f16_ref,
+                   -0x1.33p3f, MAX, 2.0f),
```
ebavier:
```suggestion
                   -0x1.37p3f, MAX, 2.0f),
```

ebavier (dismissed):
LGTM, thanks!

test/exp/exp_f16_xsfvfexpa_zvfh.c
hdelassus: Good catch!

ebavier (approved):

PR #95, opened 2026-05-12
Improve and fix CMake infrastructure for cross-compiling and linting

hdelassus:
This PR fixes minor issues with our build infrastructure:

* `CMAKE_SYSTEM_NAME` was never set when it actually should be set when cross-compiling. I've set it to `Generic` as a first step.
* Move the `SKL_ENABLE_CLANG_TIDY` option and its related setting from the `cmake/riscv.cmake` toolchain file to the project's `CMakeLists.txt`, because it could cause configure-time errors when CMake checks for a working C compiler for two main reasons: either CMake could not find a clang-tidy config file (out-of-tree build scenario), or CMake's own C source files used for the check did not conform to our linter rules.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,6 +5,9 @@
 
 cmake_minimum_required(VERSION 3.21)
 
+# This option may affect the call to `project` and thus must come first.
+option(SKL_ENABLE_CLANG_TIDY "Enable linting with clang-tidy" ON)
+
 project(SKL)
 
 set(DEFAULT_CMAKE_BUILD_TYPE "Release")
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -13,6 +13,12 @@ if(NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES)
   set(CMAKE_BUILD_TYPE "${DEFAULT_CMAKE_BUILD_TYPE}" CACHE STRING "Build type (default ${DEFAULT_CMAKE_BUILD_TYPE})" FORCE)
 endif()
 
+# Set up clang-tidy
+option(SKL_ENABLE_CLANG_TIDY "Enable clang-tidy command" ON)
+if(SKL_ENABLE_CLANG_TIDY)
+    set(CMAKE_C_CLANG_TIDY riscv64-unknown-elf-clang-tidy)
```
Aaron-Hutchinson: Prior to this PR the `CMAKE_C_CLANG_TIDY` value was configurable by providing a custom toolchain file, but here it doesn't seem to be overridable. Could we restore the prior configurability somehow?

Aaron-Hutchinson (approved):
Seems good to me.

PR #96, opened 2026-05-12
[GEMM] Optimize f32 small N x390 kernel

Aaron-Hutchinson:
This rewrites the RVV f32 "small N" kernel for x390 to provide performance improvements:

- Change from LMUL=1 to LMUL=2. This provides better performance for N values between `vsetvlmax_e32m1()` and `vsetvlmax_e32m2()`.
- Changed tile size to `8 x m2 x 12` and structured the `jj` and `kk` loops similar to that of #57 and #75 by preloading the A and B values in as far advance as possible.
- Use `ntl.p1` hints similar to other kernels.

Compared to the original version, I measured this revised kernel achieving more than a 200% increase in performance on certain problem sizes.


ericlove (approved):
Seems like a good improvement to me, and I agree that for x390 it makes more sense to put effort into the LMUL=2 case, as LMUL=1 just can't be helped that much.

PR #97, opened 2026-05-12
[Test][GEMM] Update f8 GEMM tests

myeh01:
This PR updates the f8e4m3 GEMM tests to use the new test harness format.


test/gemm/gemm_f8e4m3rcprc_f8e4m3rcprc_f32rcprc.h
myeh01: Check headers

test/gemm/xsfmm/skl_gemm_a1b01_f8e4m3c_f8e4m3_f32_xsfmm32a8f.c
```diff
@@ -3,21 +3,21 @@
 // See LICENSE file in the project root for full license information.
 // SPDX-License-Identifier: MIT
 
-#include "gemm/gemm_f16rcprc_f16rcprc_f32rcprc.h"
+#include "gemm/gemm_f8e4m3rcprc_f8e4m3rcprc_f32rcprc.h"
 #include "skl-test-driver.h"
 #include "skl.h"
 #include <stdbool.h>
```
myeh01: Check headers

test/gemm/gemm_f32rcprc_f32rcprc_f32rcprc.c
```diff
@@ -95,7 +95,6 @@ void gemm_f32rcprc_f32rcprc_f32rcprc_init(skl_test_t *t) {
 }
 
 void gemm_f32rcprc_f32rcprc_f32rcprc_verify(skl_test_t *t) {
-  /* Compute the reference matrix output. */
```
myeh01, line 98: Revert this and do in follow-up PR.

Aaron-Hutchinson:
What do you think about using a single harness for both f8 types (from this PR and #99)? It looks like the struct definitions are identical, and the harness only relies on the specific format in a few places. Seems like merging the two would amount to adding something like:
```
typedef enum { F8E4M3, F8E5M2 } ofp8_t;
```
and branching in a few places in the harness. cvt did something similar: https://github.com/sifive/skl/blob/3e16028c941c72891defed1e47ff5fdc08055329/test/cvt/cvt_f8_bf16.h#L24

myeh01:
> What do you think about using a single harness for both f8 types (from this PR and #99)? It looks like the struct definitions are identical, and the harness only relies on the specific format in a few places. Seems like merging the two would amount to adding something like:
> 
> ```
> typedef enum { F8E4M3, F8E5M2 } ofp8_t;
> ```
> 
> and branching in a few places in the harness. cvt did something similar:
> 
> https://github.com/sifive/skl/blob/3e16028c941c72891defed1e47ff5fdc08055329/test/cvt/cvt_f8_bf16.h#L24

Thanks for pointing this out! I think was able to combine them pretty well. Let me know what you think, and I'll close #99.

test/gemm/gemm_f8rcprc_f8rcprc_f32rcprc.h
```diff
@@ -28,10 +28,11 @@
 #pragma once
 
 #include "skl-test-driver.h"
-#include <stdbool.h>
 #include <stddef.h>
 #include <stdint.h>
 
+typedef enum { F8E4M3, F8E5M2 } gemm_ofp8_type_t;
```
Aaron-Hutchinson, line 34: No need to fix this in this PR, but since this enum can be used in any f8 test (like the existing cvt tests, CC @Harlen126) we should probably move it to a broader header. I can handle this as part of some future cleanup.

Aaron-Hutchinson (dismissed):
Looks great! Thanks for figuring out how to handle both the f8 formats. Glad we can avoid separate harnesses using pretty minimal adjustments here.

myeh01:
@Aaron-Hutchinson Since this PR updates `SKL_TEST_BUF_CREATE`, many of the tests now trigger the cognitive complexity warning. I think we agreed to temporarily suppress these warnings and rewrite the macro in a future PR, is that correct?

Aaron-Hutchinson:
When it was discussed internally I didn't hear any protest against it, so I think let's make that the plan :+1:

myeh01:
Need a final approval

Aaron-Hutchinson (approved):

PR #98, opened 2026-05-13
[Pack] Add RVV-based 8-bit Matrix Pack Kernel

Harlen126:
## Summary

This PR implements a RVV-based matrix pack kernel for 8-bit integer matrices, transforming 2D row/column-major layouts into blocked 4D layouts suitable for optimized computation kernels (e.g., GEMM).

**Function:** `skl_pack_e8rc_e8rcprc_zve32x`

###  Vectorized Path
- **Optimized when both input and output are in contiguous memory**
- Leverages RVV intrinsics for data movement
- Automatic padding with configurable padding values

###  Scalar Fallback
- Scalar implementation for non-contiguous memory layouts
- Handles arbitrary stride configurations
- Ensures correctness across all edge cases

###  Vectorization Strategy

**Internal Function:** `skl_pack_e8_e8prcbrc_zve32x` (constraint: `cs == 1`)
When `rs0 == 1` : Use transpose-based vectorization => Utilizes RVV segment loads/stores for efficient data movement
When `cs0 == 1`:  Use copy-based vectorization => Utilizes RVV unit-stride loads/stores for efficient data movement

### Helper Functions

- `skl_set_e8_zve32x`: Vectorized element fill (replaces memset semantics)
- `skl_copy_e8_zve32x`: Vectorized element copy (replaces memcpy semantics)
- `skl_set2d_e8_zve32x`: Vectorized element fill for a 2D subarray
- `skl_copy2d_e8_zve32x`: Vectorized element copy for a 2D subarray
- `skl_transpose_padded_m_e8_zve32x`: Transpose with M-dimension padding
- `skl_transpose_e8_zve32x`: **Optimized matrix transpose (will be made public and replace the current transpose kernel)**

## TODO / Future Work

**High Priority:**
- Implement unpack kernel (inverse operation: blocked 4D → 2D)
- Implement e16/e32 version
- Make `skl_transpose_e8_zve32x` public and replace current transpose kernel

**Low Priority / Won't Do:**
- Hybrid vectorization paths (limited performance benefit vs scalar fallback):
  - Indexed load + unit-stride store (input non-contiguous, output contiguous)
  - Unit-stride load + indexed store (input contiguous, output non-contiguous)


src/pack/rvv/skl_pack_e8_zve32x.c
```diff
@@ -0,0 +1,1067 @@
...
+
+  // Handle division by zero
+  if (m0 == 0 || n0 == 0) {
+    return;
+  }
+
+  size_t m1 = (m + m0 - 1) / m0; // Num. row blocks in the input matrix
+  size_t n1 = (n + n0 - 1) / n0; // Num. column blocks in the input matrix
+
+  if ((cs0 * n0 == cs1) && rs0 == 1) {
```
ericlove, line 1016: Can you explain what special case this is handling that wouldn't be covered by the loop nest below? I have looked at it for a bit and can't tell.

src/pack/rvv/skl_pack_e8_zve32x.c
```diff
@@ -0,0 +1,1067 @@
...
+
+/* Transposes an M x N row-major matrix (pointed by `a`) to an N x M row-major
+ * matrix (pointed by `at`) with vectorization along the M dimension.
+ */
+SKL_FUNC_PRIVATE void
+skl_transpose_mvec_e8_zve32x(size_t m, size_t n, const uint8_t *SKL_RESTRICT a,
+                             size_t rsa, uint8_t *SKL_RESTRICT at,
+                             size_t rsat) {
+
+  skl_transpose_mvec_x8_e8_zve32x(m, n / 8 * 8, a, rsa, at, rsat);
```
ericlove: Why does this one case get its own separate function, while the others are handled inline here?

test/pack/rvv/skl_pack_e8_zve32x.c
```diff
@@ -0,0 +1,130 @@
...
+  {TEST, .m = 32, .n = 8, .m0 = 8, .n0 = 8, .rs = 8, .cs = 1,
+   .rs0 = 1, .cs0 = 8, .rs1 = 64, .cs1 = 64},
+  {TEST, .m = 8, .n = 32, .m0 = 8, .n0 = 8, .rs = 32, .cs = 1,
+   .rs0 = 1, .cs0 = 8, .rs1 = 256, .cs1 = 64},
+
+  /* Larger matrices */
+  {TEST, .m = 64, .n = 64, .m0 = 8, .n0 = 8, .rs = 64, .cs = 1,
+   .rs0 = 1, .cs0 = 8, .rs1 = 512, .cs1 = 64},
+  {TEST, .m = 128, .n = 128, .m0 = 8, .n0 = 8, .rs = 128, .cs = 1,
+   .rs0 = 1, .cs0 = 8, .rs1 = 1024, .cs1 = 64},
```
ericlove, line 141: We should probably add a few tests with larger matrices that also have odd dimensions like above, so we can test a combination of full vectors and fringe handling.

src/pack/rvv/skl_pack_e8_zve32x.h
```diff
@@ -0,0 +1,55 @@
...
+ *
+ * The output layout consists of m1 × n1 blocks, where:
+ *   - m1 = ceil(m / m0) = (m + m0 - 1) / m0  (number of row blocks)
+ *   - n1 = ceil(n / n0) = (n + n0 - 1) / n0  (number of column blocks)
+ *
+ * Each block has dimensions m0 × n0. If m is not a multiple of m0 or n is not
+ * a multiple of n0, the incomplete blocks are padded with the specified padding
+ * value.
+ */
+void skl_pack_e8rc_e8prcbrc_zve32x(size_t m, size_t n, const uint8_t *src,
```
ericlove: Should we add `SKL_RESTRICT` on the API pointers so that it's clear there cannot be overlap between the two buffers?  Ad we should mention this in the documentation too.

ericlove (commented):
I like this a lot! It's a great contribution to the library.  I think the overall interface and design is good, and the implementation is too.

The case that transposes data within a block seems to me about as complete as can be, but I suspect we will eventually want to add some specialized variants of the non-transposed ("copy2d") cases.  There will be a lot of use for that, and we should confirm with internal benchmarking that performance meets expectations.

src/pack/README.md
```diff
@@ -173,28 +174,28 @@ It is for illustrative purposes only.
...
+- `skl_pack_e8_e8ptex1c_xsfmm`: Transpose (TE x K) panels using the matrix engine. Used on a row-major A matrix or column-major B matrix.
 
 ### Packing Kernels That Do Not Exist
 These names refer to kernels that users might expect to exist, but do not, because they can be implemented with an appropriate block size in the generic RVV packing kernel at reasonable performance.
 In general, so long as the block size is large enough in at least one dimension (as is the case in paneling), a generic packing kernel with a large enough vector length can achieve reasonable performance.
 
-- ~~`skl_pack_1xte_e8_zve32x`~~: Pack (K x TE) panels. Used on a column-major A matrix or row-major B matrix.
-- ~~`skl_pack_texte_e32_zve32x`~~: Pack (TE x TE) blocks without transposition. Used on a C matrix.
-- ~~`skl_pack_1xvlm8_e8_zve32x`~~: Pack (1 x VLMAX*8) column panels.
+- ~~`skl_pack_e8_e8p_1xte_zve32x`~~: Pack (K x TE) panels. Used on a column-major A matrix or row-major B matrix.
```
peichialin: skl_pack_e8_e8p1xte_zve32x

src/pack/rvv/skl_pack_e8_zve32x.c
```diff
@@ -0,0 +1,1067 @@
...
+
+      in_tile += vl * rsa;
+      out_tile += vl;
+    }
+    break;
+  }
+  default:
+    break;
+  }
+}
```
peichialin, line 440: @Harlen126 It looks like you optimize the remaning m by 7 ... 1 and abandon the decompsition to the combination of power of 2. Am I right?

src/pack/rvv/skl_pack_e8_zve32x.c
Harlen126: For `m0xn0c` satisfying `cs0 * n0 == cs1`, the layout can be reinterpreted as `m0x(n0xn1)c`, allowing us to handle `m0xn` data at once instead of `m0xn0` at a time.

I also realized that I missed handling the `1xn0` and `m0x1` specializations.

src/pack/rvv/skl_pack_e8_zve32x.c
Harlen126: The existing transpose kernel uses `skl_transpose_mvec_x8_e8_zve32x()` as a separate function. However, you are right—since I have removed the "x4", "x2", and "x1" versions, I should inline the "x8" version as well.

src/pack/rvv/skl_pack_e8_zve32x.h
Harlen126: Good suggestion! We may also need to review the use of `SKL_RESTRICT` across the other kernels.

src/pack/rvv/skl_pack_e8_zve32x.c
Harlen126: Exactly~

src/pack/rvv/skl_pack_e8_zve32x.c
Harlen126: Added the other specializations in [e53eebf](https://github.com/sifive/skl/pull/98/commits/e53eebf53080bb4a77f9fc13e9e1ccf3b90791b3)

test/pack/rvv/skl_pack_e8_zve32x.c
Harlen126: Added more tests in [2ad7725](https://github.com/sifive/skl/pull/98/commits/2ad7725ebb27d42e0290f56666b430423b541b26)

src/pack/rvv/skl_pack_e8_zve32x.c
```diff
@@ -0,0 +1,1072 @@
...
+        vl = __riscv_vsetvl_e8m1(m - jj);
+
+        vcols = __riscv_vlsseg8e8_v_u8m1x8(in_tile, input_seg_bstride, vl);
+
+        uint8_t *write_ptr = out_tile;
+        const size_t output_seg_stride = rsat;
+        // store each segment continuously along m
+        __riscv_vse8_v_u8m1(write_ptr, __riscv_vget_v_u8m1x8_u8m1(vcols, 0),
+                            vl);
+        write_ptr += output_seg_stride;
```
peichialin: Let's replace `output_seg_stride` by `rsat` and remove `output_seg_stride` to reduce LOC.

src/pack/rvv/skl_pack_e8_zve32x.c
```diff
@@ -0,0 +1,1072 @@
...
+    vuint8m2_t vrow1 = __riscv_vundefined_u8m2();
+    vuint8m2_t vrow2 = __riscv_vundefined_u8m2();
+    // NOLINTEND(clang-analyzer-deadcode.DeadStores)
+
+    for (size_t jj = 0; jj < n; jj += vl) {
+      vl = __riscv_vsetvl_e8m2(n - jj);
+
+      vrow0 = __riscv_vle8_v_u8m2(in_tile, vl);
+      vrow1 = __riscv_vle8_v_u8m2(in_tile + rsa, vl);
+      vrow2 = __riscv_vle8_v_u8m2(in_tile + 2 * rsa, vl);
```
peichialin: We can reduce the number of multiplications by incrementing the pointer by `rsa`.

src/pack/rvv/skl_pack_e8_zve32x.c
```diff
@@ -0,0 +1,1072 @@
...
+  if (skl_transpose_e8_is_mvec(m, n)) {
+    skl_transpose_mvec_e8_zve32x(m, n, a, rsa, at, rsat);
+  } else {
+    skl_transpose_nvec_e8_zve32x(m, n, a, rsa, at, rsat);
+  }
+}
+
+SKL_FUNC_PRIVATE void skl_pad_e8_zve32x(uint8_t *dst, size_t extent,
+                                        size_t stride, uint8_t padding_value,
+                                        size_t n_segs) {
```
peichialin: Please add a short description paragraph to explain the definition of the parameters.
I'm wondering `n_segs` should be renamed.

src/pack/rvv/skl_pack_e8_zve32x.c
```diff
@@ -0,0 +1,1072 @@
...
+      }
+      break;
+    }
+    default:
+      break;
+    }
+  } else {
+    for (size_t i = 0; i < n_segs; ++i) {
+      skl_set_e8_zve32x(dst, padding_value, extent);
+      dst += stride;
```
peichialin: To my understanding, the implementation of extent <=8 won't increase the element counts stored per cycle theoritically (extent / cycle) compared to `skl_set_e8_zve32x` (when extent < DLEN => throughput = extent / cycle) but it effecitvely reduce the vector instruction count. Am I right?

src/pack/rvv/skl_pack_e8_zve32x.c
```diff
@@ -0,0 +1,1072 @@
...
+    }
+  }
+}
+
+SKL_FUNC_PRIVATE void skl_transpose_padded_m_e8_zve32x(
+    size_t m_padded, size_t m, size_t n, const uint8_t *SKL_RESTRICT a,
+    size_t rsa, uint8_t *SKL_RESTRICT at, size_t rsat, uint8_t padding_value) {
```
peichialin: Please add a short description paragraph to explain the definition of the parameters.

src/pack/rvv/skl_pack_e8_zve32x.c
```diff
@@ -0,0 +1,1072 @@
...
+
+  if (m_padded == m) {
+    skl_transpose_e8_zve32x(m, n, a, rsa, at, rsat);
+    return;
+  }
+
+  if (m / 8) {
```
peichialin: (m > 8) is more straightforward.

src/pack/rvv/skl_pack_e8_zve32x.c
```diff
@@ -0,0 +1,1072 @@
...
+    skl_transpose_e8_zve32x(m / 8 * 8, n, a, rsa, at, rsat);
+  }
```
peichialin, line 733: We can save the value of (m / 8 * 8) and (m % 8) for reuse.

src/pack/rvv/skl_pack_e8_zve32x.c
```diff
@@ -0,0 +1,1072 @@
...
+  const uint8_t *in_tile = a + m / 8 * 8 * rsa;
+  uint8_t *out_tile = at + m / 8 * 8;
+  if (m % 8) {
+    size_t m_transition = m_padded - m / 8 * 8 > 8 ? 8 : m_padded - m / 8 * 8;
```
peichialin:
```
/* m_padded - m >= 1 => m_padded - (m / 8 * 8 + r) >= 1 =>  m_padded - (m / 8 * 8 ) >= (1 + r)  */
/* r = m % 8 >= 1 =>  m_padded - (m / 8 * 8 ) >= (1 + r) >= 2 */
```

src/pack/rvv/skl_pack_e8_zve32x.c
```diff
@@ -0,0 +1,1072 @@
...
+      for (size_t vl = 0, avl = n; avl > 0; avl -= vl) {
+        vl = __riscv_vsetvl_e8m2(avl);
+        switch (m % 8) {
+        case 2:
+          pad_vec_group3 = __riscv_vset_v_u8m2_u8m2x3(
+              pad_vec_group3, 1, __riscv_vle8_v_u8m2(in_tile + 1 * rsa, vl));
+          __attribute__((fallthrough));
+        case 1:
+          pad_vec_group3 = __riscv_vset_v_u8m2_u8m2x3(
+              pad_vec_group3, 0, __riscv_vle8_v_u8m2(in_tile, vl));
```
peichialin, line 773: Will compiler derive additional vmv instructions to form a continuous vector register group for this sequence of intrinsics?

src/pack/rvv/skl_pack_e8_zve32x.c
```diff
@@ -0,0 +1,1072 @@
...
+
+  // Handle division by zero
+  if (m0 == 0 || n0 == 0) {
+    return;
+  }
+
+  size_t m1 = (m + m0 - 1) / m0; // Num. row blocks in the input matrix
+  size_t n1 = (n + n0 - 1) / n0; // Num. column blocks in the input matrix
+
+  if ((cs0 * n0 == cs1) && rs0 == 1) {
```
peichialin, line 1016: We can add comments like / * block itself: column-major, between blocks: row-major */

src/pack/rvv/skl_pack_e8_zve32x.c
```diff
@@ -0,0 +1,1072 @@
...
+    for (size_t ii1 = 0; ii1 < m1; ++ii1) {
+      const uint8_t *src_block = src + ii1 * m0 * rs;
+      uint8_t *dst_block = dst + ii1 * rs1;
+      size_t m_length = m0 < m - ii1 * m0 ? m0 : m - ii1 * m0;
+      skl_transpose_padded_m_e8_zve32x(m0, m_length, n, src_block, rs,
+                                       dst_block, cs0, padding_value);
+      if (n % n0) {
+        // pad right
```
peichialin, line 1024: The pad "right" is in source's view, which is equivalent to pad "bottom" in the destination.
That's why we set the "extent" to `m0`, and "n_seg" to `n0 - n % n0`, am I right?

src/pack/rvv/skl_pack_e8_zve32x.c
```diff
@@ -0,0 +1,1072 @@
...
+        skl_pad_e8_zve32x(dst_block + cs1 * (n1 - 1) + cs0 * (n % n0), m0, cs0,
+                          padding_value, n0 - n % n0);
+      }
+    }
+    return;
+  }
+
+  if ((rs0 * m0 == rs1) && cs0 == 1) {
```
peichialin, line 1033: We can add comments like / * block itself: row-major, between blocks: column-major */

src/pack/rvv/skl_pack_e8_zve32x.c
peichialin: Shall we rename the function to `skl_pad2d_e8_zve32x`?

src/pack/rvv/skl_pack_e8_zve32x.c
Harlen126: I noticed that the compiler uses `vmv.v.x` to fill `v12` first. Then, depending on the condition, it branches to a certain section that uses `vmv1r` or `vmv2r` to fill the vector registers.
For example, in the `u8m2x3` case, it uses `vmv1r` to fill `v13`, and then uses `vmv2r` to fill `v10` and `v14`.

Harlen126: I'll rewrite it to use 1 single m8 `vmv.v.x`.

src/pack/rvv/skl_pack_e8_zve32x.c
Harlen126: There is no `skl_pad1d_e8_zve32x` since its purpose overlaps with `skl_set_e8_zve32x`.
I am not sure whether to treat pad as a 2D operation (just like transpose, which doesn't contain 2d in the name) or treat it as an N-D operation.

src/pack/rvv/skl_pack_e8_zve32x.c
Harlen126: Yes, it's "**vl**" individual `vle8(..., segN)` instructions vs. just "**one**" `vsssegNe8(...,vl)` instruction.

src/pack/rvv/skl_pack_e8_zve32x.c
Harlen126: Yeah~
I use the same illustration in the [README](https://github.com/sifive/skl/blob/dev/hchen/pack-impl/src/pack/README.md).

src/pack/rvv/skl_pack_e8_zve32x.c
Harlen126: Done~
[c494b07](https://github.com/sifive/skl/pull/98/commits/c494b07cc1e1841064993ef47d12827a7729ad30)

src/pack/rvv/skl_pack_e8_zve32x.c
Harlen126: Done~
[c494b07](https://github.com/sifive/skl/pull/98/commits/c494b07cc1e1841064993ef47d12827a7729ad30)

src/pack/rvv/skl_pack_e8_zve32x.c
Harlen126: Refactor in [c494b07](https://github.com/sifive/skl/pull/98/commits/c494b07cc1e1841064993ef47d12827a7729ad30)

src/pack/rvv/skl_pack_e8_zve32x.c
Harlen126: Refactor in [c494b07](https://github.com/sifive/skl/pull/98/commits/c494b07cc1e1841064993ef47d12827a7729ad30)

src/pack/rvv/skl_pack_e8_zve32x.c
Harlen126: Add explanation in [c494b07](https://github.com/sifive/skl/pull/98/commits/c494b07cc1e1841064993ef47d12827a7729ad30)

src/pack/rvv/skl_pack_e8_zve32x.c
Harlen126: > I'll rewrite it to use 1 single m8 vmv.v.x.

Updated in [8e5a9ed](https://github.com/sifive/skl/pull/98/commits/8e5a9eda33fbdd7b72936d4dc2edc2a2afa2b2d9).

src/pack/rvv/skl_pack_e8_zve32x.c
Harlen126: > Please add a short description paragraph to explain the definition of the parameters.
I'm wondering n_segs should be renamed.

Updated in [0ef0cd4](https://github.com/sifive/skl/pull/98/commits/0ef0cd4c63b610a3db824f9d74bde52a3888aa60)

src/pack/rvv/skl_pack_e8_zve32x.c
Harlen126: Updated in [0ef0cd4](https://github.com/sifive/skl/pull/98/commits/0ef0cd4c63b610a3db824f9d74bde52a3888aa60)

src/pack/rvv/skl_pack_e8_zve32x.c
Harlen126: Updated in [0ef0cd4](https://github.com/sifive/skl/pull/98/commits/0ef0cd4c63b610a3db824f9d74bde52a3888aa60)

src/pack/rvv/skl_pack_e8_zve32x.c
Harlen126: Updated in [0ef0cd4](https://github.com/sifive/skl/pull/98/commits/0ef0cd4c63b610a3db824f9d74bde52a3888aa60)

src/pack/rvv/skl_pack_e8_zve32x.c
```diff
@@ -0,0 +1,1141 @@
...
+ * @param pad - Value to use for padding.
+ * @param n_segments - Number of segments to fill.
+ *
+ * This function fills each segment with the padding value. The segments are
+ * stored contiguously in memory with a stride in elements between the
+ * start of each segment.
+ */
+SKL_FUNC_PRIVATE void skl_pad_e8_zve32x(uint8_t *dst, size_t extent,
+                                        size_t stride, uint8_t pad,
+                                        size_t n_segments) {
```
peichialin: After rethinking, I propose to revise the API signature as below, that will be more consistent with other SKL kernel's naming convention and become the family of "set".
Naming with "pad" makes more sense in the case where the output contains part of the input and the extended padding region.
Besides, the strided segment store implementation is one of the solutions, so we should avoid naming the parameters with this behavior.
```
SKL_FUNC_PRIVATE void skl_set_2d_e8_zve32x(uint8_t *dst, size_t rsd, uint8_t value, size_t m, size_t n);

strides --> rsd /* row stride of dest */
pad --> value /* align the name in skl_set_e8_zve32x */
n_segments --> m /* rows */
extent --> n /* cols */
```

Harlen126:
@myeh01 can you help review the test harness part?
Any feedback would be greatly appreciated!

src/pack/rvv/skl_pack_e8_zve32x.c
Harlen126: Thanks for the suggestion! I've renamed it to `skl_set2d_e8_zve32x`.
Let's see if the team prefers `set_2d` or `set2d`.

test/pack/pack_e8.h
```diff
@@ -0,0 +1,49 @@
+// Copyright (c) 2026 SiFive, Inc. All rights reserved.
+// Licensed under the MIT License.
+// See LICENSE file in the project root for full license information.
+// SPDX-License-Identifier: MIT
+
+#ifndef SKL_TEST_PACK_PACK_E8_H_
+#define SKL_TEST_PACK_PACK_E8_H_
```
myeh01: Do we need a header guard? I think we've just been using `#pragma once` in our headers.

test/pack/pack_e8.h
```diff
@@ -0,0 +1,49 @@
...
+  // Test function pointers for various steps
+  // *** This field must be placed first within this struct ***
+  skl_test_steps_t steps;
+
+  // Input matrix dimensions
+  size_t m, n;
+  // Block dimensions
+  size_t m0, n0;
+  // Input matrix strides
+  size_t rs, cs;
```
myeh01:
```suggestion
  // Input matrix strides
  size_t rs, cs;
  // Block dimensions
  size_t m0, n0;
```
Very nit: could we group the input arguments together?

test/pack/pack_e8rc_e8rcprc.h
```diff
@@ -0,0 +1,49 @@
+// Copyright (c) 2026 SiFive, Inc. All rights reserved.
```
myeh01, line 1: Should we rename the test harness from to `pack_e8` to something like `pack_e8_e8rcprc` (similar to how the GEMM harnesses are named `gemm_f32rcprc_f32rcprc_f32rcprc`, for example)?

test/pack/pack_e8.h
```diff
@@ -0,0 +1,49 @@
...
+  size_t rs1, cs1; // between blocks
+
+  uint8_t pad;
+  // Buffer generation settings
+  SKL_TEST_BUFFER(uint8_t) src, dst;
+
+  // Derived parameters & buffers (private to the test harness)
+  struct {
+    uint8_t *ref_dst;
+    size_t dst_size;
```
myeh01: Is `dst_size` necessary, or can we just use `dst.len`?

test/pack/pack_e8.c
```diff
@@ -0,0 +1,158 @@
...
+#include <stddef.h>
+#include <stdint.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+void pack_e8_init(skl_test_t *t) {
+  pack_e8_t *h = (pack_e8_t *)t->harness;
+
+  // Calculate source buffer size
```
myeh01: Maybe we could imitate the GEMM test harness `init` code here. `skl_test_check_matrix_params_rcprc` can be used to check that the dimensions and strides of a packed matrix are valid. It can be used for the input matrix too if it is viewed as a packed matrix with 1x1 blocks.

The buffer length can also be set with something like
```
  if (m1 == 0 || n1 == 0) {
    h->dst.len = 0;
  } else {
    h->dst.len = (m1 - 1) * rs1 + (n1 - 1) * cs1 + (m0 - 1) * rs0 + (n0 - 1) * cs0 + 1;
  }
```
For the row-major input matrix, maybe the formula in the `else` branch could be something like `(m - 1) * rs + (n - 1) * cs + 1`.

What do you think?

Edit: if this looks okay to you, then we might need to move some of the functions from `skl_test_gemm.h` to `skl-test.h`.

test/pack/pack_e8.c
```diff
@@ -0,0 +1,158 @@
...
+    if (dst[i] != ref_dst[i]) {
+      SKL_TEST_LOG(t, SKL_TEST_LOG_ERROR, "position [%zu]: %hhu != ref %hhu\n",
+                   i, dst[i], ref_dst[i]);
+      t->status.verify_status = SKL_TEST_FAIL;
+      return;
+    }
+  }
+}
+
+void pack_e8_param_report(skl_test_t *t) {
```
myeh01: Is it possible to put some of this reporting code for parameters and benchmark performance into a header so it can be reused in the e16 and e32 packing test harnesses, similar to what we did for the GEMM harnesses? https://github.com/sifive/skl/blob/7431fa1a3091ed514d910b06025e31f0f872fcc4/test/gemm/skl_test_gemm.h#L249-L286

test/pack/pack_e8.c
```diff
@@ -0,0 +1,158 @@
...
+
+#define INFO(fmt, ...) SKL_TEST_LOG(t, SKL_TEST_LOG_INFO, fmt, __VA_ARGS__)
+  size_t m1 = (h->m + h->m0 - 1) / h->m0;
+  size_t n1 = (h->n + h->n0 - 1) / h->n0;
+  INFO("M: %zd, N: %zd\n", h->m, h->n);
+  INFO("M0: %zd, N0: %zd\n", h->m0, h->n0);
+  INFO("M1: %zd, N1: %zd\n", m1, n1);
+  INFO("RS: %zd, CS: %zd\n", h->rs, h->cs);
+  INFO("RS0: %zd, CS0: %zd\n", h->rs0, h->cs0);
+  INFO("RS1: %zd, CS1: %zd\n", h->rs1, h->cs1);
```
myeh01: Would it be better to use `%zu` since these variables are unsigned?

test/pack/pack_e8.c
```diff
@@ -0,0 +1,158 @@
...
+  } else if (h->cs == 1) {
+    if (h->rs0 == 1) {
+      INFO("%s", "=> TRANSPOSE case!\n");
+    } else if (h->cs0 == 1) {
+      INFO("%s", "=> COPY case!\n");
+    } else {
+      INFO("%s", "=> GENERAL case!\n");
+    }
+  }
+#undef INFO
```
myeh01: Should this have a final `else` for the case when neither `rs` nor `cs` is 1 ? I guess it would be the general case?

test/pack/pack_e8.c
```diff
@@ -0,0 +1,158 @@
...
+}
+
+void pack_e8_bench_report(skl_test_t *t) {
+
+  pack_e8_param_report(t);
+  pack_e8_t *h = (pack_e8_t *)t->harness;
+#define INFO(fmt, ...) SKL_TEST_LOG(t, SKL_TEST_LOG_INFO, fmt, __VA_ARGS__)
+  INFO("Warmup: %s\n", h->steps.warmup ? "yes" : "no");
+  INFO("Cycles: %zd\n", t->counters.cycles);
+  INFO("Instructions: %zd\n", t->counters.instret);
```
myeh01: Should we use `PRIu64` to print out the counters since they are `uint64_t` type?

test/pack/pack_e8.c
myeh01: Maybe it is ok to keep `INFO` defined until the end of `pack_e8_bench_report`.

test/pack/pack_e8rc_e8rcprc.h
```diff
@@ -0,0 +1,49 @@
...
+  // Derived parameters & buffers (private to the test harness)
+  struct {
+    uint8_t *ref_dst;
+    size_t dst_size;
+  } ctx;
+} pack_e8_t;
+
+#define PACK_E8_DEFAULTS                                                       \
+  .src = {.min = 0, .max = 255, .mode = SKL_TEST_RANDOM},                      \
+  .dst = {.min = 0, .max = 255, .mode = SKL_TEST_RANDOM}, .pad = 255
```
myeh01, line 51: I don't think it matters much, but is there a reason for using a default padding value of 255 rather than 0?

test/pack/pack_e8.c
myeh01:
```suggestion
void pack_e8_test_report(skl_test_t *t) {
```
Could we rename this so it is similar to the GEMM harnesses?

test/pack/pack_e8.c
```diff
@@ -0,0 +1,158 @@
...
+    } else if (h->cs0 == 1) {
+      INFO("%s", "=> COPY case!\n");
+    } else {
+      INFO("%s", "=> GENERAL case!\n");
+    }
+  }
+#undef INFO
+}
+
+void pack_e8_bench_report(skl_test_t *t) {
```
myeh01:
```suggestion
void pack_e8_benchmark_report(skl_test_t *t) {
```
Could we rename this so it is similar to the GEMM harnesses?

test/pack/rvv/skl_pack_e8_zve32x.c
```diff
@@ -0,0 +1,176 @@
...
+                        .cleanup = pack_e8_cleanup,                            \
+  }
+
+static void execute(skl_test_t *t);
+
+// clang-format off
+pack_e8_t tests[] = {
+#ifdef SKL_ENABLE_BENCHMARKS
+  // Benchmark tests
+  {BENCH, .m = 128, .n = 128, .m0 = 8, .n0 = 8, .rs = 128, .cs = 1,
```
myeh01: It may be easier to read if the input arguments `m`, `n`, `rs`, `cs` are grouped together.

ref/pack/pack_e8rc_e8prcbrc.h
```diff
@@ -0,0 +1,52 @@
...
+ * @param cs0 - Column stride within a block of the output matrix
+ * @param rs1 - Row stride between blocks of the output matrix
+ * @param cs1 - Column stride between blocks of the output matrix
+ * @param pad - Value to insert for padded elements (usually 0)
+ *
+ * Transforms a 2D matrix into a blocked 4D layout.
+ *
+ * The output layout consists of m1 × n1 blocks, where:
+ *   - m1 = ceil(m / m0) = (m + m0 - 1) / m0  (number of row blocks)
+ *   - n1 = ceil(n / n0) = (n + n0 - 1) / n0  (number of column blocks)
```
myeh01:
```suggestion
 *   - m1 = ceil(m / m0) = (m + m0 - 1) / m0  (number of block-rows)
 *   - n1 = ceil(n / n0) = (n + n0 - 1) / n0  (number of block-columns)
```

src/pack/rvv/skl_pack_e8_zve32x.h
```diff
@@ -0,0 +1,58 @@
...
+ * @param cs0 - Column stride within a block of the output matrix
+ * @param rs1 - Row stride between blocks of the output matrix
+ * @param cs1 - Column stride between blocks of the output matrix
+ * @param pad - Value to insert for padded elements (usually 0)
+ *
+ * Transforms a 2D matrix into a blocked 4D layout.
+ *
+ * The output layout consists of m1 × n1 blocks, where:
+ *   - m1 = ceil(m / m0) = (m + m0 - 1) / m0  (number of row blocks)
+ *   - n1 = ceil(n / n0) = (n + n0 - 1) / n0  (number of column blocks)
```
myeh01:
```suggestion
 *   - m1 = ceil(m / m0) = (m + m0 - 1) / m0  (number of block-rows)
 *   - n1 = ceil(n / n0) = (n + n0 - 1) / n0  (number of block-columns)
```

test/pack/pack_e8.c
Harlen126: Thanks for the suggestion, that makes sense. I've moved the checking function to `skl-test-driver.h`.

test/pack/pack_e8.c
Harlen126: I've moved the report function to [skl_test_pack.h.](https://github.com/sifive/skl/blob/dev/hchen/pack-impl/test/pack/skl_test_pack.h)
Thanks for bringing this up.

test/pack/pack_e8rc_e8rcprc.h
Harlen126: It's more likely the residual memory value equals 0, so I used 255 as the padding value.

Harlen126:
@myeh01, thank you for your careful review. I've updated the code based on your feedback. Please let me know if you have any further comments.

peichialin (dismissed):
Overall LGTM. But I'll rely on @myeh01 's review on the details of `skl_test_check_matrix_params_rcprc` proposed in `test/skl-test-driver.h.

test/skl-test-driver.h
```diff
@@ -593,3 +593,57 @@ enum {
       (T)->status.STATUS = SKL_TEST_FAIL;                                      \
     }                                                                          \
   } while (0)
+
+/**
```
myeh01, line 597: I think this should go somewhere in the `Utility functions for use in tests` section (L. 440-L. 549)

test/pack/pack_e8rc_e8rcprc.c
```diff
@@ -0,0 +1,113 @@
...
+  size_t m0 = h->m0;
+  size_t n0 = h->n0;
+  size_t m1 = (m + m0 - 1) / m0;
+  size_t n1 = (n + n0 - 1) / n0;
+  size_t rs0 = h->rs0;
+  size_t cs0 = h->cs0;
+  size_t rs1 = h->rs1;
+  size_t cs1 = h->cs1;
+
+  skl_test_check_matrix_params_rcprc(t, m, n, 1, 1, rs, cs, 1, 1);
```
myeh01:
```suggestion
  skl_test_check_matrix_params_rcprc(t, 1, 1, m, n, 1, 1, rs, cs);
```
Can we change this call so that the block size is `1 x 1` instead of `m x n`?

For a row- or column-major matrix, `m0 = m`, `n0 = n`, `m1 = 1`, `n1 = 1` is a valid way to describe the layout. But, I think packed matrices should always have a nonzero block size, so `skl_test_check_matrix_params_rcprc` requires that `m0 > 0` and `n0 > 0`. We will get an error if we test an edge case like `m == 0` or `n == 0`. Sorry it was not clear.

ref/pack/pack_e8rc_e8rcprc.h
```diff
@@ -0,0 +1,52 @@
...
+ * blocked matrix layout.
+ *
+ * @param m - Num. rows in input matrix
+ * @param n - Num. columns in input matrix
+ * @param src - Input matrix
+ * @param rs - Row stride of input matrix
+ * @param cs - Column stride of input matrix
+ * @param m0 - Num. rows in a block of the input matrix
+ * @param n0 - Num. columns in a block of the input matrix
+ * @param dst - Output packed matrix [m1 x n1 x (m0 x n0)]
```
myeh01:
```suggestion
 * @param m0 - Num. rows in a block of the output matrix
 * @param n0 - Num. columns in a block of the output matrix
 * @param dst - Packed output matrix [m1 x n1 x (m0 x n0)]
```
Sorry, I think I made some typos in the original documentation.

myeh01:
> Michael Yeh (@myeh01), thank you for your careful review. I've updated the code based on your feedback. Please let me know if you have any further comments.

Thanks @Harlen126 for the changes! I just had a few more minor comments, but otherwise it looks great!

test/pack/pack_e8rc_e8rcprc.c
Harlen126: It makes sense. Thanks~

test/skl-test-driver.h
Harlen126: Since it uses the LOG macro  (`SKL_TEST_REQUIRE`), I moved the macro up instead.

myeh01 (approved):
LGTM!

peichialin (approved):
LGTM~

PR #101, opened 2026-05-14
[GEMM] Optimize f64 large x390 kernel

Aaron-Hutchinson:
This performs a rewrite of the RVV f64 GEMM kernel for x390 with several improvements:

- Replace intrinsics with inline assembly to provide a fixed instruction schedule.
- Change tile size to `4 x m4 x 12` and structure the jj and kk loops similar to that of #57, #75, and #96.
- Use `ntl.p1` hints similar to the other kernels for better performance with clp memory.
- `skl_instruction_schedule_barrier` is now unused in the codebase and was removed.

### DIFF
--- a/include/skl-common.h
+++ b/include/skl-common.h
@@ -94,23 +94,6 @@
 #endif
 
 /**
- * @brief An instruction scheduling barrier.
- *
- * This function acts as a barrier to instruction scheduling, by pretending to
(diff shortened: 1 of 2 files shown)


ericlove (approved):
LGTM. Happy to see the deletion of `skl_instruction_schedule_barrier()`.

PR #104, opened 2026-05-18
DepthwiseConv2D: Simplify naming styles of public APIs.

peichialin:
* Append the layout to the data type, as expressed in gemm and pack APIs.
* Simplify the string of "specialization". Only include the parameters which have contraints.
* Remove the vectorization hint. Making it invisble to the users is more reasonable.

### DIFF
--- a/src/depthwise_conv2d/depthwise_conv2d_i8_i8_i32_zve32x.h
+++ b/src/depthwise_conv2d/depthwise_conv2d_i8_i8_i32_zve32x.h
@@ -72,7 +72,7 @@ extern "C" {
  * depthwise convolution kernels with int32 accumulators where input and output
  * are in HWC data layout, and vectorizes along the channel dimension.
  */
-void skl_depthwise_conv2d_vc_fnxn_sn_dn_mn_in_hwc_i8_i8_i32_zve32x(
+void skl_depthwise_conv2d_i8hwc_i8hwim_i32hwc_zve32x(
     int32_t *output, const int8_t *input, const int8_t *filter,
     size_t input_height, size_t input_width, size_t input_channel,
     size_t filter_height, size_t filter_width, size_t output_height,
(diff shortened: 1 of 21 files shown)


src/depthwise_conv2d/depthwise_conv2d_i8_i8_i32_zve32x.h
```diff
@@ -127,7 +127,7 @@ void skl_depthwise_conv2d_vc_fnxn_sn_dn_mn_in_hwc_i8_i8_i32_zve32x(
  * @note Filter dimensions must be exactly 3x3.
  * @note Depth multiplier must be exactly 1.
  */
-void skl_depthwise_conv2d_vc_f3x3_sn_dn_m1_in_hwc_i8_i8_i32_zve32x(
+void skl_depthwise_conv2d_f3x3m1_i8hwc_i8hwim_i32hwc_zve32x(
```
Harlen126, line 130: Maybe we could just remove the `m1` and rename it to `skl_depthwise_conv2d_f3x3_i8hwc_i8hwc_i32hwc_zve32x`.

peichialin: This API is for the specialization of depth multiplier == 1 so it does not provide the "depth multiplier" parameter.
That's why we have to preserve the `m1` word.

Harlen126: I mean `hwim` => `hwc` for kernel format~

peichialin: After revisit the filter layout expression, encoding its layout in `hwim` is still clearer than flattening it into 3d hwc.
I added the description of making pairs of input and filter channel.
Please take a look. Thanks!

Harlen126 (dismissed):
LGTM!

Harlen126 (approved):

PR #107, opened 2026-05-19
Install and package skl

hdelassus:
This PR introduces installation and packaging steps in SKL's CMake infrastructure.

This allows to install the SKL library wherever `CMAKE_INSTALL_PREFIX` points to, as well as to package it into a tarball or self-extracting tarball by default.

In turn, a user application may depend on SKL in a standard CMake fashion:
```cmake
# file: CMakeLists.txt
cmake_minimum_required(VERSION 3.23)

project(MyApp)

add_executable(myapp main.c)

find_package(Skl CONFIG REQUIRED)

message("Skl_DIR: ${Skl_DIR}")

target_link_libraries(myapp
  PRIVATE
  Skl::skl
)
```

Such a project may then be configured with:

```bash
cmake -G Ninja -B build \
  -DCMAKE_PREFIX_PATH=<list containing the path to SKL install or package directory> \
  --toolchain ...
```

The documentation is not yet built, installed nor packaged. This is something we should probably plan to do in the future.


CMakeLists.txt
```diff
@@ -241,3 +260,58 @@ if (SKL_BUILD_TESTS OR SKL_BUILD_BENCHMARKS)
...
+    "List of CPack binary generators to use.")
+set(CPACK_SOURCE_GENERATOR TGZ CACHE STRING
+    "List of CPack source generators to use.")
+set(CPACK_SOURCE_IGNORE_FILES
+  ".git*"
+  build/
+  ".*~$"    # Emacs backup files
+  ".*sw[a-p]$"  # Vim swap files
+  "*.orig$"  # Git mergetool backup files
+)
```
ebavier, line 247: Does cpack just hoover up everything else in the current working directory?  Is there some way to specify more explicitly what to include?

hdelassus: I believe not. See this paragraph at https://cmake.org/cmake/help/book/mastering-cmake/chapter/Packaging%20With%20CPack.html#cpack-source-packages

> Source packages in CPack simply copy the entire source tree for a project into a package file, and no install rules are used as they are in the case of binary packages. Out of source builds should be used to avoid having extra binary stuff polluting the source package. If you have files or directories in your source tree that are not wanted in the source package, you can use the variable [CPACK_SOURCE_IGNORE_FILES](https://cmake.org/cmake/help/latest/module/CPack.html#variable:CPACK_SOURCE_IGNORE_FILES) to exclude things from the package. This variable contains a list of regular expressions. Any file or directory that matches a regular expression in that list will be excluded from the sources.

ebavier: Tread lightly, I guess.

CMakeLists.txt
```diff
@@ -241,3 +260,58 @@ if (SKL_BUILD_TESTS OR SKL_BUILD_BENCHMARKS)
...
+install(
+  FILES
+    cmake/SklConfig.cmake
+    ${CMAKE_CURRENT_BINARY_DIR}/SklConfigVersion.cmake
+  DESTINATION ${CMAKE_INSTALL_LIBDIR}/cmake/Skl
+)
+
+# Configure libskl packaging
+
+set(CPACK_PACKAGE_VENDOR "SiFive, Inc")
```
Aaron-Hutchinson, line 235: I see that the resulting tarball is named `SKL-3.0.0-Generic.tar.gz`. It might be better to use "skl" instead of "SKL" to match the public header `skl.h` inside:
```suggestion
set(CPACK_PACKAGE_VENDOR "SiFive, Inc")
set(CPACK_PACKAGE_NAME skl)
```

CMakeLists.txt
```diff
@@ -241,3 +260,58 @@ if (SKL_BUILD_TESTS OR SKL_BUILD_BENCHMARKS)
       "-DM0=1;-DN0=1;-DK0=4;-DM1=128;-DN1=128;-DK1=8;-DCSA1=(2 * K0);-DRSA1=(K1 * CSA1);-DRSB0=1;-DCSB0=0;-DALPHA=1;-DBETA=0;")
   endif()
 endif()
+
+# Configure libskl installation
```
Aaron-Hutchinson, line 197: Would it be better to move the installation and packaging logic into the `cmake` directory? It's not a big deal to me, but it might help keep the top level CMakeLists tidy.

CMakeLists.txt
```diff
@@ -45,9 +48,25 @@ set(SKL_COMPILE_OPTIONS
...
+target_sources(${SKL_LIBRARY} PUBLIC FILE_SET HEADERS
+  BASE_DIRS
+    include
+  FILES
+    include/skl.h
+    include/skl-common.h
+)
+
+file(GLOB_RECURSE SKL_SRC_H LIST_DIRECTORIES FALSE CONFIGURE_DEPENDS "src/*.h")
+target_sources(${SKL_LIBRARY} PUBLIC FILE_SET source_HEADERS TYPE HEADERS
```
Aaron-Hutchinson: Why the strange casing on `source_HEADERS`? Is this a CMake keyword?

CMakeLists.txt
```diff
@@ -45,9 +48,25 @@ set(SKL_COMPILE_OPTIONS
...
 
-target_include_directories(${SKL_LIBRARY} PUBLIC include)
 target_compile_options(${SKL_LIBRARY} PRIVATE ${SKL_COMPILE_OPTIONS})
 
+target_sources(${SKL_LIBRARY} PUBLIC FILE_SET HEADERS
+  BASE_DIRS
+    include
+  FILES
+    include/skl.h
+    include/skl-common.h
```
Aaron-Hutchinson: With `tree -L 3` I see the generated install tree as:
```
build/install
├── include
│   ├── skl
│   │   ├── cvt
│   │   ├── depthwise_conv2d
│   │   ├── exp
│   │   ├── gelu
│   │   ├── gemm
│   │   ├── logistic
│   │   ├── silu
│   │   ├── softmax
│   │   └── transpose
│   ├── skl-common.h
│   └── skl.h
├── lib
│   ├── cmake
│   │   └── Skl
│   └── libskl.a
└── share
    └── licenses
        └── skl
```
Would it be better to place `skl-common.h` inside `install/include/skl` since it's not intended to be `#include`d by user files?

Aaron-Hutchinson (commented):
Thanks for taking on this work @hdelassus! The resulting install files and package look great overall. I'm not familiar with CPack so my review there is somewhat limited.

CMakeLists.txt
hdelassus: Yeah thanks I was thinking the same thing. We could also rename the project to `skl` instead of `SKL`. What's best, in your opinion?

CMakeLists.txt
hdelassus: When a FILE_SET is not a predefined one, it needs to start with lowercase. I'll check if there's a commonly used way of naming such variables.

CMakeLists.txt
hdelassus: Actually, do we even need to ship skl-common.h? I'm starting to think it doesn't really need to be shipped.

CMakeLists.txt
Aaron-Hutchinson: Can the project share a name with a library? If so, I'm in favor of naming them both `skl`.

CMakeLists.txt
Aaron-Hutchinson: Hmm. The file itself doesn't seem useful to `#include` outside of SKL, so if there's no headers depending on it then maybe not.

CMakeLists.txt
hdelassus: Yes. Note that this is only for the `package_source` target, not the binary one.
I suppose we could make this more robust if the source directory is still a git repository by listing all files ignored by git and adding them to the ignore-list... Sounds a bit like using a bulldozer to snatch a fly, though.

CMakeLists.txt
hdelassus: Most certainly they can.

pvchupin: lower case should be fine

CMakeLists.txt
hdelassus: I would be in favor of doing this if the toplevel listfile was much larger already, or if the installation and packaging steps were not as simple. The whole file is only 317 lines, and the install/packaging steps are taking only 50-ish lines. Does that sound a reasonable approach to you?

CMakeLists.txt
hdelassus: Looks like all-lowercase snake_case is common in the wild. I'll handle the rename. Thanks for raising your concern 😄

CMakeLists.txt
Aaron-Hutchinson: I'm fine with it. It should be easy to move later if there becomes a stronger need for it.

pvchupin:
@hdelassus , can we actually include a documentation for that?
A small sub-section on README.md would be good enough I think.

hdelassus:
> Hugues de Lassus (@hdelassus) , can we actually include a documentation for that? A small sub-section on README.md would be good enough I think.

Sure. Like just how to install and package, or also how to use in a project?

pvchupin:
I think both

include/skl-ref.h
```diff
@@ -11,76 +11,78 @@
...
+#include "ref/gemm/gemm_i8rc_i8rc_i32rc.h"
+
+#include "ref/gemm/gemm_bf16rcprc_bf16rcprc_f32rcprc.h"
+#include "ref/gemm/gemm_f16rcprc_f16rcprc_f16rcprc.h"
+#include "ref/gemm/gemm_f16rcprc_f16rcprc_f32rcprc.h"
+#include "ref/gemm/gemm_f32rcprc_f32rcprc_f32rcprc.h"
+#include "ref/gemm/gemm_f64rcprc_f64rcprc_f64rcprc.h"
+#include "ref/gemm/gemm_f8e4m3rcprc_f8e4m3rcprc_f32rcprc.h"
+#include "ref/gemm/gemm_f8e5m2rcprc_f8e5m2rcprc_f32rcprc.h"
+#include "ref/gemm/gemm_i8rcprc_i8rcprc_i32rcprc.h"
```
Aaron-Hutchinson, line 30: It looks like a few of these headers were missing, which you added in this PR. Maybe we should develop some kind of assertion (maybe using `find`+`grep`) that ensures all headers are accounted for? But not necessary in this PR.

include/skl.h
```diff
@@ -12,233 +12,233 @@
  */
 
 #if defined(__riscv_zve32x)
-#include "../src/gemm/rvv/gemm_i8_i8_i32_zve32x_x390.h"
+#include "gemm/rvv/gemm_i8_i8_i32_zve32x_x390.h"
```
Aaron-Hutchinson, line 15: I feel like this change deserves putting an explanation on what directory the paths are relative to somewhere. However, a code comment inside `skl.h` would lose context once this file is copied to the install directory. So I'm not sure how to handle this, and maybe just relying on context of the repository (i.e. no changes needed) is best.

include/skl-ref.h
```diff
@@ -11,76 +11,78 @@
...
-
-#include "../ref/gemm/gemm_bf16rcprc_bf16rcprc_f32rcprc.h"
-#include "../ref/gemm/gemm_f16rcprc_f16rcprc_f16rcprc.h"
-#include "../ref/gemm/gemm_f16rcprc_f16rcprc_f32rcprc.h"
-#include "../ref/gemm/gemm_f32rcprc_f32rcprc_f32rcprc.h"
-#include "../ref/gemm/gemm_f64rcprc_f64rcprc_f64rcprc.h"
-#include "../ref/gemm/gemm_f8e4m3rcprc_f8e4m3rcprc_f32rcprc.h"
-#include "../ref/gemm/gemm_f8e5m2rcprc_f8e5m2rcprc_f32rcprc.h"
-#include "../ref/gemm/gemm_i8rcprc_i8rcprc_i32rcprc.h"
+#include "ref/gemm/gemm_bf16rc_bf16rc_f32rc.h"
```
Aaron-Hutchinson, line 14: What's the reason that `skl.h` paths are relative to `src`, but `skl-ref.h` is relative to the repository root? Should these just be relative to `ref` for consistency?

README.md
```diff
@@ -68,13 +68,16 @@ Additionally, appropriate `-march` strings should be used to enable the necessar
 
 SKL's collection of kernels can be compiled into a single static library for use as `-lskl`, and used with the single header `skl.h` that provides the declarations of all SKL kernel functions.
 
-For projects that support CMake version 3.21 or higher, SKL provides a basic build system consisting of a single, top-level [CMakeLists.txt](./CMakeLists.txt), which can be used to produce a static library `libskl.a` as a library dependency.
+For projects that support CMake version 3.23 or higher, SKL provides a basic build system consisting of a single, top-level [CMakeLists.txt](./CMakeLists.txt), which can be used to produce a static library `libskl.a` as a library dependency.
```
Aaron-Hutchinson: > build system consisting of a single, top-level [CMakeLists.txt](./CMakeLists.txt)

This is inaccurate now right? Or is it referring to only what's needed to build `libskl.a`? Even so, there are supporting files in the `cmake` directory.

README.md
```diff
@@ -133,6 +143,42 @@ extern "C" {
...
+
+CMake automatically locates the SKL library and headers, and links the specified target against `libskl.a`.
+Versioning compatibility can be checked.
+SKL follows [semantic versioning](https://semver.org) for its releases.
+
+Projects requiring consistent RISC-V ISA extension support across all dependencies can utilize the `cmake/riscv.cmake` toolchain file when configuring the build:
+```shell
+cmake -B build -DCMAKE_TOOLCHAIN_FILE="path/to/SKL/cmake/riscv.cmake"
+```
+This approach ensures that both the consuming project and SKL are compiled with compatible ISA extensions.
```
Aaron-Hutchinson, line 185: This paragraph seems redundant given that the toolchain file was already specified in the previous directions on line 151. Or is this suggesting the consuming project should use SKL's toolchain file?

hdelassus: The latter.

README.md
hdelassus: It has become arguably inaccurate indeed... The top-level CML unconditionally calls `add_subdirectory(ref)`. Supporting files in `cmake` are, I would say, optional. One can build `libskl.a` with their own toolchain file, or setting variables at configure time in the CLI (CMAKE_C_COMPILER and SKL_COMPILE_OPTIONS).

Do you think something like this looks okay?
```suggestion
For projects that support CMake version 3.23 or higher, SKL provides a basic CMake-based build system that can be used to produce a static library `libskl.a` as a library dependency.
```

include/skl-ref.h
hdelassus: It's just because we have installation steps for libskl but not for libskl-ref, so it didn't matter for libskl-ref.

README.md
Aaron-Hutchinson: That sounds great to me!

include/skl.h
hdelassus: I believe no changes would mean we'd need to keep the headers in a `../src/` path relative to `skl.h`? Wouldn't this be awkward or even non-standard in an installed package? I agree the current proposition may not be immediately straightforward, and that a comment in skl.h might lose context when the file is installed. However, that might be good enough, as we can explain this difference between the source tree and the install tree. What do you think?

Another option might be to make the paths relative to the root, so that the install include directory may look like:

```
include
└── skl
    ├── skl.h
    └── src
        └── ...
```

This example shows `skl` as a subdirectory of `include` so as not to introduce a `src` directory at the root of `include`, which may be shared with other libraries (e.g. when installed at `/usr/include` or `usr/local/include`)

include/skl-ref.h
hdelassus: Good point. I think we might avoid any of this by including the respective library header wherever such functions are used, i.e. `skl.h` or `skl-ref.h`. I don't remember if any of our tests doesn't do this already, but IMHO that's something to keep in mind.

include/skl.h
Aaron-Hutchinson: Sorry, by "no changes needed" I meant no changes from the current state of this PR.

Having a `skl` subdirectory of `include` would mean users would likely have to `#include "skl/skl.h"` instead of `#include "skl.h"`, right? Not necessarily an issue, and I've seen other projects with this usage.

> a comment in skl.h might lose context when the file is installed. However, that might be good enough, as we can explain this difference between the source tree and the install tree

This seems like the simplest approach to me and maintains the clean install tree structure, so I'm most in favor of this solution.

hdelassus: > Having a skl subdirectory of include would mean users would likely have to #include "skl/skl.h" instead of #include "skl.h", right? Not necessarily an issue, and I've seen other projects with this usage.

Yes, probably. Or, with `include/skl` instead of just `include` in the install-interface, but that would be for CMake users.

Aaron-Hutchinson (approved):
I think this is in a great state for an initial setup of the installation and packaging infrastructure. Thanks for all of your effort and time on getting this working @hdelassus!

PR #108, opened 2026-05-19
Specify C standard and C99 minimum requirement

hdelassus:
This PR introduces a C99 minimum requirement to build SKL and its dependencies.
It also explicitly sets a C23 standard with extensions enabled, if available.

Since C23 triggered new clang-tidy warnings, the PR was split into two commits: one enabling C17, and one enabling C23 with clang-tidy fixes.


CMakeLists.txt
```diff
@@ -36,6 +36,16 @@ set(SKL_LIBRARY skl)
 
 add_library(${SKL_LIBRARY} STATIC)
 
+# Try to compile SKL with a recent standard with GNU extensions.
+set_target_properties(${SKL_LIBRARY}
+    PROPERTIES
+    C_STANDARD 23
+    C_EXTENSIONS ON
+    C_STANDARD_REQUIRED ON
```
Aaron-Hutchinson: Isn't `ON` the default for `C_EXTENSIONS`?

https://cmake.org/cmake/help/latest/prop_tgt/C_EXTENSIONS.html#prop_tgt:C_EXTENSIONS

Aaron-Hutchinson (dismissed):
LGTM

include/skl-common.h
```diff
@@ -5,7 +5,9 @@
 
 #pragma once
 
+#if __STDC_VERSION__ < 202311L
```
ericlove: Could we just disable whatever clang-tidy check is causing trouble about this in C23? Wouldn't that be preferable to adding this version guard in every single file that uses `bool`?

ericlove: Or better yet, why not just stick to C99 if that is indeed our minimum requirement?

CMakeLists.txt
hdelassus: Thanks for checking, it appears this is indeed unnecessary.

include/skl-common.h
hdelassus: > Could we just disable whatever clang-tidy check is causing trouble about this in C23? Wouldn't that be preferable to adding this version guard in every single file that uses `bool`?

I think this would be fine until `stdbool.h` is effectively removed from C, which doesn't appear to be the case for C2y (C29) yet.

> Or better yet, why not just stick to C99 if that is indeed our minimum requirement?

If we can be forward-compatible from C99 to C17 or C23, I think that's even better. What do you think?

Aaron-Hutchinson (approved):
LGTM

ericlove (approved):
Now that I understand this is "C99 minimum requirement" but "C23 forward compatibility" and the conditional includes are gone, I approve this.  In an ideal world, we would add CI to test multiple standards...

PR #110, opened 2026-05-21
[CI] Update release tag

myeh01:
This PR updates the release tag to v3.0.0.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,7 +16,7 @@ jobs:
       labels: ubuntu-22.04-x64-2
 
     env:
-      SKL_TAG: v0.0.0
+      SKL_TAG: v3.0.0
       TOOLCHAIN_TAG: 2025.08.28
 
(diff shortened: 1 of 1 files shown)


Aaron-Hutchinson (approved):
Admittedly I don't understand our CI very well, but it seems strange that bumping the SKL version would require a change in CI. I'm guessing this is due to how we currently host the QEMU and clang-tidy tarballs? In the longer term we should probably find a better solution for this.

hdelassus (approved):
LGTM.
@Aaron-Hutchinson I think this might be just a one-off to rename the release from 0.0.0 to 3.0.0. Future releases would, I believe, be _added_ instead of renamed from older ones.

Aaron-Hutchinson:
That makes sense under the current setup but it just seems like CI shouldn't care about the SKL version number at all. It appears to only be used to download QEMU and clang-tidy, and it would make more sense to track those versions separately.

PR #111, opened 2026-05-21
[GEMM] Optimize f16 large x390 kernel

Aaron-Hutchinson:
This performs some optimizations of the `skl_gemm_4xm4x1_f16_f16_f16_zvfh_x390` kernel. The resulting implementation is similar to that done in #101.

- Adjust tiling to `4 x m4 x 12`.
- Maximize load to use time for A and B matrix values by scheduling loads immediately after using a register.
- Use `ntl` hints.


ericlove (approved):
Looks good and results seem good.

PR #105, opened 2026-05-18
[GEMM] Update API specs for proposed change to packed GEMM & packing kernel names

ericlove:
Following discussion on https://github.com/sifive/skl/pull/64#issuecomment-4145894552, we agreed (@myeh01 @peichialin  @Harlen126 @france5289) to change the naming scheme for packing kernels and packed GEMMs. This PR updates the markdown documentation ([`packed-gemm.md`](https://github.com/sifive/skl/blob/main/src/gemm/packed-gemm.md) and [`pack/README.md`](https://github.com/sifive/skl/blob/main/src/pack/README.md)) to reflect the new conventions.

Specifically, the changes include:

1. Aligning both kernel families to indicate layout dimensions in matrix type specifiers
2. Adopting a unified type specifier format of `<type><sew>[rc]p<m0>x<n0>[rc]`
3. Conforming the `Xsfvqdotq` GEMM kernel API examples to `[k0, k1]` format, rather than arbitrary `k`

Note that this PR **does not implement the proposed changes** in the codebase; that will be done in forthcoming work by @myeh01 and @Harlen126 .  I had debated keeping this as a draft PR, but since there is no chance we'll update both code and READMEs atomically, I'm not sure it matters much.

PR #106, opened 2026-05-19
[GEMM] Update Xsfvqdotq GEMM kernel API

myeh01:
This PR updates the Xsfvqdotq GEMM kernel API to fully conform to the packed GEMM API.

PR #113, opened 2026-05-26
[GEMM] Rename Xsfmm GEMM kernels

myeh01:
This PR renames the relevant GEMM and packing kernels according to #105:
- The public packed Xsfmm kernels have been renamed: `skl_gemm_a1b01_?pc_?cp_?rcp_xsfmm32a? -> skl_gemm_a1b01_?ptex1c_?cp1xte_?rcptexte_xsfmm32a?`.
- The private Xsfmm kernels have not been renamed since they will be substantially revised in #14.
- The corresponding Xsfmm test harnesses that have been converted to the new format have been renamed.

### DIFF
--- a/src/gemm/xsfmm/gemm_a1b01_bf16c_bf16_f32_xsfmm32a16f.c
+++ b/src/gemm/xsfmm/gemm_a1b01_bf16c_bf16_f32_xsfmm32a16f.c
@@ -1306,7 +1306,7 @@ SKL_FUNC void skl_gemm_a1b01_bf16c_bf16_f32_xsfmm32a16f(
 }
 
 SKL_XSFMM_NEW
-SKL_FUNC void skl_gemm_a1b01_bf16pc_bf16cp_f32rcp_xsfmm32a16f(
+SKL_FUNC void skl_gemm_a1b01_bf16ptex1c_bf16cp1xte_f32rcptexte_xsfmm32a16f(
     size_t m1, size_t n1, size_t k, const __bf16 *a_pack, size_t rsa1,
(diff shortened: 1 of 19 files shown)


ericlove (commented):
I think the changes look good in general, but for the VQDOT kernels I don't think we should change the name without updating the API to reflect the constraints implied by name (K needs to be a multiple of 4, expressed as `k1`).  Perhaps this should be split into three PRs: one for the packing kernels, one for the Xsfmm kernels, and one for VQDOT?  It's longer than we'd generally like anyhow.

myeh01:
@ericlove Ok, I have reduced the scope of this PR to just include the Xsfmm kernels. I'll update the names of the Xsfvqdotq kernels after #106 is merged.

src/gemm/xsfmm/gemm_a1b01_bf16c_bf16_f32_xsfmm32a16f.c
```diff
@@ -1306,7 +1306,7 @@ SKL_FUNC void skl_gemm_a1b01_bf16c_bf16_f32_xsfmm32a16f(
 }
 
 SKL_XSFMM_NEW
-SKL_FUNC void skl_gemm_a1b01_bf16pc_bf16cp_f32rcp_xsfmm32a16f(
+SKL_FUNC void skl_gemm_a1b01_bf16ptex1c_bf16cp1xte_f32rcptexte_xsfmm32a16f(
```
myeh01, line 1309: `te` -> `ete`

ericlove (dismissed):
Looks good!

myeh01:
Need a reapproval after rebasing.

ericlove (approved):

PR #114, opened 2026-05-27
[Transpose] Merge Transpose-e8 kernel to Pack

Harlen126:
This PR demonstrates how to merge the transpose kernel into Pack:
1. Remove the transpose .c file, as its implementation is duplicated in the Pack source file.
2. Move the transpose function declaration to the Pack header file, making the transpose function public (`SKL_FUNC_PRIVATE` to `SKL_FUNC`).

The PR also updates the README to indicate where the transpose functions are located.

Note: The transpose tests remain unchanged because they rely on the transpose harness and skl.h. Therefore, the migration of the transpose kernel is transparent to these tests.


Harlen126:
The new PR (https://github.com/sifive/skl/pull/118) includes the e8/e16/e32 rvv transpose migration.

PR #118, opened 2026-06-02
[Pack][Transpose] Merge RVV Transpose kernel to Pack

Harlen126:
This PR merge the rvv transpose kernel into Pack:
1. Remove the transpose .c file, as its implementation is duplicated in the Pack source file.
2. Move the transpose function declaration to the Pack header file, making the transpose function public (`SKL_FUNC_PRIVATE` to `SKL_FUNC`).

The PR also updates the README to indicate where the transpose functions are located.

Note: The transpose tests remain unchanged because they rely on the transpose harness and skl.h. Therefore, the migration of the transpose kernel is transparent to these tests.


myeh01:
I think this looks good. After the Xsfmm transpose kernels are moved to the pack directory, is the plan to completely remove the transpose source directory? And I guess the transpose tests will be merged into the pack tests too?

Harlen126:
> After the Xsfmm transpose kernels are moved to the pack directory, is the plan to completely remove the transpose source directory? 

No, we won't completely remove it. The `transpose` folder will be kept, but it will only contain a single README file to direct users to the real source.

>And I guess the transpose tests will be merged into the pack tests too?

Since the (dummy) transpose kernel still exists, I think it's okay to keep the tests under the `test/transpose` folder.
@ericlove, could you share your thoughts on this?

src/transpose/README.md
```diff
@@ -1,6 +1,10 @@
 # Transpose Kernels
 
-This directory contains optimized kernels for matrix transpose operations. Matrix transpose is a fundamental linear algebra operation that converts an M×N matrix to an N×M matrix by swapping rows and columns. These kernels provide efficient implementations for various data types and target architectures.
+> **Note:** Transpose kernel implementations can be found in the [`../pack/`](../pack/) directory, since transpose can be viewed as a special case of packing where `m0=m`, `n0=n`, `cs=1`, and `rs0=1`.
```
peichialin: Should swapping the length and stride in row/column for output matrix makes more sense?
=> `m0=n`, `n0=m`, with `cs=1` and `rs0=1`.

src/transpose/README.md
```diff
@@ -22,27 +26,6 @@ Currently supports 8-bit, 16-bit, and 32-bit element transpose operations.
 
 ## Kernel List
 
-### Scalar Implementations
```
peichialin, line 25: Shall we keep the ref API list here if we won't remove them from `ref/transpose` directory?

src/transpose/README.md
Harlen126: For an `m`x`n` row-major matrix,
the transpose kernel is equivalent to calling:
```
skl_pack_e8rc_e8rcprc_zve32x(m, n, a /*src*/, rsa /*rs*/, 1 /*cs*/, m /*m0*/,
                             n /*n0*/, at /*dst*/, 1 /*rs0*/, rsat /*cs0*/,
                             0 /* rs1, don't care*/, 0 /*cs1, don't care*/,
                             0 /*pad, don't care,*/);
```
or
```
skl_pack_e8rc_e8rcprc_zve32x(m, n, a /*src*/, rsa /*rs*/, 1 /*cs*/, 1 /*m0*/,
                             1 /*n0*/, at /*dst*/, 0 /*rs0, don't care*/,
                             0 /*cs0, don't care*/, 1 /* rs1*/, rsat /*cs1*/,
                             0 /*pad, don't care,*/);
```
Maybe I can rewrite this part for better comprehensibility.

src/transpose/README.md
Harlen126: I'm not sure if we want to consider the reference function as a kernel.
I think it's just for the test harness.

src/transpose/README.md
peichialin: Thanks for the clarification.
I was wrong, we only have to swap the stride to the other direction while keeping the m0=m and n0=n as you proposed.

peichialin: Using a single block to explain transposition is easier to understand.
```
skl_pack_e8rc_e8rcprc_zve32x(m, n, a /*src*/, rsa /*rs*/, 1 /*cs*/, m /*m0*/,
                             n /*n0*/, at /*dst*/, 1 /*rs0*/, rsat /*cs0*/,
                             0 /* rs1, don't care*/, 0 /*cs1, don't care*/,
                             0 /*pad, don't care,*/);
```

src/transpose/README.md
peichialin: Good question. Another option is moving the introduction of ref APIs to `ref/transpose/README.md`.
I can modify `src/depthwise_conv2d/README.md` in the same way. What do you think?

Harlen126: Agreed. Let's move the reference function part to `ref/*/README.md`.

src/transpose/README.md
```diff
@@ -1,6 +1,10 @@
 # Transpose Kernels
 
-This directory contains optimized kernels for matrix transpose operations. Matrix transpose is a fundamental linear algebra operation that converts an M×N matrix to an N×M matrix by swapping rows and columns. These kernels provide efficient implementations for various data types and target architectures.
+> **Note:** Transpose kernel implementations can be found in the [`../pack/`](../pack/) directory, since transpose can be viewed as a special case of packing where `m0=m`, `n0=n`, `cs=1`, and `rs0=1`.
+> - **RVV implementations:** [`../pack/rvv/`](../pack/rvv/)
+> - **Xsfmm implementations:** [`../pack/xsfmm/`](../pack/xsfmm/)
+
+The `../pack/` directory contains optimized kernels for matrix transpose operations. Matrix transpose is a fundamental linear algebra operation that converts an M×N matrix to an N×M matrix by swapping rows and columns. These kernels provide efficient implementations for various data types and target architectures.
```
ericlove: Sorry, I know it's not part of your PR, but let's use this opportunity to maintain semantic linebreaks in markdown test:
```suggestion
The `../pack/` directory contains optimized kernels for matrix transpose operations.
Matrix transpose is a fundamental linear algebra operation that converts an M×N matrix to an N×M matrix by swapping rows and columns.
These kernels provide efficient implementations for various data types and target architectures.
```

ericlove:
> Since the (dummy) transpose kernel still exists, I think it's okay to keep the tests under the `test/transpose` folder.
> Eric Love (@ericlove), could you share your thoughts on this?

Yes, that's fine by me.

src/transpose/README.md
Harlen126: Updated in [826762d](https://github.com/sifive/skl/pull/118/commits/826762dc3ac3ac612287cc8484c4cee22c342472)
Thanks~

src/transpose/README.md
Harlen126: Fixed in [826762d](https://github.com/sifive/skl/pull/118/commits/826762dc3ac3ac612287cc8484c4cee22c342472)
Thanks~

src/transpose/README.md
Harlen126: Added README for `ref/transpose` in [119b0f1](https://github.com/sifive/skl/pull/118/commits/119b0f18bff0c80d262d2958ec2a3746737c2484)~

peichialin (dismissed):
LGTM~

peichialin (approved):
LGTM~

PR #120, opened 2026-06-05
[CI] Check license headers only for pull request event

myeh01:
This PR fixes the CI workflow so that license headers are checked pre-merge only.


Aaron-Hutchinson (approved):
Thanks for noticing the issue and providing a quick fix!

PR #121, opened 2026-06-09
[GEMM] Add widening bf16 RVV kernels for x390

Aaron-Hutchinson:
This adds widening bf16 RVV GEMM kernels specialized for x390 using the zvfbfwma extension. These kernels are a copy-paste of the f16 versions, with the following changes/notes:
- `vfwmacc.vf` was changed to `vfwmaccbf16.vf`.
- Since there is no widening multiply in the zvfbfwma extension, the `vfwmul.vf`s were replaced with a widening conversion followed by `vfmul.vf` in f32.
- There is a compiler bug on using `__bf16` values in an inline asm operand for which a fix has been landed but not yet included in our toolchain: https://github.com/llvm/llvm-project/pull/184566. The workaround taken in this PR is to use `float` types for the temporary A matrix values.

The performance for the new kernels is in line with the f16 versions.

### DIFF
--- a/src/gemm/rvv/gemm_bf16_bf16_f32_zvfbfwma_x390.c
+++ b/src/gemm/rvv/gemm_bf16_bf16_f32_zvfbfwma_x390.c
@@ -112,7 +112,7 @@ SKL_FUNC_PRIVATE void skl_gemm_1xm8x18_bf16_bf16_f32_zvfbfwma_x390(
           "vfmul.vf %[acc], %[acc], %[a0] \n\t"
           : [a0] "=&f"(a0),
             [b0] "=&vr"(b0),
-            [acc] "=vr"(acc)
+            [acc] "=&vr"(acc)
           : [jj_vl_in] "r"(jj_vl),
             [a_addr] "r"(a + ii * rsa),
             [b_addr] "r"(b + jj)
@@ -1172,7 +1172,7 @@ SKL_FUNC_PRIVATE void skl_gemm_6xm4x12_bf16_bf16_f32_zvfbfwma_x390(
 
             "vsetvli zero, %[jj_vl_in], e32, m4, ta, ma \n\t"
             "vfmul.vf %[acc], %[acc], %[a0] \n\t"
-            : [a0] "=&f"(a0), [b0] "=&vr"(b0), [acc] "=vr"(acc)
+            : [a0] "=&f"(a0), [b0] "=&vr"(b0), [acc] "=&vr"(acc)
             : [a_addr_83] "r"(a + (((ii0 + 0) * rsa) + ((kk_peel + 0) * 1))),
               [jj_vl_in] "r"(jj_vl),
               [b_addr_13] "r"(b + (((kk_peel + 0) * rsb) + ((jj + 0) * 1)))
--- a/src/gemm/rvv/gemm_bf16_bf16_f32_zvfbfwma_x390.c
+++ b/src/gemm/rvv/gemm_bf16_bf16_f32_zvfbfwma_x390.c
@@ -70,12 +70,12 @@ SKL_FUNC_PRIVATE void skl_gemm_1xm8x18_bf16_bf16_f32_zvfbfwma_x390(
   float a3;
   float a4;
   float a5;
-  vfloat16m4_t b0;
-  vfloat16m4_t b1;
-  vfloat16m4_t b2;
-  vfloat16m4_t b3;
-  vfloat16m4_t b4;
-  vfloat16m4_t b5;
+  vbfloat16m4_t b0;
+  vbfloat16m4_t b1;
+  vbfloat16m4_t b2;
+  vbfloat16m4_t b3;
+  vbfloat16m4_t b4;
+  vbfloat16m4_t b5;
   vfloat32m8_t acc;
   vfloat32m8_t c0;
 
@@ -453,10 +453,10 @@ SKL_FUNC_PRIVATE void skl_gemm_6xm4x12_bf16_bf16_f32_zvfbfwma_x390(
   float a33;
   float a43;
   float a53;
-  vfloat16m2_t b00;
-  vfloat16m2_t b10;
-  vfloat16m2_t b20;
-  vfloat16m2_t b30;
+  vbfloat16m2_t b00;
+  vbfloat16m2_t b10;
+  vbfloat16m2_t b20;
+  vbfloat16m2_t b30;
   vfloat32m4_t acc0;
   vfloat32m4_t acc1;
   vfloat32m4_t acc2;
@@ -464,7 +464,7 @@ SKL_FUNC_PRIVATE void skl_gemm_6xm4x12_bf16_bf16_f32_zvfbfwma_x390(
   vfloat32m4_t acc4;
   vfloat32m4_t acc5;
   float a0;
-  vfloat16m2_t b0;
+  vbfloat16m2_t b0;
   vfloat32m4_t acc;
   vfloat32m4_t c0;
   vfloat32m4_t c1;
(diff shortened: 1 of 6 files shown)


src/gemm/rvv/gemm_bf16_bf16_f32_zvfbfwma_x390.c
```diff
@@ -0,0 +1,1224 @@
...
+    for (ii = 0; (ii + 1) <= m; ii = ii + 1) {
+      for (jj = 0; jj < n; jj = jj + jj_vl) {
+        jj_vl = __riscv_vsetvl_e32m4(n - jj);
+        c0 = __riscv_vle32_v_f32m4(c + ii * rsc + jj, jj_vl);
+        c0 = __riscv_vfmul_vf_f32m4(c0, beta, jj_vl);
+        __riscv_vse32_v_f32m4(c + ii * rsc + jj, c0, jj_vl);
+      }
+    }
+    return;
+  }
```
hdelassus, line 487: Not a big deal, but why not use LMUL=8 in this case?

hdelassus (dismissed):
LGTM

test/gemm/rvv/skl_gemm_bf16_bf16_f32_zvfbfwma_x390.c
```diff
@@ -0,0 +1,146 @@
...
+// See LICENSE file in the project root for full license information.
+// SPDX-License-Identifier: MIT
+
+#include "gemm/gemm_bf16rcprc_bf16rcprc_f32rcprc.h"
+#include "skl-test-driver.h"
+#include "skl.h"
+#include <stddef.h>
+
+#if !defined(__riscv_zvfbfwma)
+#error This file requires the Zvfbfwma extension
```
myeh01, line 12: Is the version check required here as it is in the source file?

test/gemm/rvv/skl_gemm_bf16_bf16_f32_zvfbfwma_x390.c
```diff
@@ -0,0 +1,146 @@
...
+  // Set default strides: all matrices are row-major
+  for (size_t i = 0; i < suite.num_tests; ++i) {
+    tests[i].m0 = 1;
+    tests[i].n0 = 1;
+    tests[i].k0 = 1;
+
+    tests[i].rsa0 = 1;
+    tests[i].csa0 = 1;
+    tests[i].rsa1 = tests[i].rsa1 ? tests[i].rsa1 : tests[i].k1;
+    tests[i].csa1 = tests[i].csa1 ? tests[i].csa1 : 1;
```
myeh01:
```suggestion
    tests[i].csa1 =  1;
```
I might need to go back and fix some of the other harnesses, but should we set all the `cs*1` to 1 since it is required?

src/gemm/rvv/gemm_bf16_bf16_f32_zvfbfwma_x390.c
```diff
@@ -0,0 +1,1224 @@
...
+
+  // Due to a compiler bug, temporary A matrix values used in the inline
+  // assembly below are typed as float instead of __bf16.
+  float a0;
+  float a1;
+  float a2;
+  float a3;
+  float a4;
+  float a5;
+  vfloat16m4_t b0;
```
myeh01: Is the `vbfloat16m4_t` type available?

src/gemm/rvv/gemm_bf16_bf16_f32_zvfbfwma_x390.c
```diff
@@ -0,0 +1,1224 @@
...
+    for (jj = 0; jj < n; jj = jj + jj_vl) {
+      jj_vl = __riscv_vsetvl_e16m4(n - jj);
+
+      const size_t kk_unroll_degree = 18;
+      const size_t preload_distance = 6;
+
+      __asm__ volatile(
+          // clang-format off
+          "\n\t"
+          "flh %[a0], 0(%[a_addr]) \n\t"
```
myeh01, line 107: I'm not familiar with all the extension dependencies, but does `zvfbfwma` imply `zfh`?

src/gemm/rvv/gemm_bf16_bf16_f32_zvfbfwma_x390.c
```diff
@@ -0,0 +1,1224 @@
...
+
+          "vsetvli zero, %[jj_vl_in], e16, m4, ta, ma \n\t"
+          "vle16.v %[b0], (%[b_addr]) \n\t"
+          "vfwcvtbf16.f.f.v %[acc], %[b0] \n\t"
+
+          "vsetvli zero, %[jj_vl_in], e32, m8, ta, ma \n\t"
+          "vfmul.vf %[acc], %[acc], %[a0] \n\t"
+          : [a0] "=&f"(a0),
+            [b0] "=&vr"(b0),
+            [acc] "=vr"(acc)
```
myeh01: This doesn't need an early clobber because it's the last *vector* register used in the block?

src/gemm/rvv/gemm_bf16_bf16_f32_zvfbfwma_x390.c
myeh01: Never mind, it looks like `flh` is included in `zfbfmin`.

myeh01 (dismissed):
LGTM

src/gemm/rvv/gemm_bf16_bf16_f32_zvfbfwma_x390.c
Aaron-Hutchinson: I think it's just incorrect. In the f16 code this was copied from `acc` was only used in the last instruction, but here I've reused it in the widening conversion so I think it should now be an early clobber.

src/gemm/rvv/gemm_bf16_bf16_f32_zvfbfwma_x390.c
```diff
@@ -0,0 +1,1224 @@
...
+            "flh %[a0], 0(%[a_addr_83]) \n\t"
+            "fcvt.s.bf16 %[a0], %[a0] \n\t"
+
+            "vsetvli zero, %[jj_vl_in], e16, m2, ta, ma \n\t"
+            "vle16.v %[b0], (%[b_addr_13]) \n\t"
+            "vfwcvtbf16.f.f.v %[acc], %[b0] \n\t"
+
+            "vsetvli zero, %[jj_vl_in], e32, m4, ta, ma \n\t"
+            "vfmul.vf %[acc], %[acc], %[a0] \n\t"
+            : [a0] "=&f"(a0), [b0] "=&vr"(b0), [acc] "=vr"(acc)
```
myeh01: Then `acc` might also need an early clobber here.

src/gemm/rvv/gemm_bf16_bf16_f32_zvfbfwma_x390.c
Aaron-Hutchinson: The was just a copy-paste of the f16 version, so it should be updated there at some point as well. I've changed to m8 in this kernel.

test/gemm/rvv/skl_gemm_bf16_bf16_f32_zvfbfwma_x390.c
Aaron-Hutchinson: The version check was mostly just leftover from autogenerated code. I don't think they're necessary to have.

test/gemm/rvv/skl_gemm_bf16_bf16_f32_zvfbfwma_x390.c
Aaron-Hutchinson: Thanks, I've updated it.

src/gemm/rvv/gemm_bf16_bf16_f32_zvfbfwma_x390.c
Aaron-Hutchinson: Oops, I overlooked this. It is available and seems to work just fine in the inline asm.

src/gemm/rvv/gemm_bf16_bf16_f32_zvfbfwma_x390.c
Aaron-Hutchinson: Updated this, and the other occurrence in the other kernel.

myeh01 (approved):

PR #119, opened 2026-06-04
[Pack] Remove Xsfvqdotq pack B kernel

myeh01:
Since #98 provides a general e8 packing kernel, `skl_pack_b_i8_xsfvqdotq` is no longer needed. This PR removes the latter kernel and the associated test harness.

PR #122, opened 2026-06-10
[SKL][Pack] Optimize panel block cases

Harlen126:
This PR aims to close the performance gap between the general Pack kernel and the legacy `skl_pack_b_i8_xsfvqdotq` kernel, which was removed in https://github.com/sifive/skl/pull/119.
The potential gap mainly stems from the strided segmented load/store instructions in the current Pack kernel, as the hardware may achieve better performance by using unit-strided segmented load/store instructions instead.
Another minor improvement is the reduction of redundant scalar code for non-fringe blocks.
It also enables the optimization path for panel cases where `rs0` or `cs0` are ignored in row and column panels, respectively.


peichialin (approved):
Overall LGTM!
Shall we add some descriptions of the optimizations of column/row panels in the description?

PR #123, opened 2026-06-16
[GEMM] Optimize quad widening i8 RVV x390 kernel

Aaron-Hutchinson:
This optimizes the `skl_gemm_i8_i8_i32_zve32x_x390` kernel by effectively rewriting it with the following changes:
- Use inline assembly instead of intrinsics for a fixed instruction schedule.
- Switch from `lb, vle8.v, vwmul.v.x, vwadd.v.v` sequence to `lb, vle8.v, vwcvt.x.x.v, vwmacc.vx` so that the single widening is amortized across the arithmetic operations.
- Change tile size from `4 x m4 x 2` to `4 x m4 x 4` and maximize load-to-use time of A and B values.
- Use `ntl.p1` hints for better performance with CLP memory.

Peak performance improved by about 15% with memory port and about 100% with CLP.

### DIFF
--- a/src/gemm/rvv/gemm_i8_i8_i32_zve32x_x390.c
+++ b/src/gemm/rvv/gemm_i8_i8_i32_zve32x_x390.c
@@ -43,198 +43,531 @@
  *     c, rsc, 1
  * );
  * ```
- * Uses a 4 x LMUL=4 x 2 register tile. Vectorized across the N dimension.
+ * Uses a 4 x LMUL=4 x 4 register tile. Vectorized across the N dimension.
  *
  * @note
  * Works best when `m >= 4` and `n >= __riscv_vsetvlmax_e32m4()`.
  */
-SKL_FUNC_PRIVATE void skl_gemm_4xm4x2_i8_i8_i32_zve32x_x390(
+SKL_FUNC_PRIVATE void skl_gemm_4xm4x4_i8_i8_i32_zve32x_x390(
     size_t m, size_t n, size_t k, int32_t alpha, const int8_t *a, size_t rsa,
     const int8_t *b, size_t rsb, int32_t beta, int32_t *c, size_t rsc) {
   size_t jj_vl;
   size_t ii;
   size_t jj;
-  size_t kk_init;
   size_t kk;
-  size_t kk0;
-  size_t ii0;
-  int8_t a00_0;
-  int8_t a01_0;
+  int8_t a00;
+  int8_t a10;
+  int8_t a20;
+  int8_t a30;
+  int8_t a01;
+  int8_t a11;
+  int8_t a21;
+  int8_t a31;
   int8_t a02;
+  int8_t a12;
+  int8_t a22;
+  int8_t a32;
   int8_t a03;
-  vint8m1_t b0;
-  vint16m2_t prod0_0;
-  vint16m2_t prod1_0;
-  vint16m2_t prod2;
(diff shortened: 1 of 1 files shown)


hdelassus:
> * Switch from `lb, vle8.v, vwmul.v.x, vwadd.v.v` sequence to `lb, vle8.v, vwcvt.x.x.v, vwmacc.vx` so that the single widening is amortized across the arithmetic operations.

I do not understand what this means: there are still two widenings? Neither do I understand how it can be correct as `vwmul.vx` does an `SEW*SEW = 2*SEW` product?

src/gemm/rvv/gemm_i8_i8_i32_zve32x_x390.c
```diff
@@ -43,198 +49,543 @@
...
+          "vsetvli zero, %[jj_vl], e8, m1, ta, ma \n\t"
+          "vle8.v %[b00], (%[b_addr0]) \n\t"
+          "add %[b_addr0], %[b_addr0], %[b_inc] \n\t"
+          "vwcvt.x.x.v %[b00w], %[b00] \n\t"
+
+          "vsetvli zero, %[jj_vl], e16, m2, ta, ma \n\t"
+          "vwmul.vx %[acc00], %[b00w], %[a00] \n\t"
+          "vwmul.vx %[acc10], %[b00w], %[a10] \n\t"
+          "vwmul.vx %[acc20], %[b00w], %[a20] \n\t"
+          "vwmul.vx %[acc30], %[b00w], %[a30] \n\t"
```
hdelassus, line 140: In this sequence:
* b00w has SEW=16
* a00 has SEW=8, but is treated as SEW=16, so that works because `lb` sign-extends to `XLEN`, right?

hdelassus: And the same goes for `vwmacc.vx` later. So I get why this works, now, but how is the cost of widening "amortized across arithmetic ops"?

hdelassus (approved):
Otherwise LGTM!

src/gemm/rvv/gemm_i8_i8_i32_zve32x_x390.c
Aaron-Hutchinson: Yes, that's exactly right. Sorry for not emphasizing this. The widening of the scalars is obtained "for free" through the load instructions.

> how is the cost of widening "amortized across arithmetic ops"?

I just meant that previously there were `2*unroll(m)` (ignoring k unrolling) arithmetic ops, where each accumulator register needed 2 widening operations independent of the other accumulators. Now the first set of widening ops (`vfwmul.v.x`) has been collapsed into a single op (`vwcvt.x.x.v`) that can easily be performed in advance and whose cost is fixed relative to `unroll(m)`.

ericlove:
> I do not understand what this means: there are still two widenings? 

Yes, but for a given register tiling `mr x lmul`, there is one `lmul`-wide conversion and `mr` scalar widenings.  If the latter are assumed to be overlapped with vector compute, then the vector widening conversion is amortized overall the `vwmul`s.

> Neither do I understand how it can be correct as `vwmul.vx` does an `SEW*SEW = 2*SEW` product?

In that step, SEW=16 due to the previously-completed widenings, and the result is 2*SEW=32.

ericlove (approved):
LGTM

PR #124, opened 2026-06-16
[Test] Port SiLU/Logistic test harnesses

hdelassus:
This PR ports the legacy logistic and SiLU tests to the new test harness framework.
In doing so, I uncovered an accuracy issue in the Xsfvfexp32e FP32 logistic kernel that wasn't caught before because the legacy tests didn't generate negative inputs. Indeed, they used `SKL_TEST_MIN_F32` that is defined to be `0.0f`.

Thanks to @ebavier for quickly looking into it and providing a simple fix.


ref/silu/silu_f32.c
```diff
@@ -7,16 +7,10 @@
...
-      float ex = expf(x);
-      out[i] = x * ex / (1.0f + ex);
-    } else {
-      /* For x > 0, compute silu(x) = x * (1 - logistic(-x)) */
-      float ex = expf(-x);
-      out[i] = x * (1.0f - ex / (1.0f + ex));
-    }
+    double x = in[i];
+    double e = exp(-x);
+    out[i] = (float)(x / (1. + e));
```
ebavier: Since `e` is not reused, I'd be fine with simply putting the `exp(-x)` inline in line 14.  Logistic too.

ebavier (commented):
LGTM!  Thanks for taking care of this.  Very tiny suggestion:

ebavier (approved):

PR #125, opened 2026-06-16
[GEMM] Add fused kernel API documentation

myeh01:
This PR adds documentation for the fused kernel API used in #14.

### DIFF
--- a/src/gemm/xsfmm/fused-kernel-API.md
+++ b/src/gemm/xsfmm/fused-kernel-API.md
@@ -2,10 +2,9 @@
 In applications, users might want to perform an operation such as alpha/beta scaling, adding a bias, applying an activation function, etc. after computing a matrix product.
 The simplest approach is to have distinct GEMM and post-GEMM kernels.
 The former compute a GEMM and store the result to memory, while the latter operate on matrices in memory.
-However, this approach unnecessarily stores and reloads the matrix.
-An alternate approach for GEMM kernels using SiFive's Xsfmm extension is to leave the matrix product in the tile state and pass it directly to the post-GEMM operator.
-This document describes an API for post-GEMM kernels that allow them to be "fused" to Xsfmm GEMM kernels.
-It is assumed that readers are familiar with SKL's packed GEMM API and SiFive's Xsfmm extension.
+An alternate approach for GEMM kernels using SiFive's Xsfmm extension is to leave the matrix product in the tile state and pass it directly to the post-GEMM operator, avoiding the storing and reloading of the matrix.
+This document describes an API for post-GEMM kernels that allows them to be "fused" to the Xsfmm GEMM kernels.
+It is assumed that readers are familiar with [SKL's packed GEMM API](../packed-gemm.md) and SiFive's Xsfmm extension.
 
 ## Fused Kernel API
 Fused kernels perform an elementwise operation on a single tile in the Xsfmm tile state and a single block of `C`.
@@ -23,6 +22,7 @@ where
 - `c + row1 * rsc1 + col1 * csc1` points to the block of `C` to be operated on
 - `rsc0` and `csc0` are the block's row and column strides
 - `params` points to a struct containing any other parameters the kernel needs.
+
 Let `tss[i, j]` denotes the `j`th entry of the row or column specified by `tss + i`, and let `c_block = c + row1 * rsc1 + col1 * csc1`.
 Then `tss[i, j]` should correspond to `c_block[i * rsc0 + j * csc0]` under the kernel's elementwise operation.
 Note that if `tss` has a column pattern, then the operation is effectively applied to the *transpose* of the subtile.
@@ -39,6 +39,7 @@ typedef struct {
   float beta;
 } alpha_beta_scaling_params_f32_f32_t;
 
+SKL_XSFMM_IN
 void skl_gemm_alpha_beta_scaling_f32_f32rcptexterc_xsfmmbase(
     size_t tm, size_t tn, size_t tss, float *c, size_t rsc0, size_t csc0,
     size_t rsc1, size_t csc1, size_t row1, size_t col1, void *params) {
@@ -61,6 +62,7 @@ typedef struct {
   size_t csbias1;
 } add_bias_params_f32_f32_t;
 
+SKL_XSFMM_IN
 void skl_gemm_add_bias_f32_f32rcptexterc_xsfmmbase(
     size_t tm, size_t tn, size_t tss, float *c, size_t rsc0, size_t csc0,
     size_t rsc1, size_t csc1, size_t row1, size_t col1, void *params) {
@@ -82,6 +84,7 @@ typedef struct {
   float *max;
 } matrix_max_params_f32_f32_t;
 
+SKL_XSFMM_IN
 void skl_gemm_matrix_max_f32_f32rcptexterc_xsfmmbase(
     size_t tm, size_t tn, size_t tss, float *c, size_t rsc0, size_t csc0,
     size_t rsc1, size_t csc1, size_t row1, size_t col1, void *params) {
@@ -107,6 +110,7 @@ typedef void (*fused_<type>_<type>_t)(size_t tm, size_t tn, size_t tss,
                                       size_t rsc1, size_t csc1, size_t row1,
                                       size_t col1, void *params) SKL_XSFMM_IN;
 
+SKL_XSFMM_IN
 void skl_gemm_apply_fused_<type>_<type>rcptexterc_<isa>(
     size_t m, size_t n, size_t tss, size_t rstss, size_t cstss, <type> *c,
     size_t rsc0, size_t csc0, size_t rsc1, size_t csc1, size_t row1,
--- a/src/gemm/xsfmm/fused-kernel-API.md
+++ b/src/gemm/xsfmm/fused-kernel-API.md
@@ -21,14 +21,14 @@ The fused kernel must then operate on the transpose of each tile of the result; 
 The inner loop functions have the following API:
 ```
 SKL_FUNC_PRIVATE void
-skl_gemm_inner_loop_m1xn1_<type>rcptex1c_<type>rcp1xte_<type>_<isa>(
-    size_t m, size_t n, size_t k, const <type> *a, size_t rsa1, size_t csa1,
-    const <type> *b, size_t rsb1, size_t csb1) __riscv_inout("xsfmm");
+skl_gemm_inner_loop_m1xn1_<a type>rcptex1c_<b type>rcp1xte_<accum type>_<isa>(
+    size_t m, size_t n, size_t k, const <a type> *a, size_t rsa1, size_t csa1,
+    const <b type> *b, size_t rsb1, size_t csb1) __riscv_inout("xsfmm");
 ```
 Since the inner loop functions use Xsfmm instructions to compute `A * B`, they assume `A` is packed into `ETE` x 1 column-major blocks and `B` into 1 x `ETE` row-major blocks.
 `rsa1` and `csa1` denote the inter-block row and column strides for `A`, and `rsb1` and `csb1` are the corresponding strides for `B`.
 `m`, `n`, and `k` denote the problem dimensions, and it is the caller's responsibility to ensure that `0 <= m <= m1 * ETE` and `0 <= n <= n1 * ETE`.
-If `m` or `n` is not a multiple of `ETE`, the function will only compute partial tiles along the bottom or right edges, as illustrated below for a 2 x 2 tiling. 
+If `m` or `n` is not a multiple of `ETE`, the function will only compute partial tiles along the bottom or right edges, as illustrated below for a 2 x 2 tiling.
 ```
  ┌─────────────n──────────────┐
 ┌┌────────────────┬───────────┬────┐┐
@@ -89,7 +89,7 @@ This single-tile API avoids the need for kernel authors to know how each inner l
 
 The API for fused kernels is
 ```
-void kernel(size_t tm, size_t tn, size_t tss, <type> *c, size_t rsc0,
+void kernel(size_t tm, size_t tn, size_t tss, <c type> *c, size_t rsc0,
             size_t csc0, size_t rsc1, size_t csc1, size_t row1, size_t col1,
             void *params) __riscv_in("xsfmm");
 ```
@@ -117,14 +117,16 @@ To compute a GEMM `C = alpha * A * B + beta * C`, the user can apply an alpha/be
 typedef struct {
   float alpha;
   float beta;
-} skl_gemm_alpha_beta_scaling_params_f32_f32_t;
+} skl_gemm_alpha_beta_scaling_params_f32_f32_f32rcptexterc_xsfmmbase_t;
 
-void skl_gemm_alpha_beta_scaling_f32_f32rcptexterc_xsfmmbase(
+void skl_gemm_alpha_beta_scaling_f32_f32_f32rcptexterc_xsfmmbase(
     size_t tm, size_t tn, size_t tss, float *c, size_t rsc0, size_t csc0,
     size_t rsc1, size_t csc1, size_t row1, size_t col1, void *params)
     __riscv_in("xsfmm") {
-  skl_gemm_alpha_beta_scaling_params_f32_f32_t *params_cast =
-      (skl_gemm_alpha_beta_scaling_params_f32_f32_t *)params;
+  skl_gemm_alpha_beta_scaling_params_f32_f32_f32rcptexterc_xsfmmbase_t
+      *params_cast =
+          (skl_gemm_alpha_beta_scaling_params_f32_f32_f32rcptexterc_xsfmmbase_t
+               *)params;
   float alpha = params_cast->alpha;
   float beta = params_cast->beta;
 
@@ -137,7 +139,7 @@ void skl_gemm_alpha_beta_scaling_f32_f32rcptexterc_xsfmmbase(
 ```
 This kernel is already provided by SKL.
 In actual code, `sf.vtmv.v.t` instructions are used in the inner loop to move rows or columns of the tile state to the vector register file.
-An unoptimized implementation of this kernel in inline assembly might look like: 
+An unoptimized implementation of this kernel in inline assembly might look like:
 ```
 if (tm == 0 || tn == 0) {
   return;
@@ -170,14 +172,14 @@ The following kernel can be used to compute `C = A * B + bias`, where `bias` is 
 ```
 typedef struct {
   float *bias;
-} skl_gemm_add_bias_params_f32_f32_t;
+} skl_gemm_add_bias_params_f32_f32_f32rcptexterc_xsfmmbase_t;
 
-void skl_gemm_add_bias_f32_f32rcptexterc_xsfmmbase(
+void skl_gemm_add_bias_f32_f32_f32rcptexterc_xsfmmbase(
     size_t tm, size_t tn, size_t tss, float *c, size_t rsc0, size_t csc0,
     size_t rsc1, size_t csc1, size_t row1, size_t col1, void *params)
     __riscv_in("xsfmm") {
-  skl_gemm_add_bias_params_f32_f32_t *params_cast =
-      (skl_gemm_add_bias_params_f32_f32_t *)params;
+  skl_gemm_add_bias_params_f32_f32_f32rcptexterc_xsfmmbase_t *params_cast =
+      (skl_gemm_add_bias_params_f32_f32_f32rcptextexrc_xsfmmbase_t *)params;
   float *bias = params_cast->bias;
 
   float *c_block = c + row1 * rsc1 + col1 * csc1;
@@ -194,14 +196,14 @@ To compute the maximum value of a matrix product, the following kernel can be us
 ```
 typedef struct {
   float *max;
-} skl_gemm_matrix_max_params_f32_f32_t;
+} skl_gemm_matrix_max_params_f32_f32_f32rcptexterc_xsfmmbase_t;
 
-void skl_gemm_matrix_max_f32_f32rcptexterc_xsfmmbase( 
+void skl_gemm_matrix_max_f32_f32_f32rcptexterc_xsfmmbase(
     size_t tm, size_t tn, size_t tss, float *c, size_t rsc0, size_t csc0,
     size_t rsc1, size_t csc1, size_t row1, size_t col1, void *params)
     __riscv_in("xsfmm") {
-  skl_gemm_matrix_max_params_f32_f32_t *params_cast =
-      (skl_gemm_matrix_max_params_f32_f32_t *)params;
+  skl_gemm_matrix_max_params_f32_f32_f32rcptexterc_xsfmmbase_t *params_cast =
+      (skl_gemm_matrix_max_params_f32_f32_f32rcptexterc_xsfmmbase_t *)params;
   float *max = params_cast->max;
 
   float *c_block = c + row1 * rsc1 + col1 * csc1;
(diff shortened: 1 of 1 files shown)


myeh01:
Opening this as a draft PR so people can start reviewing it. Still needs some clean up, but any high level feedback is welcome.

myeh01:
Ready for review.

src/gemm/xsfmm/fused-kernel-API.md
```diff
@@ -0,0 +1,378 @@
+# Xsfmm Fused Kernel API
+In applications, users might want to perform an operation such as alpha/beta scaling, adding a bias, applying an activation function, etc. after computing a matrix product.
+The simplest approach is to have distinct GEMM and post-GEMM kernels.
+The former compute a GEMM and store the result to memory, while the latter operate on matrices in memory.
+An alternate approach for GEMM kernels using SiFive's Xsfmm extension is to leave the matrix product in the tile state and pass it directly to the post-GEMM operator, which avoids storing and reloading the matrix.
+This document describes an API for post-GEMM kernels that allows them to be "fused" in this way to the Xsfmm GEMM kernels.
+It is assumed that readers are familiar with [SKL's packed GEMM API](../packed-gemm.md) and SiFive's Xsfmm extension.
```
ericlove, line 8: We need to mention that this is made possible only by the existence of the C ABI attributes for Xsfmm functions, and cite the spec in https://www.sifive.com/document-file/xsfmm-matrix-extensions-specification

src/gemm/xsfmm/fused-kernel-API.md
```diff
@@ -0,0 +1,378 @@
...
+
+## The Fused Kernel API
+Fused kernels perform an elementwise operation on a single tile in the Xsfmm tile state and a single block of `C`.
```
ericlove: I'm not sure "elementwise" is technically correct here, since the bias addition example is position-dependent, and max-reduce is a reduction.

src/gemm/xsfmm/fused-kernel-API.md
```diff
@@ -0,0 +1,378 @@
+# Xsfmm Fused Kernel API
+In applications, users might want to perform an operation such as alpha/beta scaling, adding a bias, applying an activation function, etc. after computing a matrix product.
+The simplest approach is to have distinct GEMM and post-GEMM kernels.
+The former compute a GEMM and store the result to memory, while the latter operate on matrices in memory.
+An alternate approach for GEMM kernels using SiFive's Xsfmm extension is to leave the matrix product in the tile state and pass it directly to the post-GEMM operator, which avoids storing and reloading the matrix.
+This document describes an API for post-GEMM kernels that allows them to be "fused" in this way to the Xsfmm GEMM kernels.
+It is assumed that readers are familiar with [SKL's packed GEMM API](../packed-gemm.md) and SiFive's Xsfmm extension.
+
```
ericlove, line 9: We should logically motivate the single-tile fused kernel API by describing the underlying GEMM implementation as a combination of individual register tilings.  We could have just exposed those, and let users operate on their results.  Instead, the value-add of this API is that it allows fused function authors to handle only a single tile, and avoid the need to understand how data is arranged in matrix registers, although we do establish a convention for this internally and describe it later in the document.

src/gemm/xsfmm/fused-kernel-API.md
```diff
@@ -0,0 +1,378 @@
...
+## The Fused Kernel API
+Fused kernels perform an elementwise operation on a single tile in the Xsfmm tile state and a single block of `C`.
+Their API is
+```
+SKL_XSFMM_IN
```
ericlove: For this document, we should either replace `SKL_XSFMM_*` by `__riscv_*("xsfmm")` or add a note explaining the use of these macros.

src/gemm/xsfmm/fused-kernel-API.md
```diff
@@ -0,0 +1,378 @@
...
+- `tss` is the tile subset specifier for the first row (column) of the subtile to be operated on
+- `tn` is the length of the row (resp. column) `tss` specifies
+- `tm` is the number of rows (resp. columns) in the subtile, i.e. the kernel operates on rows (resp. columns) `tss` to `tss + tm - 1`
+- `c + row1 * rsc1 + col1 * csc1` points to the block of `C` to be operated on
+- `rsc0` and `csc0` are the block's row and column strides
+- `params` points to a struct containing any other parameters the kernel needs.
+
+Let `tss[i, j]` denote the `j`th entry of the row or column specified by `tss + i`, and let `c_block = c + row1 * rsc1 + col1 * csc1`.
+Then `tss[i, j]` should correspond to `c_block[i * rsc0 + j * csc0]` under the kernel's elementwise operation.
+Note that if `tss` has a column pattern, then the operation is effectively applied to the *transpose* of the subtile.
```
ericlove, line 106: In addition to the arguments, we should document the implicit contract between the `kernel` and the calling GEMM routine.  Specifically:

1. The `kernel` will be called exactly once for each `TE` x `TE` tile of the output matrix produced by the calling GEMM.
2. There is no guaranteed order in which the output matrix tiles will be covered (presumably?)
3. Because SKL functions are single-thread, the fused `kernel` will only be called from the same thread as the GEMM, so it is safe to mutate state in `params` without synchronization.

I think this it? Please push back if you think it's overkill or incomplete.

src/gemm/xsfmm/fused-kernel-API.md
```diff
@@ -0,0 +1,378 @@
...
+
+### Examples
+We give some examples below of fused kernels for the `float` type.
+Pseudocode is given to illustrate the operation each kernel performs.
+
+#### Alpha/beta scaling
+To compute a GEMM `C = alpha * A * B + beta * C`, the user can apply an alpha/beta scaling kernel:
```
ericlove, line 115: We should mention that this example is different from the others in that it is actually supplied by SKL, so the user doesn't need to write this.

src/gemm/xsfmm/fused-kernel-API.md
```diff
@@ -0,0 +1,378 @@
...
+- `rsc0` and `csc0` are the block's row and column strides
+- `params` points to a struct containing any other parameters the kernel needs.
+
+Let `tss[i, j]` denote the `j`th entry of the row or column specified by `tss + i`, and let `c_block = c + row1 * rsc1 + col1 * csc1`.
+Then `tss[i, j]` should correspond to `c_block[i * rsc0 + j * csc0]` under the kernel's elementwise operation.
+Note that if `tss` has a column pattern, then the operation is effectively applied to the *transpose* of the subtile.
+
+### Examples
+We give some examples below of fused kernels for the `float` type.
+Pseudocode is given to illustrate the operation each kernel performs.
```
ericlove, line 112: Explicitly state that primary reason it's "pseudocode" due to its use of the notation `tss[i, j]` to access elements of tile registers, which in real code would need to be moved into vector registers with `sf.vtmv.v.t` instructions.  Probably best to link to a real but short example of this, or even write out the inline asm you'd write for a simple tile writeback.

src/gemm/xsfmm/fused-kernel-API.md
```diff
@@ -0,0 +1,378 @@
...
+        *max = tss[i, j];
+      c_block[i * rsc0 + j * csc0] = tss[i, j];
+    }
+}
+```
+
+### Fused Kernel Application Functions
+SKL provides (private) functions that apply a fused kernel to multiple tiles:
+```
+typedef void (*skl_gemm_fused_kernel_<type>_<type>_t)(
```
ericlove: Haven't we already stated this typedef?

src/gemm/xsfmm/fused-kernel-API.md
```diff
@@ -0,0 +1,378 @@
...
+  for (size_t i = 0; i < tm; ++i)
+    for (size_t j = 0; j < tn; ++j) {
+      if (tss[i, j] > *max)
+        *max = tss[i, j];
+      c_block[i * rsc0 + j * csc0] = tss[i, j];
+    }
+}
+```
+
+### Fused Kernel Application Functions
```
ericlove, line 219: I think this explanation would make more sense if we reorganized it and the following section to first present the convention used by all SKL GEMM register tilings for arranging matrix registers, and then explained how this allows us to write an "application function" that applies the fused kernel to some multi-tile sub-matrix.

You can explain that the inner loop functions support a transposed mode to avoid duplicate code for every tiling and its transpose, and then that will better motivate the case where "`tss` has a column pattern" as mentioned up here.

The current order is IMO harder to follow on the first read, but feel free to ignore this if you disagree or it would be too much work to change.

src/gemm/xsfmm/fused-kernel-API.md
```diff
@@ -0,0 +1,378 @@
...
+SKL provides (private) functions that apply a fused kernel to multiple tiles:
+```
+typedef void (*skl_gemm_fused_kernel_<type>_<type>_t)(
+    size_t tm, size_t tn, size_t tss, <type> *c, size_t rsc0, size_t csc0,
+    size_t rsc1, size_t csc1, size_t row1, size_t col1,
+    void *params) SKL_XSFMM_IN;
+
+SKL_XSFMM_IN
+SKL_FUNC_PRIVATE
+void skl_gemm_apply_fused_<type>_<type>rcptexterc_<isa>(
```
ericlove: What do the two separate `<type>` fields refer to? One presumably to `c`, and the other to the tile state in the accumulators? Presumably we should clarify this above (`f32_f32` suffix).  But IIUC, maybe we even would be better off using `_xsfmm32a32f` (etc.) for the accumulator part?

src/gemm/xsfmm/fused-kernel-API.md
```diff
@@ -0,0 +1,378 @@
...
+n│                │           │    │
+│├────────────────┼───────────┼────┤
+││                │           │    │
+││                │           │    │
+││      mt4       │      mt12 │    │
+└├────────────────┼───────────┘    │
+ │                │                │
+ └────────────────┴────────────────┘
+```
+So, if `tss` has a column pattern, then the application function transposes this region of the tile state, applies the kernel to each subtile, and then stores them to the leading `m` x `n` portion of `C`.
```
ericlove:
```suggestion
So, if `tss` has a column pattern, then the application function transposes this region of the tile state (implicitly, by moving columns rather than rows to the vector register file), applies the kernel to each subtile, and then stores them to the leading `m` x `n` portion of `C`.
```
Perhaps helpful to clarify this parenthetically and assuage any concerns about performance.

src/gemm/xsfmm/fused-kernel-API.md
```diff
@@ -0,0 +1,378 @@
...
+
+## Inner Loop Functions
+The Xsfmm GEMM inner loop functions accumulate a matrix product `A * B` into the current tile state and pass it directly to a fused kernel.
```
ericlove: Technically, it's the application function that passes the tile state to a fused kernel right? I would say 
```suggestion
The Xsfmm GEMM inner loop functions accumulate a partial matrix product `A * B` into the tile state for a specific matrix register tiling, and return this result in matrix registers.
```

src/gemm/xsfmm/fused-kernel-API.md
```diff
@@ -0,0 +1,378 @@
...
+The number of available tiles in the tile state determines the possible tilings of the `C` matrix.
```
ericlove:
```suggestion
The number of available tiles in the matrix engine state determines the possible register tilings, and thus the number of inner loop functions for a given datatype.
```

src/gemm/xsfmm/fused-kernel-API.md
```diff
@@ -0,0 +1,378 @@
...
+Since the inner loop functions use Xsfmm instructions to compute `A * B`, they assume `A` is packed into `ETE` x 1 column-major blocks and `B` into 1 x `ETE` row-major blocks.
+The inner loop functions have the following API:
+```
+SKL_XSFMM_INOUT
+SKL_FUNC_PRIVATE void
+skl_gemm_inner_loop_m1xn1_<type>rcptex1c_<type>rcp1xte_<type>_<isa>(
+    size_t m, size_t n, size_t k, const <type> *a, size_t rsa1, size_t csa1,
+    const <type> *b, size_t rsb1, size_t csb1);
+```
+As with the fused kernel application functions, these functions will compute partial tiles if `m` or `n` is not a multiple of `ETE`.
```
ericlove: Add a reminder that the caller is responsible for ensuring that `0 <= m <= m1 * ETE` and `0 <= n <= n1 * ETE`.

src/gemm/xsfmm/fused-kernel-API.md
```diff
@@ -0,0 +1,378 @@
...
+
+Since the fused kernels act directly on the tile state, they must be aware of which tiles each inner loop function writes its output to.
+The inner loop functions obey the following tile allocation scheme: if `m1 == n1`, tiles are arranged by increasing index in row-major order; otherwise, if `m1 < n1`, tiles are arranged by increasing index in column-major order.
```
ericlove: I guess I never noticed this before but: why not just universally row-major order?  Did this simplify the implementation somehow?

src/gemm/xsfmm/fused-kernel-API.md
```diff
@@ -0,0 +1,378 @@
...
+```
+SKL_XSFMM_INOUT
+SKL_FUNC_PRIVATE void
+skl_gemm_inner_loop_m1xn1_<type>rcptex1c_<type>rcp1xte_<type>_<isa>(
+    size_t m, size_t n, size_t k, const <type> *a, size_t rsa1, size_t csa1,
+    const <type> *b, size_t rsb1, size_t csb1);
+```
+As with the fused kernel application functions, these functions will compute partial tiles if `m` or `n` is not a multiple of `ETE`.
+
+Since the fused kernels act directly on the tile state, they must be aware of which tiles each inner loop function writes its output to.
```
ericlove: One thing I'd like to find a way of clarifying is the ambiguity between the two kinds of "fused kernel" functions: those that operate on a specific register layout, and those that process an arbitrary 1x1 tile that can be applied automatically.  Mostly we assume users will write the latter, but here you seem to refer to the former.  Maybe we could use a term like "tile-agnostic fused kernels" and "tile-specific"? Open to ideas...

ericlove (commented):
I think this document a great addition! Really happy to see an explanation of this API.

I have a mix of low-level and high-level feedback below. Feel free to reject the high-level ones if you think it would take too long to address or if you don't agree with them--just tying to help tighten up the narrative.

src/gemm/xsfmm/fused-kernel-API.md
myeh01: > 2. There is no guaranteed order in which the output matrix tiles will be covered (presumably?)
> 3. Because SKL functions are single-thread, the fused `kernel` will only be called from the same thread as the GEMM, so it is safe to mutate state in `params` without synchronization.

Would there be issues mutating state in `params` because of 2.?

src/gemm/xsfmm/fused-kernel-API.md
myeh01: For `m1` x `n1` tilings with `m1 >= n1`, using row-major order has the nice property that the orderings for a fixed `n1` are just extensions or truncations of each other. I think this makes the inner loop code a bit easier to write and maintain. But for performance reasons, we implement the inner loops for `m1 <= n1`, so we end up with the above tile ordering. Maybe it makes sense to just make everything column-major.

src/gemm/xsfmm/fused-kernel-API.md
myeh01: I see. I think the naming for these is inconsistent. In some of them, one `<type>` refers to the tile type and the other `<type>` to `c`'s type. In others, the first `<type>` is the type of `a`/`b` and the second that of `c`. Besides our usual convention for types in kernel names, we might run into multiple definition errors for functions which only depend on the accumulator type (like `tile_load`, `tile_zero`, any fused kernel) depending on how `SKL_FUNC_PRIVATE` is defined. We discussed this a bit a while back, and I think the solution was to include the input type in the name, but maybe we could use the extension `_xsfmm32a32f` in some way even though the `tile_load`, `tile_zero`, fused kernels, etc. only need `xsfmmbase`.

myeh01: Might need some input on the naming scheme. This is a list of all typedefs/kernels related to the fused API and the current names in [#14](https://github.com/sifive/skl/pull/14):
- Tile zero function (`skl_gemm_tile_zero_f32_f32_xsfmmbase`)
- Tile load function (`skl_gemm_tile_load_f32rcptexterc_f32_xsfmmbase`)
- Fused kernel params struct (`skl_gemm_alpha_beta_scaling_params_f32_f32rcptexterc_xsfmmbase_t`)
- Fused kernel typedef (`skl_gemm_fused_kernel_f32_f32rcptexterc_xsfmmbase_t`)
- Fused kernel (`skl_gemm_alpha_beta_scaling_f32_f32rcptexterc_xsfmmbase`). There are also specializations of fused kernels like `skl_gemm_alpha_beta_scaling_m2_f32_f32_xsfmmbase`, but I think these will remain private and the fused kernel will dispatch to the appropriate one.
- Fused kernel application function (`skl_gemm_apply_fused_f32_f32rcptexterc_xsfmmbase`)

I guess the extension isn't enough to distinguish between `a`/`b` types, e.g. `xsfmm32a8f`. For the above, what do you think about using `<a/b type>_<c type>_<isa>` (e.g. `f8e4m3_f32_xsfmm32a8f` or `f32_f32_xsfmm32a32f`).

I think these names work as is:
- Inner loop typedef (`skl_gemm_inner_loop_f32rcptex1c_f32rcp1xte_f32_xsfmm32a32f_t`)
- Inner loop functions (`skl_gemm_inner_loop_2x2_f32rcptex1c_f32rcp1xte_f32_xsfmm32a32f`)
- Tiling application function (`skl_gemm_apply_tiling_f32rcptex1c_f32rcp1xte_f32rcptexterc_xsfmm32a32f`)

src/gemm/xsfmm/fused-kernel-API.md
myeh01: I have moved the inner loop section first. Hopefully this addresses everything in this comment.

src/gemm/xsfmm/fused-kernel-API.md
myeh01: I tried rewording this a bit so it's clearer hopefully:
```
Other functions which initialize the tile state for or consume the output of an inner loop function must know the tile register allocation it uses.
```
I was trying to refer to fused kernel application function in the original version.

src/gemm/xsfmm/fused-kernel-API.md
myeh01: I have added a few lines under the Fused Kernel API section:
```
Once a fused kernel has been implemented, the output of any inner loop function can be processed by applying it to each tile of the output.
This single-tile API avoids the need for kernel authors to know how each inner loop function arranges data in matrix registers.
```
Let me know what you think.

src/gemm/xsfmm/fused-kernel-API.md
myeh01: I have added a short, unoptimized inline assembly implementation to this example illustrating the use of `sf.vtmv.v.t` instructions.

myeh01:
@ericlove Thanks for all the feedback! I think I've addressed most of it, but please let me know what you think of the updated version. I still need some input on the naming conventions for some of these functions https://github.com/sifive/skl/pull/125#discussion_r3581439206.

src/gemm/xsfmm/fused-kernel-API.md
myeh01: Related to this, I think we may be able to use a `void *` type for `c` in the fused kernel API. This would allow us to reuse the fused kernel application, tiling application, and outer loop functions for different `c` types. So each source file would be centered around the input and tile accumulator types (basically which `xsfmm?a?[fi]` extension is used) rather than the input and output types . I did a mock up [here](https://github.com/sifive/skl/blob/dev/myeh/xsfmm_alpha_beta_void_c/src/gemm/xsfmm/gemm_f32c_f32_f32_xsfmm32a32f.c). Please let me know what you think. If it looks reasonable, maybe I could make this change in a subsequent PR after #14 is merged.

myeh01: Proposed naming scheme:
Tile zero function
`skl_gemm_tile_zero_<a/b type>_<accum type>_xsfmmbase`
Tile load function
`skl_gemm_tile_load_<a/b type>_<c type>rcptexterc_<accum type>_xsfmmbase`

Inner loop
`skl_gemm_inner_loop_<m1>x<n1>_<a type>rcptex1c_<b type>rcp1xte_<accum type>_<isa>`

Fused kernel params struct
`skl_gemm_<op>_params_<a/b type>_<accum type>_<c type>rcptexterc_xsfmmbase_t`
Fused kernel
`skl_gemm_<op>_<a/b type>_<accum type>_<c type>rcptexterc_xsfmmbase`
Fused kernel typedef
`skl_gemm_fused_kernel_<a/b type>_<accum type>_<c type>rcptexterc_xsfmmbase_t`

Fused kernel application function
`skl_gemm_apply_fused_<a/b type>_<accum type>_<c type>rcptexterc_xsfmmbase`
Tiling application function
`skl_gemm_apply_tiling_<a type>rcptex1c_<b type>rcp1xte_<accum type>_<c type>rcptexterc_<isa>`

The type of `a`/`b` is included in some way in each function name to prevent multiple definition errors. Also see my previous comment about potentially using a `void *` type for `c`; I have made the accumulator type and type of `c` distinct in the above names.

myeh01:
@ericlove I think is ready for review. Please see in particular these two comments:
https://github.com/sifive/skl/pull/125#discussion_r3606389150
https://github.com/sifive/skl/pull/125#discussion_r3635410762

src/gemm/xsfmm/fused-kernel-API.md
ericlove: Yes, this seems like a good scheme to me.

src/gemm/xsfmm/fused-kernel-API.md
```diff
@@ -0,0 +1,450 @@
...
+ mt8  mt9 mt10 mt11
+mt12 mt13 mt14 mt15
+```
+
+## The Fused Kernel API
+Fused kernels operate on a single tile in the Xsfmm tile state and a single block of `C`.
+Once a fused kernel has been implemented, the output of any inner loop function can be processed by applying it to each tile of the output.
+This single-tile API avoids the need for kernel authors to know how each inner loop function arranges data in matrix registers.
+
+The API for fused kernels is
```
ericlove, line 90: Slightly easier (IMO) to read as:
```suggestion
A fused kernel must have the following function signature:
```

src/gemm/xsfmm/fused-kernel-API.md
ericlove: It would depend on the kernel, right?  I think most applications will be commutative and associative operations like min/max or sum reduction (modulo floating-point subtleties).  We do have an example that relies on this assumption, after all.

src/gemm/xsfmm/fused-kernel-API.md
```diff
@@ -0,0 +1,450 @@
...
+  float *c_block = c + row1 * rsc1 + col1 * csc1;
+  for (size_t i = 0; i < tm; ++i)
+    for (size_t j = 0; j < tn; ++j)
+      c_block[i * rsc0 + j * csc0] =
+          alpha * tss[i, j] + beta * c_block[i * rsc0 + j * csc0];
+}
+```
+This kernel is already provided by SKL.
+In actual code, `sf.vtmv.v.t` instructions are used in the inner loop to move rows or columns of the tile state to the vector register file.
+An unoptimized implementation of this kernel in inline assembly might look like:
```
ericlove, line 142: We should clarify what "unoptimized" means here, since fused kernel authors will immediately want to know what they should be doing for better performance.

I still think it would be good to add a general explanation at the head of the "examples" section that indicates (1) real code needs to use tile moves, not `tss[i,j]` pseudocode and (2) high-performance implementations will likely want to unroll and software-pipeline the tile->VRF moves to hide latency.  But this can be added later.

src/gemm/xsfmm/fused-kernel-API.md
```diff
@@ -0,0 +1,450 @@
...
+      [tn] "r"(tn)
+    : "v0", "v1", "v2", "v3", "v4", "v5", "v6", "v7", "v16", "v17", "v18",
+      "v19", "v20", "v21", "v22", "v23", "vtype", "vl", "memory");
+```
+
+#### Adding a bias
+The following kernel can be used to compute `C = A * B + bias`, where `bias` is a row vector that is added to each row of `A * B`:
+```
+typedef struct {
+  float *bias;
```
ericlove, line 174: Do we even need to bother defining a struct for this, since it has only one parameter?

src/gemm/xsfmm/fused-kernel-API.md
```diff
@@ -0,0 +1,450 @@
...
+        tss_tile[i0, j0] = c_block[i0 * rsc0 + j0 * csc0];
+    }
+    n_avl -= tn;
+  }
+  m_avl -= tm;
+}
+```
+Note that if `tss` has a column pattern, then the transpose of `C` is loaded into the tile layout with `rstss` interpreted as the column stride and `cstss` as the row stride, similar to the column pattern case for the fused kernel application functions.
+
+## Applying a Tiling to `C`
```
ericlove, line 381: We should add another clarification that SKL's GEMM dispatch function does this automatically; this section is just explaining how it works for illustration purposes.

ericlove (approved):
I think this looks great! I have some notes but they can really be treated as TODOs for follow-on edit PRs.  I think this is good enough to merge as-is.

PR #126, opened 2026-06-16
[CI] Parallelize build and test steps

hdelassus:
This PR parallelizes the build and test steps in GitHub Actions, to use the maximum number of cores available to the runner, which currently sits at 2.

This seems to bring a small 6% improvement over [the average 7m20s in the last month](https://github.com/sifive/skl/actions/metrics/performance). It is a bit disappointing but would hopefully scale with higher-core-count runners in the future.

I have also slightly optimized the clang-format step by removing a call to `egrep` and simplifying the command executed by `xargs`, by using a diff-filter and a glob pattern in the `git-diff` command.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,6 +27,9 @@ jobs:
       with:
         fetch-depth: 0
 
+    - name: Set CPU core count
+      run: echo "NPROC=$(nproc)" >> $GITHUB_ENV
+
     - name: Check license headers
       if: ${{ github.event_name == 'pull_request' }}
       run: |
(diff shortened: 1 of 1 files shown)


Aaron-Hutchinson:
Thanks for giving this some attention. We should soon be able to remove the "test only" and "benchmark only" steps of CI once the couple remaining harness updates are completed, which should shave off 2-3 minutes of CI time.

ebavier (approved):
Thanks, this looks good to me.

I would somewhat expect not much improvement from parallelism at the moment.  I've consistently seen a single test responsible for more than 80% of test time, and compiling a small number of C files takes not much time at all.

PR #129, opened 2026-06-18
[GEMM] Rename Xsfvqdotq GEMM kernels

myeh01:
This PR renames the Xsfvqdotq GEMM kernels to include the tile shape in accordance with [packed-gemm.md](https://github.com/sifive/skl/blob/main/src/gemm/packed-gemm.md).

### DIFF
--- a/src/gemm/xsfvqdotq/gemm_i8rcp1x4_i8p4x1c_i32_xsfvqdotq.h
+++ b/src/gemm/xsfvqdotq/gemm_i8rcp1x4_i8p4x1c_i32_xsfvqdotq.h
@@ -54,17 +54,12 @@ extern "C" {
  *
  * This kernel uses the SiFive Xsfvqdotq extension for vector quad widening 4D
  * dot product operations to achieve high performance on 8-bit integer data.
- * When A satisfies certain alignment requirements (it is 4-byte aligned
- * and rsa1 and csa1 are multiples of 4), the kernel automatically dispatches to
- * optimized internal implementations:
- * - For m=1: uses an internal GEMV kernel
- * - For m>1: uses tiled GEMM kernels
+ * Works best when A is 4-byte aligned and rsa1 and csa1 are multiples of 4.
  *
  * @note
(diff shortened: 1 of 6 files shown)


src/gemm/xsfvqdotq/gemm_i8rcp1x4_i8p4x1c_i32_xsfvqdotq.h
```diff
@@ -0,0 +1,84 @@
...
+ * and rsa1 and csa1 are multiples of 4), the kernel automatically dispatches to
+ * optimized internal implementations:
+ * - For m=1: uses an internal GEMV kernel
+ * - For m>1: uses tiled GEMM kernels
+ *
+ * @note
+ * If A does not meet the alignment requirements stated above, the kernel
+ * falls back to an unaligned 1xm4 implementation. This implementation is
+ * unlikely to get good performance since it must handle misaligned loads from
+ * A.
```
Aaron-Hutchinson: I know this existed before this PR, but should internal implementation details be mentioned in the public API docstring? This seems like a case where less is more to ease user understanding and avoiding forgetting to update the docstring if the implementation changes. Maybe something like:

```suggestion
 * Works best when A is 4-byte aligned and rsa1 and csa1 are multiples of 4.
 *
 * @note
 * If A does not meet the alignment requirements stated above, the kernel
 * falls back to an unaligned implementation that is
 * unlikely to get good performance.
```

Aaron-Hutchinson (dismissed):
Looks great to me!

Aaron-Hutchinson (dismissed):

src/gemm/xsfvqdotq/gemm_i8rcp1x4_i8p4x1c_i32_xsfvqdotq.h
myeh01: Thanks, I updated it according to your suggestion. (Will need a reapproval though.)

Aaron-Hutchinson (approved):

PR #130, opened 2026-06-18
[Test] Ports GELU test suite.

ebavier:
* test/gelu/gelu_f32.c: Delete old tests, replaced by...
* test/gelu/skl_gelu_f32_zve32f.c: ... a suite using the new testing framework.
* CMakeLists.txt (gelu_f32): Remove old suite.
* test/CMakeLists.txt (skl_gelu_f32_zve32f): Add new suite.


test/gelu/skl_gelu_f32_zve32f.c
```diff
@@ -0,0 +1,80 @@
...
+        // magnitude inputs less than 0.  Note also that the
+        // formulation with FMA above produces the wrong sign for such
+        // inputs.
+        double e = erfc(-h * sqrt2);
+        out[i] = (float)(h * e);
+      }
+    }
+  }
+}
+#endif
```
hdelassus: Would it make sense to put this in `ref/gelu/`?

ebavier: We recently replaced the logistic reference functions with implementations that calculate in double precision.  I accepted that but am still slightly conflicted about the purpose of the reference function collection: to provide accurate implementations suitable for tests to compare results against, or to provide something subjectively "close to the definition" or "simple to understand"?

I think we could adjust `ref/gelu/gelu_f32.c` to satisfy both.  Using double-precision `sqrt2`, `erfc`, and `fma`, the reference function would be correctly-rounded in f32 and still "simple"/"close to definition".

WDYT?

hdelassus: Sounds good to me! By "adjust" you also meant "create", I suppose, since there is no `ref/gelu` at the moment.

ebavier: Ha, yes, I guess you're right!  (I have a `ref/gelu` locally, but this would be a good time to check it in 😅)

ref/gelu/gelu_f32.c
```diff
@@ -0,0 +1,22 @@
...
+// Licensed under the MIT License.
+// See LICENSE file in the project root for full license information.
+// SPDX-License-Identifier: MIT
+
+#include "skl-common.h"
+#include <math.h>
+#include <stddef.h>
+
+/* This reference is based on the observation that the GELU is a
+   calculated offset from the RELU */
```
hdelassus: nit: RELU -> ReLU

hdelassus (dismissed):
LGTM

hdelassus (approved):

PR #131, opened 2026-06-22
[Test] Fix clang-tidy error about assertion

peichialin:
The error will be reported by newer toolchain. Here is the error message.
```
test/skl-test.h:626:1: error: found assert() that could be replaced by static_assert() [misc-static-assert,-warnings-as-errors]
```


ericlove (approved):
Seems fine to me, and we will delete this file soon anways.

Aaron-Hutchinson (approved):

PR #133, opened 2026-06-23
[DepthwiseConv2D] Remove zbb dependency.

peichialin:
Rewrite the way of getting smaller length and let compiler decide the code generation.
When __riscv_zbb is defined, `x < y ? x : y` will be compiled into `minu` instruction.
Otherwise, it will be compiled into `bltu` and `mv` instructions.


hdelassus (approved):
LGTM

PR #135, opened 2026-06-23
[Test] Fix performance for SKL_TEST_STATIC mode

ericlove:
Previously, the "static test data" mode repeated copies of `(BUF)->static_data` using a single `for` loop and an index modulo `(BUF)->static_data_len`, which compiled to scalar code with integer remainder operations, and negated the efficiency benefit of static mode.  This PR fixes it using `memcpy` to move entire chunks at a time, avoiding any explicit remainder.

Unfortunately, the changes to the `SKL_TEST_BUF_CREATE` macro pushed some tests (in `depthwise_conv2d`) over the preset cognitive complexity limit, causing compilation errors.  After spending some time (too much) trying to do a better job, I have given up and just hacked those messages out with `// NOLINTBEGIN` comments.  (Alas, outlining the copy logic to its own function resulted in _static analyzer errors_  so I just gave up.)

### DIFF
--- a/test/skl-test-driver.h
+++ b/test/skl-test-driver.h
@@ -464,8 +464,16 @@ typedef enum {
                    "static data from %p (len = %zd)\n", (BUF)->static_data,    \
                    (BUF)->static_data_len);                                    \
       if ((BUF)->static_data) {                                                \
-        for (size_t i = 0; i < (BUF)->len; ++i) {                              \
-          (BUF)->data[i] = (BUF)->static_data[i % (BUF)->static_data_len];     \
+        /* Copy static data in chunks for efficiency */                        \
+        size_t offset = 0;                                                     \
+        for (size_t remaining = (BUF)->len; remaining > 0;) {                  \
+          size_t n_copy = (remaining < (BUF)->static_data_len)                 \
+                              ? remaining                                      \
+                              : (BUF)->static_data_len;                        \
+          memcpy((BUF)->data + offset, (BUF)->static_data,                     \
+                 n_copy * sizeof(TYPE));                                       \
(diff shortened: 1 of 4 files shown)


myeh01 (approved):
Thanks! LGTM.

test/skl-test-driver.h
```diff
@@ -464,8 +464,16 @@ typedef enum {
...
       if ((BUF)->static_data) {                                                \
-        for (size_t i = 0; i < (BUF)->len; ++i) {                              \
-          (BUF)->data[i] = (BUF)->static_data[i % (BUF)->static_data_len];     \
+        /* Copy static data in chunks for efficiency */                        \
+        size_t offset = 0;                                                     \
+        for (size_t remaining = (BUF)->len; remaining > 0;) {                  \
+          size_t n_copy = (remaining < (BUF)->static_data_len)                 \
+                              ? remaining                                      \
+                              : (BUF)->static_data_len;                        \
+          memcpy((BUF)->data + offset, (BUF)->static_data,                     \
```
Aaron-Hutchinson, line 473: What's the behavior when `static_data_len` is mistakenly set to 0? Will this loop endlessly, or does execution halt before reaching this point?

ericlove: Yes, it will result in an infinite loop, which I believe is an improvement over the status quo ante.

Aaron-Hutchinson (approved):
I don't see any test that currently uses the `SKL_TEST_STATIC` initialization mode, and I don't think we should be relying on consumers of SKL to test something as simple as this. But this is nothing new to this PR and the change LGTM.

PR #136, opened 2026-06-23
Fix builds for Xsfmmbase

hdelassus:
bafc8c1 used SKL_RESTRICT in headers that didn't include skl-common.h, leading to build failures when Xsfmm extensions were enabled.

EDIT: I think bafc8c1 actually exposed the fact that some headers didn't include skl-common.h by removing includes from skl.h.

This commit adds #include's where it was found they were necessary.


myeh01 (approved):
Thanks! LGTM.

ericlove (approved):

PR #138, opened 2026-06-24
Update Pack CODEOWNERS

Harlen126:
Update CODEOWNERS for `/src/pack/rvv` and `/test/pack`


peichialin (approved):
LGTM~ But we should wait for @myeh01 's feedback before merge.

myeh01 (approved):
LGTM!

PR #139, opened 2026-06-24
[Test] Refactor `SKL_TEST_BUF_CREATE`

Aaron-Hutchinson:
This refactors the `SKL_TEST_BUF_CREATE` macro to reduce cognitive complexity and resolve many suppressed clang-tidy errors that have come up recently originating from it.

The approach I've currently taken is to extract each initialization case into (families of) dedicated functions, parametrized by type. `SKL_TEST_BUF_CREATE` is now effectively just dispatch into these functions.

We could reduce clang-tidy's measurement of complexity further by similarly replacing the whole `SKL_TEST_BUF_CREATE` macro with a family of `skl_test_buf_create_{type}` functions, but I'm not sure this _actually_ reduces cognitive complexity and would additionally require updating all call sites.

The circular copy optimized in #135 was modified slightly due to the type erasure introduced in this PR.

### DIFF
--- a/test/skl-test-driver.h
+++ b/test/skl-test-driver.h
@@ -367,10 +367,10 @@ enum {
 // Helper macros for generating random values (used internally by
 // SKL_TEST_BUF_CREATE)
 
-#define SKL_TEST_RANDOM_FLOAT_(TYPE, MIN, MAX, LEN)                            \
+#define SKL_TEST_RANDOM_FLOAT_(TYPE, MIN, MAX)                                 \
   ((TYPE)(TYPE)((MIN) + ((float)rand() / (float)RAND_MAX) * ((MAX) - (MIN))))
 
-#define SKL_TEST_RANDOM_INT_(TYPE, MIN, MAX, LEN)                              \
+#define SKL_TEST_RANDOM_INT_(TYPE, MIN, MAX)                                   \
   ((TYPE)(rand() % ((MAX) - (MIN) + 1) + (MIN)))
 
 /**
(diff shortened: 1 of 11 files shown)


test/skl-test-driver.h
```diff
@@ -366,38 +367,119 @@ enum {
...
+  ((TYPE)(rand() % ((MAX) - (MIN) + 1) + (MIN)))
 
-#define SKL_TEST_BUF_RANDOM__Float16(TYPE, MIN, MAX, LEN)                      \
-  SKL_TEST_BUF_RANDOM_FLOAT_(TYPE, MIN, MAX, LEN)
-#define SKL_TEST_BUF_RANDOM_float(TYPE, MIN, MAX, LEN)                         \
-  SKL_TEST_BUF_RANDOM_FLOAT_(TYPE, MIN, MAX, LEN)
-#define SKL_TEST_BUF_RANDOM_double(TYPE, MIN, MAX, LEN)                        \
-  SKL_TEST_BUF_RANDOM_FLOAT_(TYPE, MIN, MAX, LEN)
+/**
+ * @brief Populate a buffer with random data.
```
Aaron-Hutchinson, line 377: I don't know the right way to document a function-like macro that itself defines a function. I think for Doxygen to recognize this annotation correctly we'd have to document each of the individual `SKL_TEST_BUF_RANDOM_IMPL` calls below, but that seems... absurd?

ericlove: I think this is fine.

ericlove (approved):
A welcome improvement. LGTM. Thanks!

PR #141, opened 2026-07-01
[DepthwiseConv2D][Test] Fix buffer size and filter shape report.

peichialin:
* Fix buffer size for i8 tests.
* Fix filter shape in i8/f16/f32 report.


Harlen126 (approved):
LGTM!

PR #142, opened 2026-07-01
[GEMM] Add `f32_f32_f32` RVV reference kernel

Aaron-Hutchinson:
This adds a simple RVV reference GEMM kernel written in intrinsics for `f32_f32_f32` types, not tuned to any specific microarchitecture. The goal for it is to be easy to understand and that it gets good (but likely not optimal) performance with many CPUs.

I compared performance of the added reference kernel to the existing  `skl_gemm_f32_f32_f32_zve32f_x390` on X390 and found that the reference version is only slightly worse when using memory port allocated matrices, but substantially worse when using CLP memory. This is not unexpected.

This is the second of many such coming kernels (with the first being `skl_gemm_bf16_bf16_f32_zvfbfwma` already in the codebase).


ericlove (approved):
Looks like a good reference version to me 👍

PR #144, opened 2026-07-08
[GEMM][Bugfix] Fix type of A matrix values in `skl_gemm_i8_i8_i32_zve32x_x390`

Aaron-Hutchinson:
The fixes a minor bug introduced in #123.

The intermediate scalar values of the A matrix are typed as `int8_t` in the implementation. This gives the correct result in normal circumstances, but if the compiler decides to save and restore registers between loop iterations (in debug builds, for examples), it may think only the least significant 8 bits are important whereas 16 bits are actually needed since the kernel takes advantage of the sign extended widening. The fix here is to change the type to `int16_t`.


hdelassus (dismissed):
This is tricky, nice catch. Should we rename those `a`'s as `a..w`'s, since they are effectively used as widened values?

Aaron-Hutchinson:
> Should we rename those `a`'s as `a..w`'s, since they are effectively used as widened values?

I like that idea! I made the change, thanks!

hdelassus (approved):
LGTM!

PR #145, opened 2026-07-13
[Test] Add fast pseudorandom number generation

Aaron-Hutchinson:
This adds new functionality to the test infrastructure for generating pseudorandom numbers as a replacement for the existing functionality using `rand()`.

The motivation for this work is that sequential calls to `rand()` to fill an array with random values can be very slow in cycle-accurate simulators to the point of being unusable, so using an alternative vectorized implementation can help to re-enable using random values that don't have to be precomputed and stored.

The current implementation uses an [LFSR-based](https://en.wikipedia.org/wiki/Linear-feedback_shift_register) approach to generate random `uint32_t` values, which can be used as a basis for other datatypes.

### Draft status

Currently in draft status while the implementation is ironed out. I'd especially like feedback on the following aspects of the current version of this work:

1. Initialization of the vectorized implementation. The current implementation uses a precomputed array of values for the initial LFSR state. I did this to try to maximize the randomness, but this goes against the goal of reducing reliance on precomputed arrays. It would be better to initialize it at runtime, but the question is how:
    - They could be initialized to the values currently defined in this PR by using something similar to the `compute_initial_lfsr_state` function, but I think this would be quite slow.
    - We could initialize it to something like `vid.v`, but these elements aren't guaranteed to be far apart in the LFSR cycle so there's some risk of repetition if we get unlucky. Not sure what degree of "randomness" we'd like this feature to attain, so maybe this isn't a big concern.
    - Other suggestions?
2. Extending the current 32 bit LFSR to other SEW. The LFSR in the current commit is for 32 bit values. Options to support other types:
    - It should be easy to support SEW<32 using a narrowing shift, but this may be inefficient compared to dedicated implementations for each SEW.
    - Write separate implementations for each SEW to ensure maximal length vectors are used for each SEW. This would likely increase code size a lot.


test/skl-rng.c
```diff
@@ -0,0 +1,566 @@
...
+ *
+ * The addr, min, and max values are type erased for API compatibility.
+ */
+static inline void lfsr_f32_kernel(vuint32m8_t rand_v, size_t vl, void *addr,
+                                   void *min, void *max) {
+  float min_f32 = *(float *)min;
+  float max_f32 = *(float *)max;
+  float *addr_f32 = (float *)addr;
+
+  vfloat32m8_t arr_v = __riscv_vfcvt_f_xu_v_f32m8(rand_v, vl);  // (0, 2^32)
```
Aaron-Hutchinson, line 510: Not sure how much of the randomness is lost by this line or whether there's a better approach to convert the input to float types.

PR #146, opened 2026-07-14
[GEMM] Add `f16_f16_f16` RVV reference kernel

Aaron-Hutchinson:
This adds an RVV reference version of non-widening F16 GEMM in intrinsics.

When benchmarking with X390, performance of the added reference kernel is on par with the X390 specialization in the warmed cache + memory port allocated matrices + large problem sizes, but falls behind significantly in other scenarios. This is expected for a reference kernel.


myeh01 (approved):
LGTM

hdelassus (approved):
LGTM

PR #147, opened 2026-07-15
[GEMM] Add `f64_f64_f64` RVV reference kernel

Aaron-Hutchinson:
This adds an RVV reference version of F64 GEMM in intrinsics.

I compared the performance of the reference version vs the existing X390 specialization on X390 and found that the reference version actually performed slightly better in the large problem size + memory port allocated matrices + warmed cache setting, but fell behind in other scenarios. There may still be room for improvement in the X390 specialized kernel.


hdelassus (approved):
LGTM

PR #148, opened 2026-07-16
[GEMM] Add `f16_f16_f32` RVV reference kernel

Aaron-Hutchinson:
This adds a widening F16 RVV GEMM reference kernel in intrinsics.

Compared to the X390 specialized kernel `skl_gemm_f16_f16_f32_zvfh_x390`, the added reference kernel obtains about 87% of the performance of the specialization on X390 in the unwarmed cache + memory port allocated matrices + large problem size setting, and falls further behind the specialization in other scenarios.


hdelassus (approved):
LGTM

PR #149, opened 2026-07-20
[GEMM] Add `i8_i8_i32` RVV reference kernel

Aaron-Hutchinson:
This adds a quad-widening int8 RVV GEMM kernel written in intrinsics.

I used the approach of #123: load from B and widen with `vwcvt`, then use the implicit widening of the scalar load instruction to perform the `vwmacc`s from int16 into int32. Tile size remains at 4xm4 currently.

Compared to the X390 specialization, the reference kernel being added gets about 50% of the performance of the specialization when benchmarked with large problems on X390.


hdelassus (approved):
The explicit cast to int16_t effectively being a no-op, looks good to me.

Aaron-Hutchinson:
> The explicit cast to int16_t effectively being a no-op, looks good to me.

Yeah, I was trying to emphasize that there is some implicit widening going on there, similar to the x390 kernel.

PR #150, opened 2026-07-23
[DepthwiseConv2D] F16/F32: Add missing return result of post_process_output_func.

peichialin:
Fix: Store back the return value of post_process_output_func.


Harlen126 (approved):
LGTM!

Koyingtw (approved):

PR #158, opened 2026-08-05
Flatten pack and transpose source trees

ericlove:
This PR flattens the directory structure of `src/pack` and `src/transpose`.  I think there is not a good reason to keep separate subdirectories for different ISAs in these kernel families, so the level of indirection is basically clutter in part of our API.

In a separate PR, I will propose a similar change for GEMM, which is the one place where such a subdivision _might_ make sense.  We can debate the final merits there.

Also, I noticed that the source file names in `pack/` are unique in that they are the only ones in this repo that prefix `skl_` to the start of the `.c`/`.h` file name.  I will open a PR to remove that as well.


src/transpose/transpose_e16_zve32x.c
Aaron-Hutchinson, line 1: Are these three RVV files dead code? I don't see associated headers or an `add_skl_src` for them.

CMakeLists.txt
```diff
@@ -172,9 +172,9 @@ add_skl_src(src/softmax/softmax_f32_xsfvfexp32e.c "xsfvfexp32e")
 add_skl_src(src/softmax/softmax_f32_xsfvfexpa.c "xsfvfexpa")
 add_skl_src(src/softmax/softmax_f32_zve32f.c "zve32f")
 
-add_skl_src(src/transpose/xsfmm/transpose_e16_xsfmmbase.c "xsfmmbase")
-add_skl_src(src/transpose/xsfmm/transpose_e32_xsfmmbase.c "xsfmmbase")
-add_skl_src(src/transpose/xsfmm/transpose_e8_xsfmmbase.c "xsfmmbase")
+add_skl_src(src/transpose/transpose_e16_xsfmmbase.c "xsfmmbase")
+add_skl_src(src/transpose/transpose_e32_xsfmmbase.c "xsfmmbase")
+add_skl_src(src/transpose/transpose_e8_xsfmmbase.c "xsfmmbase")
```
Aaron-Hutchinson, line 178: If we're doing this flattening, please also do it in `test` to keep it consistent with `src`.

peichialin:
> Also, I noticed that the source file names in pack/ are unique in that they are the only ones in this repo that prefix skl_ to the start of the .c/.h file name. I will open a PR to remove that as well.

Good catch! I didn't notice the file name difference. I agree with you that we should rename in the same convension as other kernels.

peichialin (approved):
LGTM~

Aaron-Hutchinson (commented):
I don't really see the subdirectories as an issue. I'm in favor of keeping them, but more importantly I think we should be as consistent as possible in each kernel family. I don't like the idea of doing this for `pack` and `transpose` but not for `gemm`.

peichialin:
@Harlen126 
Are you going to remove the original implementation of rvv transpose?

src/transpose/transpose_e16_zve32x.c
Harlen126: Exactly. I forgot to remove these files after moving the implementations to `pack`.
Will remove them after this PR gets merged.

Harlen126:
> Harlen Chen (@Harlen126) Are you going to remove the original implementation of rvv transpose?

I'll do that after this PR is merged. Thanks.

CMakeLists.txt
ericlove: I agree we should do it in test, but that is not a blocking issue because it is not really a part of the public interface the way this PR is.  That is a TODO.

Harlen126 (approved):
LGTM!

PR #159, opened 2026-08-05
Remove skl_ prefix in src/pack/ srcs

ericlove:
For consistency with the naming convention for source files in other kernel families, this PR removes the `skl_` prefix in `src/pack/skl_*.{h,c}`.

PR #161, opened 2026-08-06
Flatten source directories in `src/gemm/`

ericlove:
This applies the same change as https://github.com/sifive/skl/pull/159 but in the `src/gemm` kernel family.  We should consider separately whether we want to apply this change, since this is the one family where some sort of grouping might make sense.  But I do think it would be better to be consistent, and the flat version really isn't that bad.


.github/CODEOWNERS
```diff
@@ -17,9 +17,9 @@
 /src/depthwise_conv2d @peichialin
 /src/exp @ebavier
 /src/gelu @ebavier
-/src/gemm/rvv @Aaron-Hutchinson
-/src/gemm/xsfmm @myeh01
-/src/gemm/xsfvqdotq @france5289 @myeh01 @peichialin
+/src/gemm/*zve* @Aaron-Hutchinson
```
Aaron-Hutchinson: This will not catch any of the half precision kernels. Maybe just `*zv*`?

include/skl.h
Aaron-Hutchinson: Needs a rebase I assume.

Aaron-Hutchinson (commented):
I have #146 still open but ready to merge. Should this PR merge first?

.github/CODEOWNERS
ericlove: Should be fixed in rebase.

Aaron-Hutchinson (approved):
LGTM

myeh01 (approved):

PR #162, opened 2026-08-06
Remove unused rvv source code from transpose kernel

Harlen126:
The transpose rvv source code was copied to Pack already, so this PR is to remove the unused rvv source code from `Transpose`


myeh01 (approved):
LGTM

Harlen126:
Close this PR since https://github.com/sifive/skl/pull/163 has the same changes.

PR #163, opened 2026-08-06
[Pack] Move transpose kernels to pack

myeh01:
This PR moves the transpose kernels to the packing kernels directory:
- The reference transpose kernels have been moved to `ref/pack`.
- The Xsfmm transpose kernel source files have been moved to `src/pack` and renamed to `pack_e{16,8,32}_xsfmmbase.{c,h}`.
- The transpose tests have been moved to `test/pack`.

Some questions:
- Originally, we discussed keeping the `src/transpose` directory and only having a `README.md` in this directory documenting the kernels and pointing users to `src/pack` for the source files. Should we stick to this plan, or also move `src/transpose/README.md` to something like `src/pack/transpose.md` and completely removing `src/transpose`?
- @Harlen126 Should we keep the `README.md` for the reference transpose kernels (currently moved from `ref/transpose/README.md` to `ref/pack/README.md`)? None of the other reference directories have one.


ericlove (dismissed):
1. Move looks good to me 👍 
2. I think it's a fine idea to put a README in an otherwise empty `src/transpose` directory.
3. We should also update the `src/pack/README.md` to advertise prominently at the top that this directory covers normal transposition as well.
4. In a separate PR, we should clean up API documentation to clarify more consistently that `src` and `dst` must not overlap.  We do this for the pack kernels, but not transpose.

myeh01:
@ericlove I have added a brief mention of the transpose kernels and their documentation in the README: https://github.com/sifive/skl/pull/163/changes/62818a3dbf10fd04b8cc67cfe16acdf0ebedeff3

Harlen126:
> - The transpose tests have been moved to test/pack.

My original thought was not to move `test/transpose`, since we have a transpose kernel folder (although it's empty) and the test harness is independent of the pack's. But I'm fine with moving `test/transpose` into `test/pack`.

> Should we keep the README.md for the reference transpose kernels (currently moved from ref/transpose/README.md to ref/pack/README.md)? None of the other reference directories have one.

I think we can remove the reference transpose kernels' `README.md` since the function usage is simple.

ericlove (dismissed):

myeh01:
Need a reapproval.

Harlen126 (approved):
LGTM!

PR #164, opened 2026-08-06
[Sigmoid] Refactor from logistic and SiLU APIs

hdelassus:
This patch merges the vastly similar implementations of SiLU and logistic into a more general "Sigmoid" family of APIs that can be used to implement a variety of activation functions, like Logistic (Sigmoid), SiLU, Swish, GLU, or SwiGLU.

Thin wrappers around sigmoid APIs are added for the above activation functions, which allows the compiler to optimize redundant expressions, fold some conditionals, and so on.


ref/sigmoid/sigmoid_bf16.h
```diff
@@ -0,0 +1,86 @@
...
+ * Convenience wrapper around skl_sigmoid_bf16_ref.
+ */
+void skl_swish_bf16_ref(__bf16 *out, __bf16 beta, const __bf16 *x, size_t n);
+
+/**
+ * @brief Reference BF16 GLU: out = x * sigmoid(y).
+ *
+ * Convenience wrapper around skl_sigmoid_bf16_ref.
+ */
+void skl_glu_bf16_ref(__bf16 *out, const __bf16 *x, const __bf16 *y, size_t n);
```
ebavier: This is the only function wrapper where it's the second array argument that is forwarded to the logistic.

Could we perhaps swap the argument names, or adjust the definition to `y * sigmoid(x)` for consistency?

(I assume we have the same situation in the `src` non-reference kernels)

ref/sigmoid/sigmoid_f32.c
```diff
@@ -0,0 +1,67 @@
...
+#include <stdbool.h>
+#include <stddef.h>
+#include <stdint.h>
+
+/**
+ * Exponential function approximation on a vector of f32 floating-point values
+ * with a 1-ULP error bound in [-inf; 0].
+ *
+ * @note NaNs are not propagated.
+ */
```
ebavier: Update or remove comment.

src/sigmoid/README.md
```diff
@@ -0,0 +1,90 @@
+# Sigmoid Kernels
+
+The sigmoid (logistic) function is a widely-used activation function primitive in large language models (LLMs), computer vision, and stable diffusion neural networks.
```
ebavier:
```suggestion
Sigmoid functions are widely-used activation function primitives in large language models (LLMs), computer vision, and stable diffusion neural networks.
```

src/sigmoid/README.md
```diff
@@ -0,0 +1,90 @@
...
+- For `beta * x = -∞`, implementations return NaN.
+
+## Wrapper APIs
+
+For convenience and optimization, each `skl_sigmoid_<type>_<isa>` kernel ships with a set of thin wrappers that specialize the core sigmoid into common activation functions.
+They follow the pattern `skl_<op>_<type>_<isa>` and are defined in terms of the core sigmoid as follows:
+
+| Wrapper | Computes | In terms of `skl_sigmoid` |
+| --- | --- | --- |
+| `skl_logistic_<type>_<isa>(out, x, n)` | `sigmoid(x)` | `sigmoid(out, 1, x, NULL, NULL, 0, n)` |
```
ebavier:
```suggestion
| `skl_logistic_<type>_<isa>(out, x, n)` | `logistic(x)` | `sigmoid(out, 1, x, NULL, NULL, 0, n)` |
```

And elsewhere.  This is consistent with definitions at the top of this `README` and, I think, should lead to less confusion of terminology.

src/sigmoid/sigmoid_bf16_xsfvfbfa.c
```diff
@@ -0,0 +1,98 @@
...
+    u = __riscv_vfmadd_vv_bf16m8(u, s, c, vl);
+    /* 3. Assemble scaling factor */
+    vuint16m8_t t = __riscv_vreinterpret_v_bf16m8_u16m8(c);
+    t = __riscv_vwmaccu_vx_u16m8(t, 1 << 7, k, vl);
+    vbfloat16m8_t f = __riscv_vreinterpret_v_u16m8_bf16m8(t);
+    /* 4. Reciprocate denominator */
+    vbfloat16m8_t d = __riscv_vfadd_vv_bf16m8(f, u, vl);
+    vbfloat16m8_t r = __riscv_vfrec7_v_bf16m8(d, vl);
+    /* 5. Reconstruct */
+    vbfloat16m8_t nn = __riscv_vmerge_vvm_bf16m8(f, u, m, vl);
```
ebavier:
```suggestion
    vbfloat16m8_t o = __riscv_vmerge_vvm_bf16m8(f, u, m, vl);
    vbfloat16m8_t q = __riscv_vfmul_vv_bf16m8(o, r, vl);
```
I'm surprised the toolchain didn't previously complain about `n` being defined as `size_t` and later as ` vbfloat16m8_t`.

src/sigmoid/sigmoid_bf16_xsfvfbfa.c
```diff
@@ -0,0 +1,98 @@
...
+    vbfloat16m8_t q = __riscv_vfmul_vv_bf16m8(nn, r, vl);
+    /* 6. Optionally multiply by y */
+    if (y) {
+      vbfloat16m8_t yv = __riscv_vle16_v_bf16m8(y + i, vl);
```
ebavier: I think this `yv` should be `vy` instead?  To match the `vx` above.  That, or change the previous `vx` to `xv`.  Either way.  Same with `uv` below.

src/sigmoid/sigmoid_bf16_xsfvfbfa.h
```diff
@@ -0,0 +1,85 @@
...
+ * @param beta - Scaling factor for the exponential part.
+ * @param x - Array of input elements. If NULL, the function computes nothing.
+ * @param y - Optional array. If not NULL, the result is multiplied by y.
+ * @param up - Optional array. If not NULL, the result is multiplied by
+ *             (up + delta).
+ * @param delta - Bias added to up before multiplying.
+ * @param n - Number of elements to process.
+ *
+ * Computes, elementwise:
+ *   out = sigmoid(beta * x)
```
ebavier: See comment in `README.md`, I think these single-argument functions should be referenced by the name `logistic`.

src/sigmoid/sigmoid_bf16_xsfvfexp32e.c
```diff
@@ -0,0 +1,112 @@
...
+  size_t vl;
+  for (size_t i = 0; i < n; i += vl) {
+    vl = __riscv_vsetvl_e16m4(n - i);
+    vint16m4_t vx = __riscv_vle16_v_i16m4((int16_t *)x + i, vl);
+    /* 0. Observe, Orient, Convert, & Squash */
+    vbool4_t m = __riscv_vmslt_vx_i16m4_b4(vx, 0, vl);
+    vx = __riscv_vor_vx_i16m4(vx, (int16_t)0x8000, vl); /* copysign(x,-1) */
+    /* 1. Widen, scale by beta and halve */
+    vfloat32m8_t a = skl_sigmoid_bf16_xsfvfexp32e_vfwcvt_f_x_v_f32m8(vx, vl);
+    a = __riscv_vfmul_vf_f32m8(a, (float)beta * 0.5f, vl);
```
ebavier: It seems we might want to declare constraints on the value of `beta`?  Perhaps documented to assume `beta > 0`?  Otherwise, a `beta < 0` here could cause the sign of `a` to change from the intended.

(I also wonder if the half-scaling in `bf16_xsfvfbexp16e` could be fused with the beta-scaling as done here).

src/sigmoid/sigmoid_bf16_xsfvfexpa.c
```diff
@@ -0,0 +1,118 @@
...
+    vl = __riscv_vsetvl_e16m4(n - i);
+    vint16m4_t vx = __riscv_vle16_v_i16m4((int16_t *)x + i, vl);
+    /* 0. Observe, Orient, Clamp, & Convert */
+    vbool4_t m = __riscv_vmslt_vx_i16m4_b4(vx, 0, vl);
+    vx = __riscv_vor_vx_i16m4(vx, (int16_t)0x8000, vl);  /* copysign(x,-1) */
+    vx = __riscv_vmin_vx_i16m4(vx, (int16_t)0xc2ba, vl); /* fmax(x,-0x1.74p6) */
+    vfloat32m8_t a = skl_sigmoid_bf16_xsfvfexpa_vfwcvt_f_x_v_f32m8(vx, vl);
+    /* 1. Scale by beta */
+    if (beta != 1)
+      a = __riscv_vfmul_vf_f32m8(a, (float)beta, vl);
```
ebavier: `beta > 1` here could cause `a` to stray outside its intended clamped domain.  I believe that could cause the `vfexpa` later to return spurious inf or NaN.  Same issue here as in `xsfvfexp32e` wrt `beta < 0`.

src/sigmoid/sigmoid_bf16_xsfvfexpa.c
```diff
@@ -0,0 +1,118 @@
...
+    vfloat32m8_t Q = __riscv_vfmv_v_f_f32m8(O, vl);
+    vfloat32m8_t v = __riscv_vfmadd_vf_f32m8(a, R, Q, vl);
+    /* 3. Approximate 2⁸exp(v) */
+    vfloat32m8_t e = __riscv_sf_vfexpa_v_f32m8(v, vl);
+    /* 4. Reciprocate denominator */
+    vfloat32m8_t d = __riscv_vfadd_vf_f32m8(e, 0x1p-8f, vl);
+    vfloat32m8_t r = __riscv_vfrec7_v_f32m8(d, vl);
+    /* 5. Reconstruct */
+    vfloat32m8_t nn = __riscv_vfmerge_vfm_f32m8(e, 0x1p-8f, m, vl);
+    vfloat32m8_t q = __riscv_vfmul_vv_f32m8(nn, r, vl);
```
ebavier:
```suggestion
    vfloat32m8_t o = __riscv_vfmerge_vfm_f32m8(e, 0x1p-8f, m, vl);
    vfloat32m8_t q = __riscv_vfmul_vv_f32m8(o, r, vl);
```

src/sigmoid/sigmoid_bf16_xsfvfexpa.c
```diff
@@ -0,0 +1,118 @@
...
+                                         const __bf16 *x, const __bf16 *y,
+                                         const __bf16 *up, __bf16 delta,
+                                         size_t n) {
+  if (!x)
+    return;
+
+  size_t vl;
+  for (size_t i = 0; i < n; i += vl) {
+    vl = __riscv_vsetvl_e16m4(n - i);
+    vint16m4_t vx = __riscv_vle16_v_i16m4((int16_t *)x + i, vl);
```
ebavier: Do we want this variable to be `xi` to match the `yi` and `ui` below? (Same question in `bf16_xsfvfexp32e`)

src/sigmoid/sigmoid_bf16_zve32f.c
```diff
@@ -0,0 +1,124 @@
...
+    vl = __riscv_vsetvl_e16m4(n - i);
+    vint16m4_t vx = __riscv_vle16_v_i16m4((int16_t *)x + i, vl);
+    /* 0. Observe, Orient, Clamp, & Convert */
+    vbool4_t m = __riscv_vmslt_vx_i16m4_b4(vx, 0, vl);
+    vx = __riscv_vor_vx_i16m4(vx, (int16_t)0x8000, vl);  /* copysign(x,-1) */
+    vx = __riscv_vmin_vx_i16m4(vx, (int16_t)0xc2ba, vl); /* fmax(x,-0x1.74p6) */
+    vfloat32m8_t a = skl_sigmoid_bf16_zve32f_vfwcvt_f_x_v_f32m8(vx, vl);
+    /* 1. Scale by beta */
+    if (beta != 1)
+      a = __riscv_vfmul_vf_f32m8(a, (float)beta, vl);
```
ebavier: Same possible issue here wrt `beta > 1` as in `bf16_xsfvfexpa`, and `beta < 0`.

Easiest fix might be moving the conditional up; if `beta != 1` we'd need to do f32-conversion and beta-scaling right away.

src/sigmoid/sigmoid_bf16_zve32f.c
```diff
@@ -0,0 +1,124 @@
...
+    vuint32m8_t t = __riscv_vreinterpret_v_f32m8_u32m8(c);
+    k = __riscv_vsll_vx_u16m4(k, 8, vl);
+    t = __riscv_vwmaccu_vx_u32m8(t, 1 << 15, k, vl);
+    vfloat32m8_t f = __riscv_vreinterpret_v_u32m8_f32m8(t);
+    /* 5. Reciprocate denominator */
+    vfloat32m8_t d = __riscv_vfadd_vv_f32m8(f, u, vl);
+    vfloat32m8_t r = __riscv_vfrec7_v_f32m8(d, vl);
+    /* 6. Reconstruct */
+    vfloat32m8_t nn = __riscv_vmerge_vvm_f32m8(f, u, m, vl);
+    vfloat32m8_t q = __riscv_vfmul_vv_f32m8(nn, r, vl);
```
ebavier:
```suggestion
    vfloat32m8_t o = __riscv_vmerge_vvm_f32m8(f, u, m, vl);
    vfloat32m8_t q = __riscv_vfmul_vv_f32m8(o, r, vl);
```

src/sigmoid/sigmoid_bf16_zve32f.c
```diff
@@ -0,0 +1,124 @@
...
+
+  size_t vl;
+  for (size_t i = 0; i < n; i += vl) {
+    vl = __riscv_vsetvl_e16m4(n - i);
+    vint16m4_t vx = __riscv_vle16_v_i16m4((int16_t *)x + i, vl);
+    /* 0. Observe, Orient, Clamp, & Convert */
+    vbool4_t m = __riscv_vmslt_vx_i16m4_b4(vx, 0, vl);
+    vx = __riscv_vor_vx_i16m4(vx, (int16_t)0x8000, vl);  /* copysign(x,-1) */
+    vx = __riscv_vmin_vx_i16m4(vx, (int16_t)0xc2ba, vl); /* fmax(x,-0x1.74p6) */
+    vfloat32m8_t a = skl_sigmoid_bf16_zve32f_vfwcvt_f_x_v_f32m8(vx, vl);
```
ebavier:
```suggestion
    vint16m4_t xi = __riscv_vle16_v_i16m4((int16_t *)x + i, vl);
    /* 0. Observe, Orient, Clamp, & Convert */
    vbool4_t m = __riscv_vmslt_vx_i16m4_b4(xi, 0, vl);
    xi = __riscv_vor_vx_i16m4(xi, (int16_t)0x8000, vl);  /* copysign(x,-1) */
    xi = __riscv_vmin_vx_i16m4(xi, (int16_t)0xc2ba, vl); /* fmax(x,-0x1.74p6) */
    vfloat32m8_t a = skl_sigmoid_bf16_zve32f_vfwcvt_f_x_v_f32m8(xi, vl);
```

src/sigmoid/sigmoid_f16_xsfvfexp16e.c
```diff
@@ -0,0 +1,91 @@
...
+  const vfloat16m8_t one = __riscv_vfmv_v_f_f16m8(1, vlmax);
+  for (size_t i = 0; i < n; i += vl) {
+    vl = __riscv_vsetvl_e16m8(n - i);
+    /* 0. Load and scale by beta */
+    vfloat16m8_t vx = __riscv_vle16_v_f16m8(x + i, vl);
+    if (beta != 1)
+      vx = __riscv_vfmul_vf_f16m8(vx, beta, vl);
+
+    /* 1. Approximate exp(-|beta x|) */
+    vfloat16m8_t ex = __riscv_vfmul_vf_f16m8(vx, -0.5f16, vl);
```
ebavier:
```suggestion
    /* 1. Approximate exp(-|beta x|) */
    vfloat16m8_t ex = __riscv_vfmul_vf_f16m8(vx, -0.5f16 * beta, vl);
```

can we fuse this `vfmul` and the previous `vfmul` for `beta`?

src/sigmoid/sigmoid_f16_xsfvfexp16e.c
```diff
@@ -0,0 +1,91 @@
...
+    ex = __riscv_vfmul_vv_f16m8(ex, ex, vl);
+
+    /* 2. Approximate 1 / (1 + exp(-|beta x|)) */
+    const vfloat16m8_t d = __riscv_vfadd_vf_f16m8(ex, 1, vl);
+    vfloat16m8_t r = __riscv_vfrec7_v_f16m8(d, vl);
+    vfloat16m8_t t = __riscv_vfnmsub_vv_f16m8(d, r, one, vl); // 1 - x * r
+    r = __riscv_vfmadd_vv_f16m8(r, t, r, vl); // r + r * (1 - x * r)
+
+    /* 3. Calculate quotient */
+    const vbool2_t m = __riscv_vmflt_vf_f16m8_b2(vx, 0, vl);
```
ebavier: If beta-scaling and half-scaling are fused, this conditional will need to be moved up.

src/sigmoid/sigmoid_bf16_xsfvfexpa.c
```diff
@@ -0,0 +1,118 @@
...
+    vfloat32m8_t a = skl_sigmoid_bf16_xsfvfexpa_vfwcvt_f_x_v_f32m8(vx, vl);
+    /* 1. Scale by beta */
+    if (beta != 1)
+      a = __riscv_vfmul_vf_f32m8(a, (float)beta, vl);
+    /* 2. Reduce x ~ (k + j/64) ln2 */
+    const float R = -0x1.715476p0f; /* -1/ln2 */
+    const float O = 0x1.003b80p17f; /* 2^(24-6-1)+127-8 */
+    vfloat32m8_t Q = __riscv_vfmv_v_f_f32m8(O, vl);
+    vfloat32m8_t v = __riscv_vfmadd_vf_f32m8(a, R, Q, vl);
+    /* 3. Approximate 2⁸exp(v) */
```
ebavier: I believe this actually calculates `2⁻⁸exp(v)`

src/sigmoid/sigmoid_f16_xsfvfexpa_zvfh.c
```diff
@@ -0,0 +1,126 @@
...
+#if !defined(__riscv_xsfvfexpa) || !defined(__riscv_zvfh)
+#error This file requires the Xsfvfexpa and Zvfh extensions
+#endif
+
+#include "skl-common.h"
+#include <riscv_vector.h>
+#include <sifive_vector.h>
+#include <stddef.h>
+
+SKL_FUNC_PRIVATE vfloat16m8_t skl_leftexp_xsfvfexpa_zvfh_f16m8(vfloat16m8_t x,
```
ebavier: I'm slightly bothered that the f16 and f32 `xsfvfexpa` are the only kernels that split the exponentiation out into a separate function.  This is probably fine for now.

I believe these and the `zvfh` and `zve32f` kernels can do some of the same exponent-adjustment and scaling reduction techniques as used in the `bf16` kernels.  If we do that some day, then the exponentiation function can be subsumed.

src/sigmoid/sigmoid_f16_xsfvfexp16e.c
```diff
@@ -0,0 +1,91 @@
...
+    vfloat16m8_t res = __riscv_vfmul_vv_f16m8_mu(m, r, r, ex, vl);
```
ebavier: Can we use `q`, in place of `res`, for the quotient result for `f16` and `f32` kernels, as in the `bf16` kernels?

src/sigmoid/sigmoid_f32_xsfvfexp32e.c
```diff
@@ -0,0 +1,94 @@
...
+    vl = __riscv_vsetvl_e32m8(n - i);
+    /* 0. Load and scale by beta */
+    vfloat32m8_t vx = __riscv_vle32_v_f32m8(x + i, vl);
+    if (beta != 1)
+      vx = __riscv_vfmul_vf_f32m8(vx, beta, vl);
+
+    /* 1. Approximate exp(-|beta x|). Halve the argument to stay in range,
+     *    then square to recover the full exponential. Keeping the sign
+     *    negative saves a scalar register. */
+    vfloat32m8_t ex = __riscv_vfmul_vf_f32m8(vx, -0.5f, vl);
```
ebavier: Can fuse with beta-scaling?

src/sigmoid/sigmoid_f32_xsfvfexp32e.c
```diff
@@ -0,0 +1,94 @@
...
+    ex = __riscv_vfsgnj_vf_f32m8(ex, -0.5f, vl);
+    ex = __riscv_sf_vfexp_v_f32m8(ex, vl);
+    ex = __riscv_vfmul_vv_f32m8(ex, ex, vl);
+
+    /* 2. Approximate 1 / (1 + exp(-|beta x|)) */
+    const vfloat32m8_t d = __riscv_vfadd_vf_f32m8(ex, 1, vl);
+    vfloat32m8_t r = __riscv_vfrec7_v_f32m8(d, vl);
+    vfloat32m8_t t = __riscv_vfnmsub_vv_f32m8(d, r, one, vl); // 1 - x * r
+    r = __riscv_vfmadd_vv_f32m8(r, t, r, vl);    // r + r * (1 - x * r)
+    t = __riscv_vfnmsub_vv_f32m8(d, r, one, vl); // 1 - x * r
```
ebavier: The `logistic` kernels previously used a `t = t * t` update.  I believe we should still do that here.

src/sigmoid/sigmoid_f32_xsfvfexp32e.c
```diff
@@ -0,0 +1,94 @@
...
+#error This file requires the Xsfvfexp32e extension
+#endif
+
+#include "skl-common.h"
+
+#include <riscv_vector.h>
+#include <sifive_vector.h>
+#include <stddef.h>
+
+SKL_FUNC void skl_sigmoid_f32_xsfvfexp32e(float *out, float beta,
```
ebavier, line 16: The old f32 `logistic` kernel used inline assembly because we had trouble convincing toolchains not to spill registers.  Do we have that same problem here?  

The situation is more complicated for this generic `sigmoid`, but we can provide inline assembly specializations if we need.

src/sigmoid/sigmoid_f16_xsfvfexpa_zvfh.c
```diff
@@ -0,0 +1,126 @@
...
+
+  size_t vl;
+  for (size_t i = 0; i < n; i += vl) {
+    vl = __riscv_vsetvl_e16m8(n - i);
+    /* 0. Load and scale by beta */
+    vfloat16m8_t vx = __riscv_vle16_v_f16m8(x + i, vl);
+    if (beta != 1)
+      vx = __riscv_vfmul_vf_f16m8(vx, beta, vl);
+
+    /* Approximate exp(-|beta x|) */
```
ebavier: Could you format the comments in this kernel similarly to others?

src/sigmoid/sigmoid_f32_xsfvfexpa.c
```diff
@@ -58,7 +75,8 @@ SKL_FUNC void skl_silu_52u_f32_xsfvfexpa(float *out, const float *in,
 SKL_FUNC_PRIVATE vfloat32m8_t skl_leftexp_xsfvfexpa_f32m8(vfloat32m8_t x,
                                                           size_t vl) {
   /* 0. Clamp inputs to lower bound */
-  x = __riscv_vfmax_vf_f32m8(x, -0x1.9fe36ap6f, vl);
+  vbool4_t nn = __riscv_vmfeq_vv_f32m8_b4(x, x, vl); // Propagate NaN inputs
```
ebavier: Some other kernels specifically do not propagate NaNs, e.g. `f32_zve32f`.  Maybe this should be the expected behavior, documented in `README.md`.  In which case, we don't need to compute this mask.

src/sigmoid/sigmoid_f32_zve32f.c
```diff
@@ -92,3 +60,74 @@ SKL_FUNC_PRIVATE vfloat32m8_t skl_leftexp_zve32f_f32m8(vfloat32m8_t x,
...
+    const vfloat32m8_t e = skl_sigmoid_leftexp_zve32f_f32m8(
+        __riscv_vfsgnj_vf_f32m8(vx, -1, vl), vl);
+
+    /* Approximate 1 / (1 + exp(-|beta x|)) */
+    const vfloat32m8_t d = __riscv_vfadd_vf_f32m8(e, 1, vl);
+    const vfloat32m8_t one = __riscv_vfmv_v_f_f32m8(1, vl);
+    vfloat32m8_t r = __riscv_vfrec7_v_f32m8(d, vl);
+    vfloat32m8_t t = __riscv_vfnmsub_vv_f32m8(d, r, one, vl);
+    r = __riscv_vfmadd_vv_f32m8(r, t, r, vl);
+    t = __riscv_vfnmsub_vv_f32m8(d, r, one, vl); // 1 - x * r
```
ebavier: Let's use `t = t * t` here.

ebavier (changes requested):
This PR looks really great, a lot of work!  I'm impressed.  I think the factoring of these functions will work out quite well.

src/sigmoid/README.md
```diff
@@ -0,0 +1,90 @@
...
+This makes it possible to fuse SiLU (`y == x`, `beta == 1`, `up == NULL`), GLU (`y != x`, `up == NULL`) and SwiGLU (via `up`/`delta`) into a single kernel.
+
+## Kernel Naming Convention
+
+The kernel names follow the pattern: `skl_sigmoid_<type>_<isa>`
+
+where:
+
+- `<type>`: denotes the type of the input and output arrays.
+- `<isa>`: denotes the instruction set architecture extension(s) the kernel requires.
```
ericlove: I'm not sure we even need a whole section on this, because it is entirely subsumed by the SKL-level [naming convention](https://github.com/sifive/skl/tree/main/src#skl-kernel-naming-convention). The only detail worth clarifying is that each sigmoid kernel's I/Os are always all the same type.

ref/sigmoid/sigmoid_f16.c
```diff
@@ -0,0 +1,60 @@
...
+                                  size_t n) {
+  if (!x)
+    return;
+
+  for (size_t i = 0; i < n; i++) {
+    float o;
+    float a = x[i];
+    /* Sigmoid */
+    float b = beta * a;
+    if (b >= 0)
```
ericlove: Is there a reason why this code uses an `if` statement while the f32 version uses `copysign`?

ericlove (dismissed):
Interfaces and README look good 👍

ref/sigmoid/sigmoid_bf16.h
```diff
@@ -0,0 +1,86 @@
...
+ * Convenience wrapper around skl_sigmoid_bf16_ref.
+ */
+void skl_swiglu_bf16_ref(__bf16 *out, const __bf16 *gate, const __bf16 *up,
+                         __bf16 delta, size_t n);
+
+#ifdef __cplusplus
+} // extern "C"
+#endif
+#ifdef __cplusplus
+} // extern "C"
```
Aaron-Hutchinson: Extraneous closing brace here, and similarly in other files.

Is there a way we can ensure compilation with C++ to catch this kind of thing in CI?

src/sigmoid/README.md
```diff
@@ -0,0 +1,82 @@
...
+This makes it possible to fuse SiLU (`y == x`, `beta == 1`, `up == NULL`), GLU (`y != x`, `up == NULL`) and SwiGLU (via `up`/`delta`) into a single kernel.
+
+## Constraints
+
+- Input and output arrays may overlap only for in-place computation.
+- For maximum performance, input and output arrays should be naturally aligned.
+
+## Kernel List
+
+- FP32 APIs
```
ebavier, line 29: Do we have a reason, besides inertia, for using "FP32" in prose versus "f32" in APIs (not just here but everywhere)?  If not, I'd like to suggest we attempt to use "F32" and "F16" instead.  The meaning should be abundantly clear.

hdelassus: I believe FP32 is more commonly used than F32 (in prose) to refer to IEEE 754 binary32. On the other hand, I think `f32` is more commonly used in code.

ref/sigmoid/sigmoid_bf16.h
hdelassus: Thanks for catching this. I think there is no way to catch this in CI as of now. We could add a C++ smoketest just for that purpose.

ref/sigmoid/sigmoid_f16.c
hdelassus: No, I have refactored all reference kernels so that they look more similar to each other.

src/sigmoid/README.md
hdelassus: Removed.

src/sigmoid/sigmoid_f32_xsfvfexp32e.c
hdelassus: I see the same problem here. Can we defer that to a subsequent pull request?

src/sigmoid/sigmoid_bf16_xsfvfbfa.c
hdelassus: Updated.

src/sigmoid/sigmoid_bf16_xsfvfbfa.h
hdelassus: Updated.

src/sigmoid/sigmoid_bf16_xsfvfexp32e.c
hdelassus: Yes, that's my oversight. My guess is beta is always positive, so perhaps documenting it as a requirement is best?

src/sigmoid/sigmoid_bf16_xsfvfexpa.c
hdelassus: Yes, that's my oversight.

src/sigmoid/sigmoid_f32_xsfvfexp32e.c
ebavier: Deferring to later is fine with me.

src/sigmoid/sigmoid_f16_xsfvfexpa_zvfh.c
hdelassus: I've inlined them now.

src/sigmoid/sigmoid_f16_xsfvfexp16e.c
hdelassus: Done. Had to rename q->zi in zve32f/zvfh kernels.

src/sigmoid/sigmoid_f16_xsfvfexpa_zvfh.c
hdelassus: Done

src/sigmoid/sigmoid_f32_xsfvfexpa.c
hdelassus: Done.

src/sigmoid/sigmoid_bf16_xsfvfexp32e.c
```diff
@@ -0,0 +1,121 @@
...
+    vint16m4_t vx = __riscv_vle16_v_i16m4((int16_t *)x + i, vl);
+    vfloat32m8_t a;
+    vbool4_t m;
+    /* 0. Observe, Orient, Convert, & Squash */
+    if (beta != 1) {
+      a = skl_sigmoid_bf16_xsfvfexp32e_vfwcvt_f_x_v_f32m8(vx, vl);
+      a = __riscv_vfmul_vf_f32m8(a, (float)beta * 0.5f, vl);
+      m = __riscv_vmflt_vf_f32m8_b4(a, 0, vl);
+      a = __riscv_vfsgnj_vf_f32m8(a, -0x1.74p6f, vl);
+      a = __riscv_vfmax_vf_f32m8(a, -0x1.74p6f, vl);
```
ebavier: Unlike the `sigmoid_bf16_xsfvfexpa` kernel, here we don't need to clamp the input, since the scaling by half will bring all inputs into `sf.vfexp.v`'s valid domain.

We do still need the sign injection.  We could use `__riscv_vfsgnjn_vf_f32m8(a, 0.5f, vl)` if we still wanted to minimize the number of materialized scalar constants.

test/sigmoid/sigmoid_bf16_xsfvfexp32e.c
```diff
@@ -0,0 +1,91 @@
...
+#define SWISH_MIN (-40.f)
+#define SWISH_MAX (+6.f)
+#define SWISH_BETA (__bf16)2
+#define GLU_MIN (-40.f)
+#define GLU_MAX (+6.f)
+#define SWIGLU_MIN (-40.f)
+#define SWIGLU_MAX (+6.f)
+#define SWIGLU_DELTA (__bf16)0.5
+
+static void test_swish_bf16_xsfvfexp32e(__bf16 *out, const __bf16 *in,
```
ebavier: Could we defined a few additional `swish` tests to get coverage over `beta == 1.0` and `beta < 0` cases?

src/sigmoid/sigmoid_bf16_xsfvfexp32e.c
hdelassus: Oops, copy-paste error. Thank you

ref/sigmoid/sigmoid_bf16.h
hdelassus: I'm inclined to keep it as-is because [Pytorch](https://docs.pytorch.org/docs/2.13/generated/torch.nn.GLU.html) uses `GLU(a, b) = a * sigmoid(b)`, but open to hearing differing opinions.

ref/sigmoid/sigmoid_f32.c
hdelassus: Done

src/sigmoid/sigmoid_f32_xsfvfexp32e.c
hdelassus: This is slightly less accurate because of the difference in the second-order term, right? In the Zve32f the tests still pass, but with this kernel I get 6 ulps for logistic(-0x1.8f99e6p+3). Should I update the tolerance from 5 to 6?

ebavier: Yes, slightly less accurate.  The reciprocal itself ends up being ~1ULP, compared to 0.52ULP.  Tolerance of 6 seems right.

ebavier: And actually, I missed that this update changes the splitting strategy used, so the error will be greater in that regard too, using a `0.5` split.

I find maximum 7.395 ULP error in logistic at -0x1.09eecep+2.  Compare this to the previous algorithm's 4.517 ULP maximum.

ebavier (dismissed):
This cleaned up pretty nice 👍

ebavier (approved):

PR #165, opened 2026-08-06
[GEMM] Clean up reference GEMM kernels

myeh01:
This PR cleans up the packed GEMM kernels and tests.
Proposed changes:
- Rename `{a,b,c}_pack` simply to `{a,b,c}` in kernel APIs and in test harnesses
- Clean up comments/headers (TODO)
- Use SKL_XSFMM_TE in Xsfmm tests rather than assume ETE=64 (TODO, perhaps appropriate to make a separate PR?)

Edit: I think this will touch too many files, so I will break this up into several PRs.
- This PR will be limited to the reference kernels.
- The Xsfmm source files are already handled by the alpha/beta scaling PRs.
- Renaming `{a,b,c}_pack` to `{a,b,c}` and comment/header clean up for test harnesses will be another PR.
- If there's time, I'll open one more PR for SKL_XSFMM_TE.

### DIFF
--- a/ref/pack/transpose_e16.h
+++ b/ref/pack/transpose_e16.h
@@ -5,6 +5,7 @@
 
 #pragma once
 
+#include "skl-common.h"
 #include <stddef.h>
 #include <stdint.h>
 
--- a/ref/gemm/gemm_bf16rc_bf16rc_f32rc.h
+++ b/ref/gemm/gemm_bf16rc_bf16rc_f32rc.h
@@ -5,7 +5,6 @@
 
 #pragma once
 
-#include "skl-common.h"
 #include <stddef.h>
 
 #if defined(__cplusplus)
--- a/ref/gemm/gemm_bf16rc_bf16rc_f32rc.h
+++ b/ref/gemm/gemm_bf16rc_bf16rc_f32rc.h
@@ -13,8 +13,7 @@ extern "C" {
 #endif
 
 /**
- * @brief Reference bfloat16 matrix-matrix multiplication with float32
- * accumulator.
+ * @brief Reference widening bfloat16 GEMM.
  *
  * @param m - Number of rows in matrices A and C.
  * @param n - Number of columns in matrices B and C.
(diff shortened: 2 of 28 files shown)


myeh01:
Changes made in this PR:
- remove `_pack` suffix from array pointer names
- Simplify some of the descriptions
- Remove unnecessary `#include "skl-common.h"` in reference GEMM headers, add missing `#pragma once`
- Add necessary `#include "skl-common.h"` in reference transpose headers (since the GEMM kernels come before the transpose kernels in `skl-ref.h`, the change above requires making this change)

Aaron-Hutchinson (dismissed):
Seems fine to me. I like the dropping of the `_pack` suffix!

ericlove (dismissed):
Yes, we should aim to keep PR lengths smaller in the future as part of this project's contribution guidelines, but I agree with the changes made here.

myeh01:
Need reapproval after rebasing and making one more change.

ref/pack/transpose_e16.h
Aaron-Hutchinson, line 1: Not relevant to this PR, but why is there an asymmetry between `ref` and `src` for transpose declarations/definitions? In `src` they are in the `pack` files, but in `ref` they have dedicated headers.

CC @Harlen126 @peichialin

ref/pack/transpose_e16.h
```diff
@@ -5,6 +5,7 @@
 
 #pragma once
 
+#include "skl-common.h"
```
Aaron-Hutchinson, line 8: @hdelassus I think this implies we need `skl-common.h` to be included in the install files during the CPack packaging. I can't remember if we included it or not.

Aaron-Hutchinson (approved):
Couple comments that don't impact this PR. Thanks for the cleanup!

ref/pack/transpose_e16.h
myeh01: In `src`, the pack and transpose kernel implementations share functions, so I think it makes sense to have those in a single file. I think that's why they are also declared in the same header. In `ref`, each one can be in its own file.

PR #166, opened 2026-08-06
[GEMM][Pack] Replace Xsfmm ABI macros

myeh01:
This PR replaces the Xsfmm ABI macros `SKL_XSFMM_{IN,OUT,INOUT,NEW}` with the appropriate attribute `__riscv_{in,out,inout,new}("xsfmm")`.

Currently, a work in progress as some Xsfmm kernels (f8e5m2 GEMM, e8 and e16 pack/transpose) will have updated implementations soon.

### DIFF
--- a/include/skl-common.h
+++ b/include/skl-common.h
@@ -57,29 +57,6 @@
  */
 #define SKL_FUNC_UTIL static inline
 
-/*
- * Xsfmm ABI macros:
- *
- * Placeholders for function attributes that indicate how Xsfmm matrix tile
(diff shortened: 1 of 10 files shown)


myeh01:
Need a review on this.

Aaron-Hutchinson:
Should this PR affect the SKL version in any way? I'm not up to speed on these macros/attributes but it sounds like this change requires a more recent version of the toolchain than used previously (which I don't think we've specified anywhere).

PR #170, opened 2026-08-10
Adds READMEs for activations and renames 2D Softmax.

ebavier:
* src/exp/README.md, src/gelu/README.md, src/softmax/README.md: New files.

In gelu and softmax headers, "FP32" -> "F32" and "FP16" -> "F16".  Add [out] and [in] tags to doxygen parameters.  Drop "p" prefix on source and destination pointers.

Name 2D Softmax kernels similarly to GEMM kernels, using "r" suffix on datatype, for "r"ow-major matrix.

E.g. Previous `skl_softmax_2d_f32_zve32f` kernel is now named `skl_softmax_f32r_zve32f`.  This allows a straightforward choice of what to name possible future column-major kernels.


src/exp/README.md
```diff
@@ -0,0 +1,105 @@
...
+constraints, ultimately the circumstances in which a particular
+implementation should be used.
+
+While individual kernels document relevant characteristics in their
+respective headers, it is nevertheless useful to concisely convey some
+of these characteristics directly in the kernel's name, which, amongst
+other benefits, allows for the definition of kernels having different
+characteristics but the same `<type>_<isa>`.
+
+The descriptors used here follow roughly the following grammar:
```
ericlove, line 21: It would be nice to have a high-level summary of this grammar in words, like "The descriptor records an error bound in ULPs terminated by `u`, followed an optional validity domain consisting of either an inequality, an absolute magnitude."

src/gelu/gelu_f32_zve32f.h
```diff
@@ -16,11 +16,11 @@ extern "C" {
 #include <stddef.h>
 
 /**
- * @brief Fast vector FP32 GELU function.
+ * @brief Fast vector F32 GELU function.
  *
- * @param dst - Array of output elements.
- * @param src - Array of input elements.
- * @param n - Number of elements to process.
+ * @param[out] dst - Array of output elements.
```
ericlove: This change is the opposite of our [Doxygen style guide](https://github.com/sifiveinc/skl/tree/main/doc#doxygen-style)

ericlove (commented):
API changes look good, and so do the READMEs. Only issue is we should use at least sentence-level linebreaks in markdown, rather than hard-wrap.

src/gelu/gelu_f32_zve32f.h
ebavier: I see!  I'll revert those bits.

src/exp/README.md
ebavier: Yes.  I tried to capture some of that idea in the `Examples` section, but I can add something up top too.

Anyways, I'll adjust line breaks also.

Thanks for review.

src/gelu/gelu_f32_zve32f.h
ericlove: It's (arguably) better than our current guideline, but that's a decision that will have to be made later.

src/exp/README.md
ericlove: Thanks!

hdelassus (commented):
The README's LGTM. Thank you for adding a "rough" grammar because I realize I didn't have it quite 100% right in my mental model 😆 

As @ericlove said, some of the Doxygen updates must be reverted.

src/exp/README.md
ebavier: I reworked the wording under `<desc>` in the previous section.  I feel like this is a good enough high-level summary? WDYT?

hdelassus (approved):
LGTM!

PR #173, opened 2026-08-12
[Infra] Avoid upward propagation of clang-tidy checks

Aaron-Hutchinson:
Setting the `CMAKE_C_CLANG_TIDY` cache variable can cause clang-tidy checks to propagate upward into any project that uses SKL via CMake's `add_subdirectory`. This PR replaces the usage of that variable with setting the `C_CLANG_TIDY` property of targets manually.


ericlove (approved):
Agree that we should not override higher-level `CMAKE_C_CLANG_TIDY`.

hdelassus (approved):
Thank you, I think that's the right fix.

PR #175, opened 2026-08-12
Fix inconsistent indentation in `CMakeLists.txt`

Aaron-Hutchinson:
Lines in `CMakeLists.txt` are inconsistently indented with either 2 or 4 spaces. 2 space indentation is more common throughout the repository, so this changes any 4 space indentations to 2 spaces.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -21,14 +21,14 @@ endif()
 
 # Set up clang-tidy
 if(SKL_ENABLE_CLANG_TIDY)
-    find_program(SKL_CLANG_TIDY_EXE NAMES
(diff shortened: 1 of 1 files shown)


hdelassus (dismissed):
We should perhaps add a vim modeline like `# vi: set sw=2 et` and the equivalent for Emacs, which I know exists but do not know the syntax of.

Aaron-Hutchinson:
> We should perhaps add a vim modeline like `# vi: set sw=2 et` and the equivalent for Emacs, which I know exists but do not know the syntax of.

Sounds great to me! Let's do this in a follow up PR.

hdelassus (approved):

PR #177, opened 2026-08-13
[Infra][Test] Support using a custom `add_test` CMake command.

Aaron-Hutchinson:
This adds support for overriding the default behavior of the existing `add_test` command.

There are already variables `SKL_TESTER` and `SKL_TESTER_OPTIONS` that allow some customization, such as setting the appropriate tester and any options it may need. However in some cases this may be insufficient, such as if the tester expects arguments in a different order, or if more complex steps are needed in order to prepare/run the test in a given environment. This PR enables swapping out the `add_test` call with a custom CMake function.

Specific changes:
- Extracted `add_test` out of `skl_add_test` into a separate function `skl_add_test_from_executable`.
- Add a new cache variable `SKL_CUSTOM_ADD_TEST` specifying which CMake function should be used to add the test, defaulting to `skl_add_test_from_executable`.
- Call `${SKL_CUSTOM_ADD_TEST}` instead of `add_test` in `skl_add_test`.
- Add `TEST_EXEC` variable to centralize the test executable name.

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -45,38 +45,37 @@ function(skl_add_test_default TEST_NAME HARNESS_SRC TEST_SRC REQUIRED_EXTENSIONS
   endforeach()
 
   set(TEST_NAME test-${TEST_NAME})
-  set(TEST_EXEC ${TEST_NAME})
   set(TEST_DIR "${CMAKE_CURRENT_BINARY_DIR}")
 
   # Create the test executable
-  add_executable(${TEST_EXEC} ${HARNESS_SRC} ${TEST_SRC})
+  add_executable(${TEST_NAME} ${HARNESS_SRC} ${TEST_SRC})
   target_link_libraries(
-    ${TEST_EXEC}
+    ${TEST_NAME}
     PRIVATE ${SKL_LIBRARY} ${SKL_TEST_LIBS} skl-test-driver
   )
   target_compile_options(
-    ${TEST_EXEC}
+    ${TEST_NAME}
     PRIVATE ${SKL_COMPILE_OPTIONS} ${SKL_TEST_COMPILE_OPTIONS} ${OPTS}
   )
-  target_include_directories(${TEST_EXEC} PRIVATE .)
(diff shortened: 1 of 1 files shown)


test/CMakeLists.txt
```diff
@@ -27,34 +43,37 @@ function(skl_add_test TEST_NAME HARNESS_SRC TEST_SRC REQUIRED_EXTENSIONS OPTS)
     endif()
   endforeach()
 
-  add_executable(test-${TEST_NAME} ${HARNESS_SRC} ${TEST_SRC})
+  set(TEST_NAME test-${TEST_NAME})
+  set(TEST_EXEC ${TEST_NAME})
```
ericlove: Why not just:

```suggestion
  set(TEST_EXEC test-${TEST_NAME})
```

ericlove (dismissed):
I think this is fine but:

Taking a step back, it seems like the main purpose for this change is to allow an external project to use the [pre-populated list of SKL tests](https://github.com/sifiveinc/skl/blob/7e3cb40865b96decca462235da80352d9e8a1248/test/CMakeLists.txt#L60-L668) with their own build system.

Perhaps what we really want is to separate that list of test files & ISA requirements out into an independently-accessible form, so projects using SKL's tests can just process the list themselves and customize the build process completely? I realize it will lead to some redundant code in most cases, but the interface seems cleaner that way.  Just a thought.

Aaron-Hutchinson:
I think your proposal is equivalent to allowing an override of the entire `skl_add_test` function instead of just the internal `add_test` portion? Or is putting the list into a separate file more flexible in some way?

ericlove:
> I think your proposal is equivalent to allowing an override of the entire `skl_add_test` function instead of just the internal `add_test` portion?

That's probably the easiest and cleanest solution, yes.

Aaron-Hutchinson:
I've made the change, which also simplified the overall changes being made here. Let me know what you think.

hdelassus (approved):
LGTM

ericlove (approved):
LGTM! Good to abstract away the prefix of `test-${TEST_NAME}` in addition to supporting this `skl_add_test` override.

PR #174, opened 2026-08-12
[Test] Clean up f32 GEMM tests 

myeh01:
This is a follow-up PR to #165 that cleans up the f32 GEMM tests:
- The `_pack` suffix has been removed from array names,
- Uncecessary headers were removed, comments updated, default strides fixed,
- The Xsfmm tests use SKL_XSFMM_TE rather than a hard-coded value of 64,
- The arguments of the function checking for clobbered matrix entries have been renamed.

Subsequent PRs will be opened for the other types.

PR #179, opened 2026-08-14
[Test] Clean up f16 GEMM tests

myeh01:
This is a follow-up PR to #174 for the f16 GEMM tests.


Aaron-Hutchinson (approved):
LGTM, thanks!

PR #180, opened 2026-08-17
[Test][GEMM] Add padded-stride coverage for f32 RVV GEMM

Srinidhi-Magesh:
Adds padded leading-dimension coverage for generic and X390 F32 RVV GEMM.

Tests independently padded A, B, and C strides plus a combined padded case using irregular M=7, N=17, K=13 dimensions.

The optimized kernels are checked against the scalar reference, and existing clobber detection verifies C padding integrity.

Validation:
- Targeted generic/X390 suites: PASS
- 8/8 new padded-stride executions: PASS
- Full correctness regression: 39/39 PASS
- clang-format: PASS
- License header check: PASS
- git diff --check: PASS


myeh01:
@Srinidhi-Magesh Could you push an empty commit or rebase your branch? There was an issue triggering CI when this PR was opened, but it should be fixed now.

Aaron-Hutchinson (approved):
Thanks for adding these tests, @Srinidhi-Magesh. I think we're missing these kind of nontrivial leading dimension cases in many of the GEMM tests and they should probably be added for extra coverage.

PR #182, opened 2026-08-20
[GEMM] Use void * type for c in f32 Xsfmm fused kernel API

myeh01:
This PR updates `c` to have a `void *` type rather than `float *` in the fused kernel API. This enables reuse of the GEMM dispatch function for different `c` types.

The tile load function API has been modified from
```
void skl_gemm_tile_load_f32_f32rcptexterc_f32_xsfmmbase(
    size_t m, size_t n, const float *c, size_t rsc0, size_t csc0, size_t rsc1,
    size_t csc1, size_t tss, size_t rstss, size_t cstss) SKL_XSFMM_OUT;
```
to
```
void skl_gemm_tile_load_f32_f32rcptexterc_f32_xsfmmbase(
    size_t m, size_t n, const void *c, size_t rsc0, size_t csc0, size_t rsc1,
    size_t csc1, size_t row1, size_t col1, size_t tss, size_t rstss,
    size_t cstss) SKL_XSFMM_OUT;
```

`skl_gemm_apply_fused_f32_f32_f32rcptexterc_xsfmmbase` now takes `void *c` instead of `float *c` and has been renamed to `skl_gemm_apply_fused_f32_f32_rcptexterc_xsfmmbase`.

`skl_gemm_apply_tiling_f32rcptex1c_f32rcp1xte_f32rcptexterc_xsfmm32a32f` now takes an addition `tile_load` argument which points to the tile load function to be used and also takes `void *c` instead of `float *c`. It has been renamed to `skl_gemm_apply_tiling_f32rcptex1c_f32rcp1xte_f32_rcptexterc_xsfmm32a32f`. In other words, the type specifier substring has been changed from `<a type>rcptex1c_<b type>rcp1xte_<c type>rcptexterc` to `<a type>rcptex1c_<b type>rcp1xte_<accum type>_rcptexterc`.

The GEMM dispatch function `skl_gemm_fused_f32rcptex1c_f32rcp1xte_f32rcptexterc_xsfmm32a32f` takes `void *c` instead of `float *c` and has been renamed to `skl_gemm_fused_f32rcptex1c_f32rcp1xte_f32_rcptexterc_xsfmm32a32f`. The name changes are the same as those for the tiling application function.

### DIFF
--- a/src/gemm/fused-kernel-API.md
+++ b/src/gemm/fused-kernel-API.md
@@ -87,9 +87,9 @@ Fused kernels operate on a single tile in the Xsfmm tile state and a single bloc
 Once a fused kernel has been implemented, the output of any inner loop function can be processed by applying it to each tile of the output.
 This single-tile API avoids the need for kernel authors to know how each inner loop function arranges data in matrix registers.
 
-The API for fused kernels is
+A fused kernel must have the following function signature:
 ```
-void kernel(size_t tm, size_t tn, size_t tss, <c type> *c, size_t rsc0,
+void kernel(size_t tm, size_t tn, size_t tss, void *c, size_t rsc0,
             size_t csc0, size_t rsc1, size_t csc1, size_t row1, size_t col1,
             void *params) __riscv_in("xsfmm");
 ```
@@ -97,7 +97,8 @@ where
 - `tss` is the tile subset specifier for the first row (column) of the subtile to be operated on
 - `tn` is the length of the row (resp. column) `tss` specifies
 - `tm` is the number of rows (resp. columns) in the subtile, i.e. the kernel operates on rows (resp. columns) `tss` to `tss + tm - 1`
-- `c + row1 * rsc1 + col1 * csc1` points to the block of `C` to be operated on
+- `c + row1 * rsc1 + col1 * csc1` points to the block of `C` to be operated on.
+Kernels should cast `c` to the appropriate type.
 - `rsc0` and `csc0` are the block's row and column strides
 - `params` points to a struct containing any other parameters the kernel needs.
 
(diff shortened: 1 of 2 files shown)


src/gemm/fused-kernel-API.md
```diff
@@ -166,23 +168,18 @@ __asm__ volatile(
     : "v0", "v1", "v2", "v3", "v4", "v5", "v6", "v7", "v16", "v17", "v18",
       "v19", "v20", "v21", "v22", "v23", "vtype", "vl", "memory");
 ```
+High-performance implementations may need to unroll and software pipeline this loop to hide the latency of the `sf.vtmv.v.t` instructions.
 
 #### Adding a bias
-The following kernel can be used to compute `C = A * B + bias`, where `bias` is a row vector that is added to each row of `A * B`:
+The following kernel can be used to compute `C = A * B + bias`, where `bias` is a row vector that is added to each row of `A * B`. Since this kernel requires one only pointer, `float *bias`, we can pass it directly as `params` to avoid defining an extra struct.
```
hdelassus, line 174: Semantic line break + "one only" -> "only one"?
```suggestion
The following kernel can be used to compute `C = A * B + bias`, where `bias` is a row vector that is added to each row of `A * B`.
Since this kernel requires only one pointer, `float *bias`, we can pass it directly as `params` to avoid defining an extra struct.
```

src/gemm/fused-kernel-API.md
```diff
@@ -192,21 +189,15 @@ void skl_gemm_add_bias_f32_f32_f32rcptexterc_xsfmmbase(
 Note that if `tss` specifies a column, then the bias is added to the *columns* of the subtile.
 
 #### Computing a global maximum
-To compute the maximum value of a matrix product, the following kernel can be used tile-by-tile to update the current maximum:
+To compute the maximum value of a matrix product, the following kernel can be used tile-by-tile to update the current maximum. As with the bias GEMM example, we can simply pass `float *max` directly as `params`.
```
hdelassus, line 192:
```suggestion
To compute the maximum value of a matrix product, the following kernel can be used tile-by-tile to update the current maximum.
As with the bias GEMM example, we can simply pass `float *max` directly as `params`.
```

hdelassus (commented):
Looks good to me but it looks like none of these APIs are exposed in `skl.h`. If this is true, could we maybe add a disclaimer saying this is work in progress and subject to backward-incompatible changes like these? If this was already stated somewhere, I missed it.

PR #183, opened 2026-08-20
[Sigmoid] Fix benchmark-only tests compile errors

hdelassus:
When SKL_ENABLE_TESTS=OFF and SKL_ENABLE_BENCHMARKS=ON, some static
symbols were defined yet not used, leading to the compiler emitting
errors at compile-time.  This patch makes those symbols extern. Since
we're building executables, this shouldn't cause any problem.

### DIFF
--- a/test/sigmoid/sigmoid_bf16_xsfvfbfa.c
+++ b/test/sigmoid/sigmoid_bf16_xsfvfbfa.c
@@ -35,27 +35,27 @@
 #define SWIGLU_MAX (+6.f)
 #define SWIGLU_DELTA (__bf16)0.5
 
-static void test_swish_bf16_xsfvfbfa(__bf16 *out, const __bf16 *in, size_t n) {
+void test_swish_bf16_xsfvfbfa(__bf16 *out, const __bf16 *in, size_t n) {
   skl_swish_bf16_xsfvfbfa(out, SWISH_BETA, in, n);
 }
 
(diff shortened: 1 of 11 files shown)


test/sigmoid/sigmoid_bf16_xsfvfbfa.c
```diff
@@ -39,25 +39,27 @@ static void test_swish_bf16_xsfvfbfa(__bf16 *out, const __bf16 *in, size_t n) {
...
 
-static void ref_glu_bf16(__bf16 *out, const __bf16 *in, size_t n) {
-  skl_glu_bf16_ref(out, in, in, n);
-}
-
 static void test_swiglu_bf16_xsfvfbfa(__bf16 *out, const __bf16 *in, size_t n) {
   skl_swiglu_bf16_xsfvfbfa(out, in, in, SWIGLU_DELTA, n);
 }
 
+#if defined(SKL_ENABLE_TESTS)
```
Aaron-Hutchinson: I'm not a huge fan of putting `#ifdef SKL_ENABLE_TESTS` everywhere to conditionally define functions. Is there any harm in just omitting `static` from the reference wrappers?

hdelassus: I don't think so. Thanks for the suggestion, I will revert and update.

Aaron-Hutchinson (approved):
LGTM, thanks for the fix!

PR #184, opened 2026-08-24
[CI] Add workflow_dispatch trigger

myeh01:
This PR adds the `workflow_dispatch` trigger to the CI yaml to enable manual triggering of CI.


hdelassus (approved):
LGTM

PR #191, opened 2026-09-01
[Test] Clean up GEMM f64 tests

myeh01:
This is a follow-up PR to #174 for the f64 GEMM tests.

Thanks @Aaron-Hutchinson for reviewing all of these! This should be the last one.


Aaron-Hutchinson (approved):
LGTM. Glad we got all of these cleaned up! Thanks for working through them all.

